PR #10, opened 2019-08-30
Create LinkedList

Korfat:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,162 @@
...
+
     @Override
     public void add(T value) {
-
+        Node<T> newElem;
+        if (first.item == null) {
+            newElem = new Node<>(null, value, last);
+            first = newElem;
+            last.prev = first;
+        } else if (last.item == null) {
```
boroda4436: @Korfat Do you need this?? Why you want to add at the end? Can't we add one by one starting from the first?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,162 @@
...
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            checkIndexException(index);
+        }
+        if (index == size) {
+            add(value);
```
boroda4436: if you want add to the tail you may create a separate method addToTail(value) and move the functionality from add(value) method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,162 @@
...
+
+    private void checkIndexException(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException();
+        }
+    }
+
+    private void addToTail(T value) {
+        Node<T> newElem = new Node<>(last, value, null);
+        if (last.item == null) {
```
boroda4436: @Korfat this is a bit confusing: you are checking the last.item, but using first.item... The other developer should find the code and move to it

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,162 @@
...
+        } else {
+            addToTail(value);
+        }
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
+            checkIndexException(index);
```
boroda4436: @Korfat should you use checkIndexException inside if statement or instead of it?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,162 @@
...
     }
 
     @Override
     public T remove(T t) {
+        Node<T> result = first;
+        for (int i = 0; i < size; i++) {
+            if (result.item == t) {
+                return remove(i);
+            }
+            result = result.next;
```
boroda4436: @Korfat shouldn't you manage two way link here? You linked to the new node only in one direction

PR #17, opened 2019-08-31
Solution added

gkhrshch:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,192 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private static final int CAPACITY = 16;
```
boroda4436: @gkhrshch do you need this?)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,192 @@
...
+    private int size = 0;
+    private Node<T> header;
+    private Node<T> currentNode;
+    private Node<T> first;
```
boroda4436: @gkhrshch what is the conceptual difference between first and header? These variable names could be misunderstanding. Let's call them header and tail

PR #23, opened 2019-09-02
linkedlist

AlexKukhtinDev:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,141 @@
...
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("The index is out of range: " + index);
+        }
+        if (index == size) {
+            add(value);
+        } else {
+            addFirst(value);
```
boroda4436: @AlexKukhtinDev Shouldn't you add the value at the specific position (index). As I understand this method will add the value to the head of LinkedList

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,141 @@
...
+            firstNode.previous = newNode;
+        }
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-
+        for (T t : list) {
+            this.add(t);
```
boroda4436:
```suggestion
            add(t);
```
_this_ in this case is not required

PR #27, opened 2019-09-02
hw-8.1-solution

SviatoslavExpert:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,150 @@
...
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index > size || index < 0) {
+            throw new IndexOutOfBoundsException();
+        }
+        if (index == size) {
+            Node<T> previous = tail;
```
Sofasmile: we can call method add(T t) here

SviatoslavExpert: Fixed.  The method add(value) is there.  Thank you.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,142 @@
...
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index > size || index < 0) {
+            throw new IndexOutOfBoundsException();
+        }
+        if (index == size) {
+            this.add(value);
```
Sofasmile:
```suggestion
            add(value);
```

PR #38, opened 2019-12-12
Added solution for hw 8.1

ElvenNurse:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -2,49 +2,152 @@
...
+        checkElementIndex(index);
+        Node<T> x = first;
+        for (int i = 0; i < index; i++) {
+            x = x.next;
+        }
+        return x;
+    }
+
+    private static class Node<T> {
+        T item;
```
Tarabarnik: add access modifiers to node's fields

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -2,49 +2,152 @@
...
     public T remove(int index) {
-        return null;
+        checkElementIndex(index);
+        return remove(get(index));
     }
 
     @Override
     public T remove(T t) {
+        for (Node<T> i = first; i != null; i = i.next) {
+            if (Objects.equals(i.item, t)) {
```
Tarabarnik: try to solve without Objects class

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -2,49 +2,152 @@
...
+        return size == 0;
+    }
+
+    private void checkElementIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Index " + index + " out of bounds");
+        }
+    }
+
+    private void checkPositionIndex(int index) {
```
Tarabarnik: redundant method

ElvenNurse: Not at all, it has different top limit, if I check index only less than size, than I cannot add new element in empty list with Index 0 out of bounds Exception (size = 0, and I cheking for index < size), if I check index that may be equal to size, than I recive NullPointerException when try to get element with index = size;

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,147 @@
...
-
+        checkElementIndex(index);
+        getNode(index).item = value;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkElementIndex(index);
+        return remove(get(index));
```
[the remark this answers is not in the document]
ElvenNurse: getNode(index) will return Node, but get(index) will return value (T) of that node

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -24,49 +21,43 @@ public void add(T value) {
...
-        }
+        next.prev = newNode;
         size++;
     }
 
     @Override
     public void addAll(List<T> list) {
-        for (int i = 0; i < list.size(); i++) {
-            add(list.get(i));
+        for (T t : list) {
```
Sofasmile: t - не зрозуміла назва

PR #73, opened 2020-04-08
added solution

popovychyb:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,141 @@
...
+        checkIndex(index, size);
+        if (index == size) {
+            add(value);
+        } else if (index == 0) {
+            head = new Node<>(null, value, head);
+            head.next.previous = head;
+            size++;
+        } else {
+            Node<T> actual = getNode(index);
+            Node newNode = new Node<>(actual.previous, value, actual);
```
RomanDubovskyi, line 34:
```suggestion
            Node<T> newNode = new Node<>(actual.previous, value, actual);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,141 @@
...
+            newNode.previous.next = newNode;
```
RomanDubovskyi, line 35:
```suggestion
            newNode.previous.next = newNode;
            newNode.next.previous = newNode;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,141 @@
...
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        checkIndex(index, size);
+        T result = getNode(index).value;
+        remove(index);
+        add(value, index);
+        return result;
```
RomanDubovskyi, line 60:
```suggestion
      Node<T> node = getNode(index);
        T oldValue = node.value;
        node.value = value;
        return oldValue;
```
this way you don't need to call two methods add and remove, because you just need to override value

popovychyb: Done. Thank you!

src/main/java/core/basesyntax/MyLinkedList.java
popovychyb: Done. Thank you!

src/main/java/core/basesyntax/MyLinkedList.java
popovychyb: Done. Thank you!

PR #80, opened 2020-04-08
added solution

tanchak123:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,111 @@
...
-        return null;
+        return getNode(index).item;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        T old = getNode(index).item;
+        getNode(index).item = value;
+        return old;
```
RomanDubovskyi, line 50:
```suggestion
       Node<T> node = getNode(index);
        T oldValue = node.item;
        node.item = value;
        return oldValue;
```
it's better to save node that you've found so you don't need to call getNode() twice

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,111 @@
...
+        element.next.prev = element.prev;
+        size--;
+        return past;
     }
 
     @Override
     public boolean remove(T t) {
+        for (int i = 0; i < size; i++) {
+            if (getNode(i).item == null && t == null
+                    || getNode(i).item.equals(t) && getNode(i).item != null) {
```
RomanDubovskyi, line 67: calling getNode(i) every time is bad for performance, moreover you're calling it three times per iteration 
consider iterate nodes in this method with node = node.next so you need to do it only once

PR #83, opened 2020-04-08
Hw 8.1

fruts:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,203 @@
...
+    private Node<T> first;
+    private Node<T> last;
+    private int size;
+
+    private static class Node<T> {
+        T element;
+        Node<T> next;
+        Node<T> previous;
+
+        public Node(Node<T> previous, T element, Node<T> next) {
```
RomanDubovskyi, line 16: do we need this constructor to be public ?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,203 @@
...
+        if (index == 0) {
+            addFirst(value);
+        } else {
+            if (index == size) {
+                addToEnd(value);
+            } else {
+                if (isIndexExist(index)) {
+                    Node<T> newNode = new Node<>(null,value,null);
+                    Node<T> oldNode = findForIndex(index);
+                    Node<T> oldPrev = findForIndex(index - 1);
```
RomanDubovskyi, line 50:
```suggestion
                    Node<T> oldPrev = oldNode.prev;
```
you don't need to call findForIndex() twice here

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,203 @@
...
+                    if (oldNode.element != null) {
```
RomanDubovskyi, line 51: you can put null as value, so i don't understand why we need this check ?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,203 @@
...
+                }
+            }
+        }
     }
 
     @Override
     public boolean addAll(List<T> list) {
-        return false;
+        if (list == null) {
+            throw new NullPointerException();
```
RomanDubovskyi, line 66: you don't need to throw NullPointer, it will be thrown automatically if you try  to null.someMethod or null.someField

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,203 @@
...
+        }
+        for (T one: list) {
+            add(one);
+        }
+        return true;
     }
 
     @Override
     public T get(int index) {
+        if (isIndexExist(index)) {
```
RomanDubovskyi, line 76: this check is redundant, you're doing it inside of findForIndex() method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,203 @@
...
+            Node<T> requiredNode = findForIndex(index);
+            if (index == 0 && size < 2) {
+                size--;
+                first = null;
+                last = null;
+                requiredNode.element = null;
+                return null;
+            }
+            if (index == 0) {
+                first = findForIndex(index + 1);
```
RomanDubovskyi, line 105: you should unlink node, that you removed 
in your case first.prev still pointing at your "removed" node

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,200 @@
...
+                    size++;
+                }
+            }
+        }
     }
 
     @Override
     public boolean addAll(List<T> list) {
-        return false;
+        if (list == null) {
```
RomanDubovskyi, line 63: you don't need this check

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,200 @@
...
     public boolean isEmpty() {
+        if (last == null && first == null) {
+            return true;
+        }
         return false;
     }
+
+    private Node<T> findForIndex(int index) {
+        if (isIndexExist(index)) {
+            Node<T> requiredNode = first;
```
RomanDubovskyi, line 151: try to optimize this check if index < size / 2 gor from first to middle of list, if not go from last ot middle

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,214 @@
...
+    private boolean isIndexExist(int index) {
+        if (index < 0 || index >= size()) {
+            throw new IndexOutOfBoundsException();
+        }
+        return true;
+    }
+
+    public static void main(String[] args) {
+        for (int i = 10; i > 6; i--) {
+            System.out.println(i);
```
RomanDubovskyi, line 213: don't think we need it here

PR #84, opened 2020-04-08
added solution at 22:22))

6ALLIKA:


6ALLIKA:
ill try refactor after lecture, but i think u will give me some tips before i start)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,139 @@
...
+        if (index == size) {
+            add(value);
+            return;
+        }
+        if (index > 0 && index < size) {
+            Node<T> atIndexNode = head;
+            for (int i = 0; i < index; i++) {
+                atIndexNode = atIndexNode.nextElement;
+            }
+            atIndexNode.prevElement.nextElement
```
RomanDubovskyi, line 35: atIndexNode.prev  should reference you new node, but it still references old node

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,139 @@
...
+        return true;
     }
 
     @Override
     public T get(int index) {
-        return null;
+        Node<T> currentNode = head;
+        if (index >= 0 && index < size) {
+            for (int i = 0; i < index; i++) {
+                currentNode = currentNode.nextElement;
```
RomanDubovskyi, line 56: try to optimize this method: 
if your    size / 2 > index > 0 - go from first to index 
if your    index > size / 2 go from last to index

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,139 @@
...
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        T result = get(index);
+        Node<T> atIndexNode = head;
+        if (isIndexExists(index)) {
+            for (int i = 0; i < index; i++) {
+                atIndexNode = atIndexNode.nextElement;
```
RomanDubovskyi, line 69: you don't need this for loop, you just got your node by index, so you can just override it's value, Example: 
```
        T oldValue = node.item;
        node.item = value;
        return oldValue;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,151 @@
...
+        Node<T> node = getNode(index);
+        T oldValue = node.element;
+        node.element = value;
+        return oldValue;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (size == 0) {
```
RomanDubovskyi, line 71: why not to use your indexExtends() method ? 
because current check doesn't cover situation where size = 3 and index = 8

PR #112, opened 2020-08-05
leonid-linkedlist

Rommelua:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,50 +1,162 @@
...
+    }
+
+    private void unLink(Node<T> node) {
+        Node<T> previous = node.prev;
+        Node<T> next = node.next;
+
+        if (previous == null) {
+            first = next;
+        } else {
+            previous.next = next;
```
KseniiaMakarova, line 136:
```suggestion
            previous.next = next;
            node.prev = null;
```
in this task it is important to understand the concept of links

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,50 +1,162 @@
...
+        }
+        if (next == null) {
+            last = previous;
+        } else {
+            next.prev = previous;
```
KseniiaMakarova, line 141:
```suggestion
            next.prev = previous;
            node.next = null;
```

PR #113, opened 2020-08-06
Added solution 9

DmGrt:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,140 @@
...
     public T get(int index) {
-        return null;
+        return getEntry(index).element;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        T oldValue = getEntry(index).element;
+        getEntry(index).element = value;
```
dmytro-hryhoruk, line 64: let's save  getEntry() to local variable in order not to iterate twice

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,143 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Entry<T> first;
+    private Entry<T> last;
+    private int size;
+
+    public MyLinkedList() {
+        this.size = 0;
```
Sofasmile, line 11: do we need this constructor?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,143 @@
...
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Index out of range!");
+        }
+        if (size == index) {
+            add(value);
```
Sofasmile, line 33: let's add return here, and then we can remove one `else` block

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,143 @@
...
+            first = removedEntry.next;
+        } else {
+            removedEntry.previous.next = removedEntry.next;
+        }
+        if (removedEntry.next == null) {
+            last = removedEntry.previous;
+        } else {
+            removedEntry.next.previous = removedEntry.previous;
+        }
+        size--;
```
Sofasmile, line 84: let's move this code at separate method `unlink(Node<T> nodeToRemove)`  
then we can call this method in remove(T t)

PR #114, opened 2020-08-06
Added solution

AndrewPolishchuc:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,189 @@
...
+        movingElement.prev = inputNode;
+        if (currentPrev == null) {
+            first = inputNode;
+        } else {
+            currentPrev.next = inputNode;
+        }
+        size++;
+    }
+
+    public Node<T> node(int index) {
```
KseniiaMakarova, line 151:
```suggestion
    public Node<T> findNode(int index) {
```
what's wrong with the access modifiers for all the helper methods?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,189 @@
...
+        Node<T> currentElement = first;
+        for (int i = 0; i < index; i++) {
+            currentElement = currentElement.next;
+        }
+        return currentElement;
```
KseniiaMakarova, line 156: in this method let's start moving from the first Entry OR from the last one
depending on whether index is closer to the start or to the end

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,189 @@
...
+            Node<T> inputNode = new Node<>(elementAddArray, null, prev);
+            if (prev == null) {
+                first = inputNode;
+            } else {
+                prev.next = inputNode;
+            }
+            prev = inputNode;
+        }
+        size += lengthOfArray;
+        return true;
```
KseniiaMakarova, line 55: can you just make use of add() method here?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,189 @@
...
+                }
+            }
+        } else {
+            for (Node<T> i = first; i != null; i = i.next) {
+                if (i.element.equals(t)) {
+                    unlink(i);
+                    return true;
+                }
+            }
+        }
```
KseniiaMakarova, line 95: refactor this using one `while` loop and one `if` check

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,189 @@
...
+    private void linkFirst(T e) {
+        final MyLinkedList.Node<T> f = first;
+        final MyLinkedList.Node<T> newNode = new MyLinkedList.Node<>(e, f, null);
+        first = newNode;
+        if (f == null) {
+            last = newNode;
+        } else {
+            f.prev = newNode;
+        }
+        size++;
```
KseniiaMakarova, line 118: 1. create informative variable names
2. the use of `final` keyword must have a reason other than just bypassing the checkstyle
3. you call this method if size == 0, can `first` be other than null?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,189 @@
...
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index < 0 || index > size) {
+            throw new ArrayIndexOutOfBoundsException("Out of bounds search index");
+        }
+        if (index == size) {
+            linkLast(value);
```
KseniiaMakarova, line 30:
```suggestion
            add(value);
```
what if you call `new MyLinkedList<Integer>().add(3, 0)`?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,189 @@
...
+    void linkLast(T e) {
+        final Node<T> l = last;
+        final Node<T> newNode = new MyLinkedList.Node<>(e, null, l);
+        last = newNode;
+        if (l == null) {
+            first = newNode;
+        } else {
+            l.next = newNode;
+        }
+        size++;
```
KseniiaMakarova, line 130: 1. create informative variable names
2. the use of `final` keyword must have a reason other than just bypassing the checkstyle
3. you call this method if size != 0, can `last` be null?

src/main/java/core/basesyntax/MyLinkedList.java
AndrewPolishchuc: Done)

src/main/java/core/basesyntax/MyLinkedList.java
AndrewPolishchuc: Fixed.

src/main/java/core/basesyntax/MyLinkedList.java
AndrewPolishchuc: Of course. Done)
Why is addAll not implemented through add in the original class?

src/main/java/core/basesyntax/MyLinkedList.java
AndrewPolishchuc: I changed the code. I used do while but it took me 2 checks(if) because it might come null. Done

src/main/java/core/basesyntax/MyLinkedList.java
AndrewPolishchuc: Of course.. Done.

src/main/java/core/basesyntax/MyLinkedList.java
AndrewPolishchuc: 1. Done;
2. Done;
3. We can call method when size != 0 and the first element is no longer null. Of course, if we only insert when size == 0 check unnecessary.

src/main/java/core/basesyntax/MyLinkedList.java
AndrewPolishchuc: 1. Done;
2. Done;
3. Similarly linkFirst.

src/main/java/core/basesyntax/MyLinkedList.java
AndrewPolishchuc: Done)

src/main/java/core/basesyntax/MyLinkedList.java
AndrewPolishchuc: 3. I am calling the method "linkLast" in method "add(T value, int index)" and size inside this method can be == 0 and last can be null

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,178 @@
...
+            last = newNode;
+        } else {
+            firstElement.prev = newNode;
+        }
+        size++;
+    }
+
+    private void linkLast(T e) {
+        Node<T> lastElement = last;
+        Node<T> newNode = new MyLinkedList.Node<>(e, null, lastElement);
```
RomanDubovskyi, line 102:
```suggestion
        Node<T> newNode = new Node<>(e, null, lastElement);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,178 @@
...
 
     @Override
     public boolean remove(T t) {
+        int i = 0;
+        while (i != size) {
+            if (t == get(i) || get(i).equals(t)) {
+                unlink(findNode(i));
+                return true;
+            }
+            i = i + 1;
```
RomanDubovskyi, line 73:
```suggestion
            i++;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,178 @@
...
+        checkIndex(index);
+        return unlink(findNode(index));
     }
 
     @Override
     public boolean remove(T t) {
+        int i = 0;
+        while (i != size) {
+            if (t == get(i) || get(i).equals(t)) {
+                unlink(findNode(i));
```
RomanDubovskyi, line 70: what if get(i) is null?

PR #116, opened 2020-08-06
added solution

IlyaKvitkovskyi:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,118 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private Node<T> entry(int index) {
```
Sofasmile, line 88:
```suggestion
    private Node<T> getNodeByIndex(int index) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,118 @@
...
     public T remove(int index) {
-        return null;
+        Node<T> element = entry(index);
+        element.prev.next = element.next;
+        element.next.prev = element.prev;
+        element.next = null;
+        element.prev = null;
+        T value = element.item;
+        element.item = null;
+        size--;
```
Sofasmile, line 61: let's move this code to separate private method `unlink(Node<T> nodeToRemove)` and then we can use it in method remove(T t)

PR #120, opened 2020-08-06
added solution

AnaLanda:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,50 +1,163 @@
...
+import java.util.Iterator;
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size;
+    private Node<T> firstNode;
+    private Node<T> lastNode;
+
+    public MyLinkedList() {
+        size = 0;
```
Sofasmile, line 12: these are default values - so this may be omitted

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,50 +1,163 @@
...
+    }
+
+    private static class Node<T> {
```
Sofasmile, line 15: let's move this inner class at the end of external class

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,50 +1,163 @@
...
+        }
     }
 
     @Override
     public boolean addAll(List<T> list) {
-        return false;
+        if (list.size() == 0) {
+            return false;
+        }
+        Iterator iterator = list.iterator();
```
Sofasmile, line 59: let's use foreach here

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,50 +1,163 @@
...
 
     @Override
     public T get(int index) {
-        return null;
+        checkIndex(index);
+        Node<T> soughtForNode = findNodeByIndex(index);
+        return soughtForNode.value;
     }
 
+    // DOES THIS METHOD REPLACE THE NODE OR SHIFT IT?
```
Sofasmile, line 73: this method should replace the value of the node)
let's remove this comment from code

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,50 +1,163 @@
...
+        return soughtForNode;
+    }
+
+    private void removeNodeLinks(Node<T> nodeToBeRemoved) {
+        Node<T> previousNode = nodeToBeRemoved.prev;
+        Node<T> nextNode = nodeToBeRemoved.next;
+        if (previousNode == null) {
+            firstNode = nextNode;
+        } else {
+            previousNode.next = nextNode;
```
Sofasmile, line 137: in this task it is important to understand the concept of links

AnaLanda:
> why do you remove methods: findNodeByIndex() and removeNodeLinks()?

I rewrote the whole code after the code along with Bohdan today. Should I go back to the previous version where I had  findNodeByIndex() and removeNodeLinks()?

PR #123, opened 2020-08-07
added realization

tarkaiev:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,153 @@
...
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size;
+    private Node<T> head;
+    private Node<T> tail;
+
+    public MyLinkedList() {
+        head = null;
+        tail = null;
+        size = 0;
+    }
```
KseniiaMakarova, line 14: these are default values - so this may be omitted

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,153 @@
...
+        } else {
+            next.previous = previous;
+        }
+        size--;
+    }
+
+    private static class Node<T> {
+        private T value;
+        private Node<T> previous;
+        private Node<T> next;
```
KseniiaMakarova, line 153: let's create a constructor and use it to create new nodes

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,153 @@
...
+        }
+    }
+
+    private Node<T> getNode(int index) {
+        checkIndex(index);
+        Node<T> node = head;
+        for (int i = 0; i < index; i++) {
+            node = node.next;
+        }
+        return node;
```
KseniiaMakarova, line 130: in this method let's start moving from the first Entry OR from the last one
depending on whether index is closer to the start or to the end

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,153 @@
...
+    }
+
+    private void removingAction(Node<T> target) {
+        Node<T> previous = target.previous;
+        Node<T> next = target.next;
+
+        if (previous == null) {
+            head = next;
+        } else {
+            previous.next = next;
```
KseniiaMakarova, line 140:
```suggestion
            previous.next = next;
            target.previous = null;
```
in this task it is important to understand the concept of links

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,153 @@
...
+        }
+        if (next == null) {
+            tail = previous;
+        } else {
+            next.previous = previous;
```
KseniiaMakarova, line 145:
```suggestion
            next.previous = previous;
            target.next = null;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,153 @@
...
+        if (index == size) {
+            Node<T> previousTail = tail;
+            tail = new Node<>();
+            tail.previous = previousTail;
+            tail.value = value;
+            if (previousTail == null) {
+                head = tail;
+            } else {
+                previousTail.next = tail;
+            }
```
KseniiaMakarova, line 49: if `index == size`, can we use our own code?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,153 @@
...
+        } else {
+            Node<T> nodeBeforeInput = getNode(index);
+            Node<T> oldPrevious = nodeBeforeInput.previous;
+            Node<T> inputNode = new Node<>();
+            inputNode.previous = oldPrevious;
+            inputNode.value = value;
+            inputNode.next = nodeBeforeInput;
```
KseniiaMakarova, line 56: pass all three values to Node constructor

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,153 @@
...
+        head = null;
+        tail = null;
+        size = 0;
+    }
+
     @Override
     public boolean add(T value) {
-        return false;
+        if (size == 0) {
+            Node<T> node = new Node<>();
```
KseniiaMakarova, line 19: when creating any Node, pass all three values to the constructor

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,153 @@
...
+        Node<T> target = getNode(index);
+        removingAction(target);
+        return target.value;
     }
 
     @Override
     public boolean remove(T t) {
+        Node<T> node = head;
+        while (node != null) {
+            if (node.value == t || node.value.equals(t)) {
```
KseniiaMakarova, line 99: there is a risk of NPE

src/main/java/core/basesyntax/MyLinkedList.java
tarkaiev: removed

src/main/java/core/basesyntax/MyLinkedList.java
tarkaiev: got it, added

src/main/java/core/basesyntax/MyLinkedList.java
tarkaiev: same here

src/main/java/core/basesyntax/MyLinkedList.java
tarkaiev: done

src/main/java/core/basesyntax/MyLinkedList.java
tarkaiev: done

src/main/java/core/basesyntax/MyLinkedList.java
tarkaiev: done

src/main/java/core/basesyntax/MyLinkedList.java
tarkaiev: yes, we can
refactored

src/main/java/core/basesyntax/MyLinkedList.java
tarkaiev: done additional null-checking at the beginning of the method
is it enough or need to do something else here to avoid npe?

src/main/java/core/basesyntax/MyLinkedList.java
tarkaiev: done

src/main/java/core/basesyntax/MyLinkedList.java
tarkaiev: replaced while loop to for with size-limit
no risk of npe now, i guess

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,155 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size;
+    private Node<T> head;
+    private Node<T> tail;
+
+    public MyLinkedList() {
+    }
```
RomanDubovskyi, line 11: we don't need to declare default public constructor, unless we have other constructors

PR #125, opened 2020-08-07
MyLinkedList

Alkey:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,163 @@
...
+    }
+
+    private void removeElement(Entry<T> entry) {
+        if (entry.equals(last)) {
+            last.previous.next = null;
+            last = entry.previous;
+        } else {
+            entry.previous.next = entry.next;
+            entry.next.previous = entry.previous;
+        }
```
KseniiaMakarova, line 150: here I would suggest to create two local variables that point to `entry.previous` and `entry.next`
then write the unlinking logic for `entry.previous` (add check if your `entry` is `first` here)
and the unlinking logic for `entry.next` (add check if your `entry` is `last` here`)

PR #129, opened 2020-08-08
Solution

DariaPikul:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,173 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+
+    private int currentAmount = 0;
```
RomanDubovskyi, line 7:
```suggestion
    private int size = 0;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,173 @@
...
+    private transient Node<T> first;
```
RomanDubovskyi, line 8: what does "transient" do?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,173 @@
...
+    private transient Node<T> last;
+
+    public void check(int index) {
```
RomanDubovskyi, line 11: what about access modifiers in class inner util methods?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,173 @@
...
+            throw new ArrayIndexOutOfBoundsException("Non-valid index");
+        }
+    }
+
     @Override
     public boolean add(T value) {
-        return false;
+        int previousAmount = currentAmount;
+        addElement(value);
+        return currentAmount - previousAmount != 0;
```
```suggestion
        return true;
```
if method completed its work correctly you can just return true, so this method in general looks redundant to me

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,173 @@
...
+            this.next = next;
+        }
+
+        private void setData(T value) {
+            data = value;
+        }
+
+        private T getData() {
+            return data;
+        }
```
RomanDubovskyi, line 135: private setters and getters in this case makes no sense, let's remove them

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,173 @@
...
-
+        if (index != currentAmount) {
+            check(index);
+        }
+        if (index == 0) {
+            addFirst(value);
+        } else if (index == currentAmount) {
+            addLast(value);
+        } else {
+            Node<T> nodeIndex = nodeBy(index);
```
RomanDubovskyi, line 34:
```suggestion
            Node<T> oldNode = nodeBy(index);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,173 @@
...
+            nodeIndex.prev = new Node<>(nodeIndex.prev, value, nodeIndex);
```
RomanDubovskyi, line 35: if you add element to the middle of linkedlist you oldNode.previous.next should point not to the oldNode but to the currently created node, add this logic to method. Because if you iterate node= node.next, you will skip your new node

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,173 @@
...
+            currentAmount++;
+        }
     }
 
     @Override
     public boolean addAll(List<T> list) {
-        return false;
+        int previousAmount = currentAmount;
```
RomanDubovskyi, line 42: this is redundant

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,173 @@
...
+        private void setData(T value) {
+            data = value;
+        }
+
+        private T getData() {
+            return data;
+        }
+    }
+
+    private Node<T> nodeBy(int index) {
```
RomanDubovskyi, line 138: is that correct method naming? 
https://google.github.io/styleguide/javaguide.html#s5.2.3-method-names

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,173 @@
...
+        check(index);
+        Node<T> resultNode = last;
+        while (index != currentAmount - 1) {
```
RomanDubovskyi, line 141: let's iterate our linkedlist depending on location of our node:
* if node is closer to the start, we will go from first to middle, if it’s closer to the end of list we will go from last to middle

src/main/java/core/basesyntax/MyLinkedList.java
DariaPikul: if I got the point right, it means that this object contains not the certain object (static I'd say), but the link on the object, that (link) is changing during the program working

RomanDubovskyi: it is used during serialization to ignore certain fields
https://www.geeksforgeeks.org/transient-keyword-java/

DariaPikul: thanks, now I know why it should be used :)
should I remove those for our project or let them stay?

RomanDubovskyi: let's use only things we need

README.md
```diff
@@ -1,2 +1,20 @@
...
+                index++;
+            }
+        } else {
+            resultNode = first;
+            while (index >= 0) {
+                resultNode = resultNode.prev;
+                index--;
+            }
+        }
+        return resultNode;
```
RomanDubovskyi, line 19: we don't need this code in readme

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,165 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+
+    private int size = 0;
+    private transient Node<T> first;
```
RomanDubovskyi, line 8:
```suggestion
    private Node<T> first;
```

PR #131, opened 2020-08-08
LinkedList class recreated with basic methods

fyarkiy:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,50 +1,148 @@
 package core.basesyntax;
 
 import java.util.List;
+import java.util.NoSuchElementException;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+
+    private T value;
+    private MyLinkedList<T> next;
+    private MyLinkedList<T> prev;
```
RomanDubovskyi, line 10: let's use Nodes in our linkedlist as in original implementation
create inner class Node with these three fields
https://habr.com/ru/post/337558/

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,50 +1,148 @@
...
+                first = newElement;
+            }
+            size++;
+        }
     }
 
     @Override
     public boolean addAll(List<T> list) {
-        return false;
+        if (list.isEmpty()) {
```
RomanDubovskyi, line 68: this check here is redundant, you do it every time you call add method anyway

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,143 @@
...
+        Node<T> element = first;
+        int counter = 0;
+        while (counter < index) {
+            element = element.next;
+            counter++;
+        }
+        return element;
+    }
+
+    private class Node<T> {
```
RomanDubovskyi, line 140:
```suggestion
    private static class Node<T> {
```
we can use static class here

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,143 @@
...
+        return size == 0;
+    }
+
+    private Node<T> getElement(int index) {
+        if (index >= size || index < 0) {
+            throw new IndexOutOfBoundsException(index);
+        }
+        Node<T> element = first;
+        int counter = 0;
+        while (counter < index) {
```
RomanDubovskyi, line 133: let's iterate our linkedlist depending on location of our node:
* if node is closer to the start, we will go from first to middle, if it’s closer to the end of list we will go from last to middle

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,143 @@
...
     public T remove(int index) {
-        return null;
+        T removedItemValue = getElement(index).value;
+
+        if (size == 1) {
+            first = null;
+            lastElement = null;
+        } else {
+            if (getElement(index).next == null) {
+                lastElement = getElement(index - 1);
```
RomanDubovskyi, line 89: instead of calling getElement() method each time (that's whole loop execution) let's save method result into temp variable

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,143 @@
...
     @Override
     public T get(int index) {
-        return null;
+        return getElement(index).value;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        T oldItem = getElement(index).value;
```
RomanDubovskyi, line 75:
```suggestion
        T oldValue = getElement(index).value;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,143 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+
+    private int size;
+    private Node<T> first;
+    private Node<T> lastElement;
+
+    private void firstElement(T value) {
```
RomanDubovskyi, line 11: what about method naming ? 
https://google.github.io/styleguide/javaguide.html#s5.2.3-method-names

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,143 @@
...
     @Override
     public boolean add(T value) {
-        return false;
+        if (size == 0) {
+            firstElement(value);
+        } else {
+            Node<T> newElement = new Node<>();
+            newElement.value = value;
+            newElement.prev = lastElement;
+            lastElement.next = newElement;
```
RomanDubovskyi, line 27: let's use constructor for initializing Node

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,154 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+
+    private int size;
+    private Node<T> first;
+    private Node<T> lastElement;
+    private Node<T> element;
```
RomanDubovskyi, line 10: this variable is redundant, we use as local variable in methods

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,153 @@
...
     public void add(T value, int index) {
-
+        if (index > size || index < 0) {
+            throw new IndexOutOfBoundsException(index);
+        }
+        if (size == 0 || index == size) {
+            add(value);
+        } else {
+            Node<T> element = getElement(index);
+            Node<T> newElement = new Node<T>(value, element, element.prev);
```
RomanDubovskyi, line 36:
```suggestion
            Node<T> newElement = new Node<>(value, element, element.prev);
```
it's redundant

PR #136, opened 2020-08-09
hw.9

OlessiaKharchenko:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,156 @@
...
+        return true;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index == size) {
+            addLastValue(value);
+        } else {
+            addFirstValue(value, findNodeByIndex(index));
```
RomanDubovskyi, line 24: addFirstValue is not quite correct naming since, you don't only use this method to add value to first node

PR #137, opened 2020-08-09
homework 9 solution

tkhambir:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,221 @@
...
+        private Node prev;
+        private Node next;
+
+        Node(T value) {
+            this.value = value;
+        }
     }
 }
+
+
```
RomanDubovskyi, line 223: remove redundant empty lines

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,221 @@
...
+            }
+            result = result.next;
+            if (result == null) {
+                return null;
+            }
+        }
+        return result;
+    }
+
+    private class Node {
```
RomanDubovskyi, line 212:
```suggestion
    private class Node<T> {
```
let's use parameterized class

RomanDubovskyi, line 212:
```suggestion
    private static class Node {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,221 @@
...
+        private T value;
+        private Node prev;
```
RomanDubovskyi, line 214:
```suggestion
        private Node<T> prev;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,221 @@
...
+        if (size == 0) {
+            tail = new Node(value);
+            head = tail;
+            wire(head, tail);
+        } else {
+            Node tmp = tail;
+            tail = new Node(value);
+            wire(tmp, tail);
+        }
+        size += 1;
```
RomanDubovskyi, line 25:
```suggestion
        size++;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,221 @@
...
+
     @Override
     public boolean add(T value) {
-        return false;
+        if (size == 0) {
+            tail = new Node(value);
+            head = tail;
+            wire(head, tail);
+        } else {
+            Node tmp = tail;
```
```suggestion
            Node tempNode = tail;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,221 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return (size == 0);
+    }
+
+    private void validateIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayIndexOutOfBoundsException(index);
```
RomanDubovskyi, line 156: let's add message to make exception more informative

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,221 @@
...
+
+        if (index < size / 2) {
+            result = head;
+            for (int i = 0; i <= index; i++) {
+                if (i == index) {
+                    return result;
+                }
+                result = result.next;
+            }
+        } else {
```
RomanDubovskyi, line 181:
```suggestion
            for (int i = 0; i < index; i++) {
                result = result.next;
            }
        } else {
```
we don't need if check, because we gonna exit loop whenever  i < index is not valid anymore

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,221 @@
...
+        }
+        return result;
+    }
+
+    private Node getNodeByValue(T value) {
+        Node result = head;
+
+        if (result == null) {
+            return null;
+        }
```
RomanDubovskyi, line 198: this check is redundant, because head can only be null if size = 0, so you won't go inside for loop

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,221 @@
...
+        for (int i = 0; i < size; i++) {
+            if (result.value == value
+                    || result.value != null && result.value.equals(value)) {
+                return result;
+            }
+            result = result.next;
+            if (result == null) {
+                return null;
+            }
```
RomanDubovskyi, line 207: same here, if you set for loop condition correctly, you won't go outside your list border, so result won't be null, try to remove this check, so altogether it doesn't seem that you need whole new method just for one loop

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,221 @@
...
+            tail.next = null;
+        } else {
+            wire(target.prev, target.next);
+        }
+        size--;
+        return result;
     }
 
     @Override
     public boolean remove(T t) {
```
RomanDubovskyi, line 110: in this method if else construction are too complicated, try to think how to simplify them

src/main/java/core/basesyntax/MyLinkedList.java
tkhambir: integer value will be wrapped in default message, which is used across whole jdk libraries

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,161 @@
...
+
+    public MyLinkedList() {
+        size = 0;
+    }
+
     @Override
     public boolean add(T value) {
-        return false;
+        if (size == 0) {
+            tail = new Node<>(value);
```
RomanDubovskyi, line 17: this line may be moved outside if else not to repeat code

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,161 @@
...
+        }
+        wire(target.prev, target.next);
+        size--;
+        return target.value;
     }
 
     @Override
     public boolean remove(T t) {
+        for (int i = 0; i < size; i++) {
+            if (get(i) == t || get(i).equals(t)) {
```
RomanDubovskyi, line 108: what if get(i) will be null?

PR #139, opened 2020-08-09
added solution

aleksandr-hrankin:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,145 @@
...
+            head = next;
+        } else {
+            prev.next = next;
+        }
+        if (next == null) {
+            tail = prev;
+        } else {
+            next.prev = prev;
+        }
+        size -= 1;
```
Sofasmile, line 86:
```suggestion
        size--;
```

Sofasmile, line 86: let's move this code at separate method `unlink(Node<> nodeToRemove)` and let's use it at method remove(T t)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,145 @@
...
+        return temp.item;
     }
 
     @Override
     public boolean remove(T t) {
+        for (int i = 0; i < size; i++) {
+            if (t == get(i) || getNodeByIndex(i).item.equals(t)) {
```
Sofasmile, line 93: let's change logic of this method because we should call method get(i) a lot of times
It makes this method very complex

PR #142, opened 2020-08-09
i need one more solution for 'removeIndex' :(

vladislavshvets:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,155 @@
...
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size;
+    private Node<T> head;
+    private Node<T> tail;
+
+    public MyLinkedList() {
+        size = 0;
+    }
+
+    private static class Node<T> {
```
RomanDubovskyi, line 14: let's move this class declaration to the end of the file since it has private modifier

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,155 @@
...
+        node.prev.next = node.next;
+        node.next.prev = node.prev;
+        size--;
+        return node.item;
     }
 
     @Override
     public boolean remove(T t) {
+        for (int i = 0; i < size; i++) {
+            if (t == get(i) || get(i).equals(t)) {
```
RomanDubovskyi, line 122: what if get(i) is null?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,155 @@
...
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index == size) {
+            addLast(value);
+        } else {
+            Node<T> currentNode = getNode(index);
+            Node<T> prev = getNode(index).prev;
```
RomanDubovskyi, line 38:
```suggestion
            Node<T> prev = currentNode.prev;
```
there is no need to call method once again, you execute extra loop

PR #143, opened 2020-08-09
added solution

Rudiachenko:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,143 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size;
+    private Node<T> first;
+    private Node<T> last;
+
+    private static class Node<T> {
```
RomanDubovskyi, line 10: since class is private let's move it's declaration to the end

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,143 @@
...
     @Override
     public boolean add(T value) {
-        return false;
+        Node<T> node = new Node<>(null, value, null);
+        if (size == 0) {
+            first = node;
+        } else {
+            node.prev = last;
+            last.next = node;
+        }
```
RomanDubovskyi, line 30:
```suggestion
        Node<T> node = new Node<>(last, value, null);
        if (size == 0) {
            first = node;
        } else {
            last.next = node;
        }
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,143 @@
...
+        size--;
+        return node.element;
     }
 
     @Override
     public boolean remove(T t) {
+        Node<T> currentNode = first;
+        for (int i = 0; i < size; i++) {
+            if (t == currentNode.element || currentNode.element.equals(t)) {
```
RomanDubovskyi, line 104: what if currentNode.element  is null and t is't null

PR #153, opened 2020-08-10
Dev

yutsevich:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,136 @@
...
+    private int size;
+    private Node<T> head;
+    private Node<T> tail;
+
+    private static class Node<T> {
+        T item;
+        Node<T> next;
+        Node<T> prev;
+
+        public Node(Node<T> prev, T element, Node<T> next) {
```
RomanDubovskyi, line 16:
```suggestion
        Node(Node<T> prev, T element, Node<T> next) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,136 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+
+    private int size;
+    private Node<T> head;
+    private Node<T> tail;
+
+    private static class Node<T> {
```
RomanDubovskyi, line 11: let's move private class declaration to the end of the file

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,136 @@
...
+            node.next.prev = node.prev;
+        }
+        size--;
+        return node.item;
     }
 
     @Override
     public boolean remove(T t) {
+        for (int i = 0; i < size; i++) {
+            if (getNode(i).item == t || getNode(i).item.equals(t)) {
```
RomanDubovskyi, line 104: what if getNode(i).item will be null?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,136 @@
...
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private Node<T> getNode(int index) {
+        checkIndex(index);
+        Node<T> node = head;
+
+        for (int i = 0; i < index; i++) {
```
RomanDubovskyi, line 126: let's iterate our linkedlist depending on location of our node:
* if node is closer to the start, we will go from first to middle, if it’s closer to the end of list we will go from last to middl

src/main/java/core/basesyntax/MyLinkedList.java
yutsevich: first check is for situation when one of comparable items is null

PR #154, opened 2020-08-10
Hw9 old

vremere-a:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,145 @@
...
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException();
+        }
+        return getNode(index).element;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        Node<T> x = getNode(index);
```
Sofasmile, line 54: let's rename variable `x`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,145 @@
...
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
+        return true;
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index < 0 || index > size) {
```
Sofasmile, line 46: do we need this checking here?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,145 @@
...
     public T remove(int index) {
-        return null;
+        Node<T> node = getNode(index);
+        if (size == 1) {
+            lastNode = null;
+            firsNode = null;
+            size--;
+            return node.element;
+        }
+        if (index == 0) {
```
Sofasmile, line 69: let's use else-if and then we can move 
```
size--;
return node.element;
```
at the end of the method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,145 @@
...
+        node.prev.next = node.next;
+        node.next.prev = node.prev;
+        size--;
+        return node.element;
     }
 
     @Override
     public boolean remove(T t) {
+        for (int i = 0; i < size; i++) {
+            if (t == getNode(i).element || getNode(i).element.equals(t)) {
```
Sofasmile, line 90: let's use getNode(i) only one time
what happens if getNode(i) == null?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,145 @@
...
+        } else {
+            Node<T> newNode = new Node<>(lastNode, value, null);
+            lastNode.next = newNode;
+            lastNode = newNode;
+        }
+        size++;
+    }
+
+    // определяется узел, который находиться под нужным индексом
+    Node<T> getNode(int index) {
```
Sofasmile, line 135: - in this method let's start moving from the first Entry OR from the last one
- add access modifier

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,146 @@
...
+            while (i < index) {
+                node = node.next;
+                i++;
+            }
+            return node;
+        }
     }
 }
+
+
```
boroda4436, line 148: @vremere-a redundant empty lines

vremere-a:
del

PR #160, opened 2020-08-10
solution

Vas35:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,141 @@
...
+    private Node<T> last;
+    private int size;
+
     @Override
     public boolean add(T value) {
-        return false;
+        if (isEmpty()) {
+            Node<T> node = new Node<>(value, null, null);
+            first = node;
+            last = node;
```
Sofasmile, line 15: let's move `last = node;` after if-else construction

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,141 @@
...
+        return true;
     }
 
     @Override
     public T get(int index) {
-        return null;
+        return getNode(index).value;
+    }
+
+    private Node<T> getNode(int index) {
```
Sofasmile, line 59: let's move all private methods after public

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,141 @@
...
+        if (index < 0 || index >= size) {
+            throw new ArrayIndexOutOfBoundsException(
+                    "This index is not suitable for the given size");
+        }
+        int i = 0;
+        Node<T> node = first;
+        while (i < index) {
```
Sofasmile, line 66: in this method let's start moving from the first Entry OR from the last one

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,141 @@
...
+        node.value = value;
+        return oldValue;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        Node<T> node = getNode(index);
+        if (size == 1) {
+
```
Sofasmile, line 85: remove empty line

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,141 @@
...
+            first = null;
+            last = null;
+            size--;
+            return node.value;
+        }
+        if (index == size - 1) {
```
Sofasmile, line 91: let's use else-if and then we can move 
```
size--;
return node.value;
```
at the end of the method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,141 @@
...
+        }
+        if (index == 0) {
+            node.next.prev = null;
+            first = node.next;
+            size--;
+            return node.value;
+        }
+        node.prev.next = node.next;
+        node.next.prev = node.prev;
+        size--;
```
Sofasmile, line 105: better move all this code to `unlink(Node<T> node)` method and call it here and in the remove(T t)

src/main/java/core/basesyntax/MyLinkedList.java
Vas35: But why? I can write like this first = last = node;

Vas35: Or do you mean initialize in the constructor last and first?

src/main/java/core/basesyntax/MyLinkedList.java
Vas35: ok

src/main/java/core/basesyntax/MyLinkedList.java
Vas35: What do you mean? I start moving from the first Entry.

src/main/java/core/basesyntax/MyLinkedList.java
Vas35: I did

src/main/java/core/basesyntax/MyLinkedList.java
Vas35: OK

src/main/java/core/basesyntax/MyLinkedList.java
Vas35: I broke the program ( Give more explanation.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,138 @@
...
+        return node.value;
     }
 
     @Override
     public boolean remove(T t) {
+
```
boroda4436, line 94: empty line after method body started is redundant

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,138 @@
...
+        private Node<T> prev;
+
+        Node(T value, Node<T> next, Node<T> prev) {
+            this.value = value;
+            this.next = next;
+            this.prev = prev;
+        }
     }
 }
+
```
boroda4436, line 140: only one empty line at the OEF is required

src/main/java/core/basesyntax/MyLinkedList.java
Vas35: I did, thanks.

src/main/java/core/basesyntax/MyLinkedList.java
Vas35: Thanks, I made it.

src/main/java/core/basesyntax/MyLinkedList.java
Vas35: You told me left like that.

PR #162, opened 2020-08-10
Done HW-9

Dubovaya:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,137 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size = 0;
```
Sofasmile, line 6:
```suggestion
    private int size;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,137 @@
...
+        return true;
     }
 
     @Override
     public T get(int index) {
-        return null;
+        return getNode(index).value;
+    }
+
+    private Node<T> getNode(int index) {
```
Sofasmile, line 58: let's move all private methods after public

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,137 @@
...
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException();
+        }
+
+        int i = 0;
```
Sofasmile, line 63: in this method let's start moving from the first Entry OR from the last one

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,137 @@
...
+            return node.value;
+        }
+        if (index == 0) {
+            node.next.prev = null;
+            head = node.next;
+            size--;
+            return node.value;
+        }
+        node.prev.next = node.next;
+        node.next.prev = node.prev;
```
Sofasmile, line 102: let's move this code at separate method `unlink(Node<T> node)` and call it here and at the remove(T t)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,137 @@
...
+        size--;
+        return node.value;
     }
 
     @Override
     public boolean remove(T t) {
+        for (int i = 0; i < size; i++) {
+            if (t == get(i) || get(i).equals(t)) {
```
Sofasmile, line 110: - what does happen if get(i) == null?
- let's call method get(i) only one time

src/main/java/core/basesyntax/MyLinkedList.java
Dubovaya: Agry

src/main/java/core/basesyntax/MyLinkedList.java
Dubovaya: ok. moved it to the end & after privet class Node<T>

Dubovaya:
made it more elegance

src/main/java/core/basesyntax/MyLinkedList.java
Dubovaya: didn't understand... make : 1. size-1
                                              2.0
                                              2.1
?

src/main/java/core/basesyntax/MyLinkedList.java
Dubovaya: ok--> unlink(Node<T> current)

PR #172, opened 2020-12-09
Implemented class MyLinkedList and overrode  all methods

YuraMitryakhin:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,151 @@
...
     @Override
     public int size() {
-        return 0;
+        return linkedListSize;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return linkedListSize == 0 ? true : false;
```
dmytro-hryhoruk, line 139:
```suggestion
        return linkedListSize == 0;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,151 @@
...
     @Override
     public T remove(int index) {
-        return null;
+        Node<T> findNode = head;
+        for (int i = 0; i < linkedListSize; i++) {
+            if (i == index) {
+                delete(findNode);
+                return findNode.item;
+            }
+            findNode = findNode.next;
```
dmytro-hryhoruk, line 97: I would remove finding the node by index to separate method since you are using it in remove(), get() set() methods

dmytro-hryhoruk (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,135 @@
...
+
+    public MyLinkedList() {
+        linkedListSize = 0;
+    }
+
     @Override
     public boolean add(T value) {
-        return false;
+        if (linkedListSize == 0) {
+            Node<T> firstNode = new Node<>(null, value, null);
```
Sofasmile, line 17: let's move the creation of new node before if-else block

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,135 @@
...
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index == linkedListSize) {
+            add(value);
+            return;
+        }
+        Node<T> findNode = findNode(index);
```
Sofasmile, line 35:
```suggestion
        Node<T> currentNode = findNode(index);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,135 @@
...
+        } else {
+            tail = prevNode;
+        }
+        linkedListSize--;
+        return true;
+    }
+
+    private Node<T> findNode(int index) {
+        Node<T> findNode = head;
+        for (int i = 0; i < linkedListSize; i++) {
```
Sofasmile, line 117: if the node is at the second part of the list, let's start moving from the tail

Sofasmile (changes requested):
let's fix all comments)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -113,12 +113,23 @@ private boolean delete(Node<T> node) {
     }
 
     private Node<T> findNode(int index) {
```
Sofasmile, line 115: we can simplify like this:
```
        if (index < 0 || index >= linkedListSize) {
            throw new IndexOutOfBoundsException("Index [" + index + "] does not exist!");
        }
        Node<T> findNode;
        if (index < linkedListSize / 2) {
            findNode = head;
                findNode = findNode.next;
            }
        } else {
            findNode = tail;
            for (int i = linkedListSize - 1; i > index; i--) {
                findNode = findNode.prev;
            }
        }
        return findNode;
```

Sofasmile (approved):

PR #173, opened 2020-12-09
Implemented MyLinkedList and all methods

Vedroid:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,207 @@
...
     @Override
     public boolean add(T value) {
-        return false;
+        if (size == 0) {
+            Node<T> newNode = new Node<>(null, value, null);
+            head = newNode;
+            tail = newNode;
+            size++;
+            return true;
+        } else {
```
dmytro-hryhoruk, line 36: don't have to add else after return

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,207 @@
...
+            Node<T> currentNode = head;
+            while (true) {
+                if (currentNode.next == null) {
+                    Node<T> newNode = new Node<>(currentNode, value, null);
+                    currentNode.next = newNode;
+                    tail = newNode;
+                    size++;
+                    return true;
+                }
+                currentNode = currentNode.next;
```
dmytro-hryhoruk, line 46: you should add the Node at the end of list. You have head and tail, noo reason to iterate

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,207 @@
...
+    
+    public T getLast() {
+        return tail.value;
+    }
+    
+    private Node<T> getNode(int index) {
+        Node<T> currentNode = head;
+        int i = 0;
+        while (currentNode != null) {
+            if (i == index) {
```
dmytro-hryhoruk, line 189: What if your node would be the last before tail? Probably in some cases, it's better to start from tail?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,207 @@
...
+                if (nextNode == null) {
+                    prevNode.next = null;
+                    tail = prevNode;
+                    size--;
+                    return true;
+                }
+                nextNode.prev = prevNode;
+                prevNode.next = nextNode;
+                size--;
+                return true;
```
dmytro-hryhoruk, line 160: let's move this to a separate method that will unlink neighbor nodes and use it here when the node is found

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,207 @@
...
+                if (nextNode == null) {
+                    prevNode.next = null;
+                    tail = prevNode;
+                    size--;
+                    return currentNode.value;
+                }
+                nextNode.prev = prevNode;
+                prevNode.next = nextNode;
+                size--;
+                return currentNode.value;
```
dmytro-hryhoruk, line 123: that's repeating logic from remove(T value). Here you can use getNode()and then unlink method

dmytro-hryhoruk (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,183 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+    
+    public MyLinkedList() {
+        head = null;
```
Sofasmile, line 11: if you want you can remove initialization from the constructor because all these values are default

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,183 @@
...
+    
     @Override
     public boolean add(T value) {
-        return false;
+        if (size == 0) {
+            Node<T> newNode = new Node<>(null, value, null);
+            head = newNode;
+            tail = newNode;
+            size++;
+            return true;
```
Sofasmile, line 35: better move there lines after if-else at the end of the method)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,183 @@
...
+        if (nextNode == null) {
+            prevNode.next = null;
+            tail = prevNode;
+            size--;
+            return currentNode;
+        }
+        nextNode.prev = prevNode;
+        prevNode.next = nextNode;
+        size--;
+        return currentNode;
```
Sofasmile, line 171:
```suggestion
        if (prevNode == null) {
            head = nextNode;
        } else {
            prevNode.next = nextNode;
            currentNode.prev = null;
        }
        if (nextNode == null) {
            tail = prevNode;
        } else {
            nextNode.prev = prevNode;
            currentNode.next = null;
        }
        size--;
        return currentNode;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,183 @@
...
+            int i = size - 1;
+            while (currentNode != null) {
+                if (i == index) {
+                    return currentNode;
+                }
+                i--;
+                currentNode = currentNode.prev;
+            }
+        }
+        return new Node<>(null, null, null);
```
Sofasmile, line 150:
```suggestion
        if (index < size / 2) {
            Node<T> currentNode = head;
            int i = 0;
            while (i < index) {
                i++;
                currentNode = currentNode.next;
            }
            return currentNode;
        } else {
            Node<T> currentNode = tail;
            int i = size - 1;
            while (i > index) {
                i--;
                currentNode = currentNode.prev;
            }
            return currentNode;
        }
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,183 @@
...
     @Override
     public T remove(int index) {
-        return null;
+        checkIncludeIndex(index);
+        if (size == 1 && index == 0) {
+            head = null;
+            tail = null;
+            size = 0;
+            return null;
+        }
```
Sofasmile, line 94: let's remove this code

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,183 @@
...
+        Node<T> currentNode = head;
+        while (currentNode != null) {
+            if (currentNode.value == object
+                    || (currentNode.value != null && currentNode.value.equals(object))) {
+                if (size == 1) {
+                    head = null;
+                    tail = null;
+                    size = 0;
+                    return true;
+                }
```
Sofasmile, line 109: remove this checking

Sofasmile (changes requested):

Sofasmile (approved):
great)

PR #177, opened 2020-12-10
added solution

ivlytvynets:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,166 @@
...
     public T get(int index) {
-        return null;
+        return getNodeByIndex(index).item;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        T oldItem;
+        if (isIndexValid(index)) {
```
boroda4436, line 70: This check can be moved to `getNodeByIndex(index)` method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,166 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private static class Node<T> {
```
boroda4436, line 124: this class can be moved after all methods, not between them

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,166 @@
...
+            this.next = next;
+        }
+    }
+
+    private boolean isIndexValid(int index) {
+        return index >= 0 && index < size;
+    }
+
+    private Node<T> getNodeByIndex(int index) {
+        if (!isIndexValid(index)) {
```
boroda4436, line 141: better let's create method `checkIndexIsValid(index)` that will throw an exception if the index is not valid

boroda4436 (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,162 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private boolean isIndexValid(int index) {
```
boroda4436, line 120: where are you using this method? Also, this method can be void

boroda4436 (changes requested):

boroda4436 (approved):

PR #184, opened 2020-12-11
Overridden MyLinkedListInterface Interface methods, created helper methods

Likh-Alex:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,152 @@
...
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    public MyLinkedList() {
+        this.head = null;
+        this.tail = null;
+        this.size = 0;
```
RomanDubovskyi, line 13: these are default values, we don't need to set them explicitly, co you can remove constructor

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,152 @@
...
+        } else {
+            predecessor.next = newNode;
+        }
+        size++;
+    }
+
+    private void linkToTail(T elementToInsert) {
+        Node<T> predecessor = this.tail;
+        Node<T> newNode = new Node<>(predecessor, elementToInsert, null);
+        this.tail = newNode;
```
RomanDubovskyi, line 47:
```suggestion
        tail = newNode;
```
keyword 'this' are redundant here

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,152 @@
...
+        if (predecessor == null) {
+            head = newNode;
+        } else {
+            predecessor.next = newNode;
+        }
+        size++;
+    }
+
+    private Node<T> getNodeByIndex(int index) {
```
RomanDubovskyi, line 56: let's iterate LL from the head to the middle or from the tail to the middle depending on index, that will be faster

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,152 @@
...
+        }
+        return targetNode;
+    }
+
+    private T removeNode(Node<T> nodeToRemove) {
+        Node<T> beforeTarget = nodeToRemove.prev;
+        Node<T> afterTarget = nodeToRemove.next;
+        if (afterTarget == null) {
+            tail = beforeTarget;
+            nodeToRemove.prev = null;
```
RomanDubovskyi, line 69: if nodeToRemove next == null we should also make beforeTarget.next = null cause now it is still pointing at the nodeToRemove. Same situation with the head

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,152 @@
...
+        linkToTail(value);
+        return true;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index == size) {
+            linkToTail(value);
+        } else if (index < size && index >= 0) {
```
RomanDubovskyi, line 91: can we here use isInvalidIndex instead?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,152 @@
...
+            linkToNode(value, getNodeByIndex(index));
+        } else {
```
RomanDubovskyi, line 93: move throwing of exception into if block, so you won't have additional else clause

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,152 @@
...
+        if (isInvalidIndex(index)) {
+            throw new IndexOutOfBoundsException(index + " index is out of bounds");
+        }
+        return removeNode(getNodeByIndex(index));
     }
 
     @Override
     public boolean remove(T object) {
+        for (int i = 0; i < size; i++) {
+            T targetValue = getNodeByIndex(i).value;
```
RomanDubovskyi, line 136: you use nested loop (one here, second in getNodeByIndex) can we do it with single loop? 
that decrease algorithmic complexity

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,152 @@
...
+
+    private Node<T> getNodeByIndex(int index) {
+        Node<T> targetNode = head;
+        for (int i = 0; i < index; i++) {
+            targetNode = targetNode.next;
+        }
+        return targetNode;
+    }
+
+    private T removeNode(Node<T> nodeToRemove) {
```
RomanDubovskyi, line 64: group your methods by logic, if this method is related to remove they should near each other

RomanDubovskyi (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
Likh-Alex: done, removed the constructor and made this values default

src/main/java/core/basesyntax/MyLinkedList.java
Likh-Alex: done, removed

src/main/java/core/basesyntax/MyLinkedList.java
Likh-Alex: for simplification purpose removed variables `afterTarget` and `beforeTarget` - implemented addressing to `next` and `prev` directly through `nodeToRemove`. Refactored rebinding logic for `nodeToRemove` pointers.

src/main/java/core/basesyntax/MyLinkedList.java
Likh-Alex: done

src/main/java/core/basesyntax/MyLinkedList.java
Likh-Alex: used `while` loop instead, checking until `node` is `null`

src/main/java/core/basesyntax/MyLinkedList.java
Likh-Alex: done

src/main/java/core/basesyntax/MyLinkedList.java
Likh-Alex: done

src/main/java/core/basesyntax/MyLinkedList.java
Likh-Alex: done

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,157 @@
...
+            }
+        }
+        return targetNode;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        if (isInvalidIndex(index)) {
+            throw new IndexOutOfBoundsException(index + " index is out of bounds");
```
RomanDubovskyi, line 103: since you always throw the same exception it would be better to move it to the method as to reduce code duplication

RomanDubovskyi (approved):

src/main/java/core/basesyntax/MyLinkedList.java
Likh-Alex: Done

PR #187, opened 2020-12-12
LL implementation

Artificial-Friend:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,156 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    int size;
```
Sofasmile, line 6: let's add access modifiers

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,156 @@
...
+    Node<T> head;
+    Node<T> tail;
+
+    public MyLinkedList() {
+        size = 0;
```
Sofasmile, line 11: if you want you can remove initialization from the constructor because `int` value by default has 0 value.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,156 @@
...
+        if (index == size) {
+            linkLast(value);
+        } else if (index == 0) {
+            linkFirst(value);
+        } else {
+            indexValidator(index);
+            Node<T> currentNode = head;
+            for (int i = 0; i < index; i++) {
+                currentNode = currentNode.next;
+            }
```
Sofasmile, line 32: can we use method getNode() here?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,156 @@
...
+            Node<T> newNode = new Node<>(tail, value, null);
+            tail.next = newNode;
+            tail = newNode;
+        }
+    }
+
+    private void unlinkNode(Node<T> node) {
+        if (node == tail) {
+            tail.next = null;
+            tail = node.previous;
```
Sofasmile, line 130: does it work correctly?
we should firstly do:
`tail = node.previous;`
and then:
`tail.next = null;`

Sofasmile (changes requested):

Sofasmile (approved):

PR #188, opened 2020-12-12
Added class Node and implementation methods

sanek98cv:
MyLinkedList class

### DIFF
--- a/src/main/java/core/basesyntax/MyLinkedList.java
+++ b/src/main/java/core/basesyntax/MyLinkedList.java
@@ -3,48 +3,162 @@ package core.basesyntax;
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+
+    private static class Node<T> {
+        private T element;
+        private Node<T> first;
+        private Node<T> last;
+
+        public Node(Node<T> first, T element, Node<T> last) {
+            this.element = element;
+            this.first = first;
+            this.last = last;
+        }
+    }
+
+    private Node<T> head;
+    private Node<T> tail;
+    private int sizeList = 0;
+
     @Override
     public boolean add(T value) {
-        return false;
+        Node<T> lastNode = tail;
+        Node<T> newNode = new Node<>(lastNode, value, null);
+        tail = newNode;
+        if (lastNode == null) {
+            head = newNode;
+            sizeList++;
+        } else {
+            lastNode.last = newNode;
+            sizeList++;
+        }
+        return true;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index > sizeList || index < 0) {
+            throw new IndexOutOfBoundsException("Incorrect index");
+        }
+        if (index == sizeList) {
+            add(value);
+        } else {
+            linkBefore(value, nodeIndex(index));
+        }
     }
 
     @Override
     public boolean addAll(List<T> list) {
-        return false;
+        for (int i = 0; i < list.size(); i++) {
+            T value = list.get(i);
+            add(value);
+        }
+        return true;
     }
 
     @Override
     public T get(int index) {
-        return null;
+        isIndexExist(index);
+        return nodeIndex(index).element;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        isIndexExist(index);
+        Node<T> currentIndex = nodeIndex(index);
+        T oldVal = currentIndex.element;
+        currentIndex.element = value;
+        return oldVal;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        isIndexExist(index);
+        return position(nodeIndex(index));
     }
 
     @Override
     public boolean remove(T object) {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,162 @@
...
+        public Node(Node<T> first, T element, Node<T> last) {
+            this.element = element;
+            this.first = first;
+            this.last = last;
+        }
+    }
+
+    private Node<T> head;
+    private Node<T> tail;
+    private int sizeList = 0;
```
RomanDubovskyi, line 21: in the class order elements this way : fields -> init blocks -> constructors -> methods / inner class

```suggestion
    private int size = 0;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,162 @@
...
+
     @Override
     public boolean add(T value) {
-        return false;
+        Node<T> lastNode = tail;
+        Node<T> newNode = new Node<>(lastNode, value, null);
+        tail = newNode;
+        if (lastNode == null) {
+            head = newNode;
+            sizeList++;
```
RomanDubovskyi, line 30: this line is present in if and else, so you can move it outside if else block

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,162 @@
...
     public T remove(int index) {
-        return null;
+        isIndexExist(index);
+        return position(nodeIndex(index));
     }
 
     @Override
     public boolean remove(T object) {
+        if (object == null) {
+            for (Node<T> newNode = head; newNode != null; newNode = newNode.last) {
```
RomanDubovskyi, line 83: let's use a single loop independently if the object is null or not. You can compare it with null and non null values without risk of NPE if you write correct if condition

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,162 @@
...
+        if (prev == null) {
+            head = newNode;
+            sizeList++;
+        } else {
+            prev.last = newNode;
+            sizeList++;
+        }
+    }
+
+    T position(Node<T> currentIndex) {
```
RomanDubovskyi, line 144: Let's try to improve naming: 
is input param index or it's a node? 
make method name a verb

RomanDubovskyi (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,152 @@
...
+            }
+        } else {
+            for (int i = size - 1; i > index; i--) {
+                newNodeTail = newNodeTail.first;
+            }
+        }
+        return index < (size >> 1) ? newNodeHead : newNodeTail;
+    }
+
+    void linkBefore(T item, Node<T> current) {
```
RomanDubovskyi, line 109: what about access modifier? 
```suggestion
   private void linkBefore(T item, Node<T> current) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,152 @@
...
+        if (prev == null) {
+            head = newNode;
+            size++;
+        } else {
+            prev.last = newNode;
+            size++;
+        }
+    }
+
+    T position(Node<T> nodePosition) {
```
RomanDubovskyi, line 122:
```suggestion
    T unlink(Node<T> nodePosition) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,152 @@
...
+        return index < (size >> 1) ? newNodeHead : newNodeTail;
+    }
+
+    void linkBefore(T item, Node<T> current) {
+        Node<T> prev = current.first;
+        Node<T> newNode = new Node<>(prev, item, current);
+        current.first = newNode;
+        if (prev == null) {
+            head = newNode;
+            size++;
```
RomanDubovskyi, line 115: size++ is present in if and else clause, move it outside if else so you won't need to write it twice

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,152 @@
...
+        return size == 0;
+    }
+
+    private void isIndexExist(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Element can't be found!");
+        }
+    }
+
+    Node<T> nodeIndex(int index) {
```
RomanDubovskyi, line 94: make method name verb and don't forget about access modifiers

RomanDubovskyi (changes requested):

RomanDubovskyi (approved):

PR #194, opened 2020-12-13
Implement MyLinkedList

Resci:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,159 @@
...
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node head;
+    private Node tail;
+    private int size;
+
+    public MyLinkedList() {
+        size = 0;
+        head = null;
+        tail = null;
```
RomanDubovskyi, line 13: these values will be set by default

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,159 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node head;
```
RomanDubovskyi, line 6:
```suggestion
    private Node<T> head;
```
same with others

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,159 @@
...
+        tail = node;
+        size++;
+        return true;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (!(checkIndex(index) || index == size)) {
+            throw new IndexOutOfBoundsException("Failed add to LinkedList. Please check index");
```
RomanDubovskyi, line 32: this code is duplicated, le'ts move it to the separate method and call it instead

ivlytvynets, line 32: throw this exception in method checkIndex()

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,159 @@
...
+            add(list.get(i));
+        }
+        return true;
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (!checkIndex(index)) {
+            throw new IndexOutOfBoundsException("Failed get from LinkedList. Please check index");
```
ivlytvynets, line 60: don't repeat yourself. the same moment

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,159 @@
...
+        } else if (node == tail) {
+            node.previous.next = null;
+            tail = node.previous;
+        } else if (node == head) {
+            node.next.previous = null;
+            head = node.next;
+        } else {
+            node.previous.next = node.next;
+            node.next.previous = node.previous;
+        }
```
RomanDubovskyi, line 95: Try to simplify it:
```suggestion
     if (node.prev == null) {
            first = node.next;
        } else {
            node.prev.next = node.next;
        }
        if (node.next == null) {
            last = node.prev;
        } else {
            node.next.prev = node.prev;
        }
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,159 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private class Node {
```
RomanDubovskyi, line 123: don't use raw class

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,159 @@
...
+        private Node next;
+
+        public Node(Node previous, T value, Node next) {
+            this.previous = previous;
+            this.value = value;
+            this.next = next;
+        }
+    }
+
+    private Node findIndex(int index) {
```
RomanDubovskyi, line 135:
```suggestion
    private Node findByIndex(int index) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,159 @@
...
     public boolean remove(T object) {
-        return false;
+        if (isEmpty()) {
+            throw new IndexOutOfBoundsException("Cannot delete from empty list");
+        }
+        int index = getIndex(object);
+        if (index == -1) {
+            return false;
+        }
+        remove(index);
```
RomanDubovskyi, line 109: you search index by value and call remove(index). In remove you search node by index, so that's 2 loops instead of 1

src/main/java/core/basesyntax/MyLinkedList.java
RomanDubovskyi, line 135: let's iterate our LL from head to middle of tail to middle depending on the index

RomanDubovskyi (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,50 +1,162 @@
...
+
+    private Node<T> findByValue(T value) {
+        Node<T> node = head;
+        while (node != null) {
+            if (node.value == value || (node.value != null && node.value.equals(value))) {
+                return node;
+            }
+            node = node.next;
+        }
+        throw new NoSuchElementException("No such element in list");
```
RomanDubovskyi, line 160: we don't need exception here, if you use it for remove just return boolean and you won't need to use try catch

RomanDubovskyi (changes requested):

PR #198, opened 2020-12-13
added LL implementation

oroboro-cpu:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,150 @@
...
+        node.prev = newNode;
+        if (nodeBefore == null) {
+            head = newNode;
+        } else {
+            nodeBefore.next = newNode;
+        }
+        size++;
+    }
+
+    public Node<T> getNode(int index) {
```
RomanDubovskyi, line 68: should this method be public ?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,150 @@
...
+        checkSize(index);
+        Node<T> node = head;
+        for (int i = 0; i < index; i++) {
```
RomanDubovskyi, line 71: let's iterate LL from head to the middle or from tail to the middle depending on the index

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,150 @@
...
+            head = node.next;
+            head.prev = null;
+        } else {
+            node.prev.next = node.next;
+            node.next.prev = node.prev;
+        }
+        size--;
+    }
+
+    public void checkSize(int index) {
```
RomanDubovskyi, line 91: same here with access modifier
check your methods, if you use them only inside this class - make it private

RomanDubovskyi (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,158 @@
...
+        addTail(value);
+        return true;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Wrong index!");
+        }
+        if (index == size) {
```
RomanDubovskyi, line 37: BTW you could have put this check first and then just use checkSize()

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,158 @@
...
+        unlink(node);
+        return node.value;
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> removeNode = head;
+        while (removeNode != null) {
+            if ((object == null && removeNode.value == null) || (removeNode.value != null
+                    && removeNode.value.equals(object))) {
```
RomanDubovskyi, line 142:
```suggestion
            if (object == removeNode.value || (removeNode.value != null
                    && removeNode.value.equals(object))) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,158 @@
...
+        } else {
+            for (int i = 0; i < (size - 1) - index; i++) {
+                nodeFromTail = nodeFromTail.prev;
+            }
+            return nodeFromTail;
+        }
+    }
+
+    private void unlink(Node<T> node) {
+        if (node == tail) {
```
RomanDubovskyi, line 86:
```suggestion
        if (node.next == null) {
```
it's a bad practice to compare objects with ==. 
This way you can know if node is tail

RomanDubovskyi (changes requested):
added comments on simplification, rest is good

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -31,14 +31,12 @@ public boolean add(T value) {
...
     public void add(T value, int index) {
-        if (index < 0 || index > size) {
-            throw new IndexOutOfBoundsException("Wrong index!");
-        }
         if (index == size) {
             addTail(value);
         } else {
             addHead(value, getNode(index));
         }
+        checkSize(index);
```
RomanDubovskyi, line 39: if (index == size) {
            addTail(value);
            return;
        }
        checkElementIndex(index);
        addHead(value, getNode(index));

RomanDubovskyi (approved):

PR #200, opened 2020-12-14
LinkedList implementation

less08:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,124 @@
...
     }
 
     @Override
     public void add(T value, int index) {
+        if (index == size) {
+            add(value);
+            return;
+        }
 
+        Node<T> oldNode = getNodeByIndex(index);
```
Sofasmile, line 43:
```suggestion
        Node<T> currentNode = getNodeByIndex(index);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,124 @@
...
+        } else {
+            head = newNode;
+        }
+        size++;
     }
 
     @Override
     public boolean addAll(List<T> list) {
-        return false;
+        for (T t : list) {
```
Sofasmile, line 57:
```suggestion
        for (T value : list) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,124 @@
...
+        } else {
+            head = temp.next;
+        }
+        if (temp.next != null) {
+            temp.next.prev = temp.prev;
+        } else {
+            tail = temp.prev;
+        }
+        size--;
+        return temp.value;
```
Sofasmile, line 90: let's move this logic to the separate method delete()/unlink() and use it here and in the method remove(T object)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,124 @@
...
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private Node<T> getNodeByIndex(int index) {
+        if (index >= size || index < 0) {
+            throw new IndexOutOfBoundsException("");
+        }
+        Node<T> temp = head;
```
Sofasmile, line 120: What if the index is closer to the tail? let's start moving from the tail in this case

Sofasmile (changes requested):
great job))

Sofasmile (approved):

PR #205, opened 2020-12-14
first commit

entertainment-veks:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,204 @@
...
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    public MyLinkedList() {
+        head = null;
+        tail = null;
+        size = 0;
```
Sofasmile, line 14: if you want you can remove initialization from the constructor because `int` value by default has 0 value and objects are null

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,204 @@
...
+            this.prev = prev;
+        }
+
+        public Node<T> getNext() {
+            return next;
+        }
+
+        public void setNext(Node<T> next) {
+            this.next = next;
+        }
```
Sofasmile, line 203: do we need getters and setters in the inner class?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,204 @@
...
 
     @Override
     public void add(T value, int index) {
+        if (index == size) { //add on tail
+            add(value);
+            return;
+        }
+
+        Node<T> nodeBefore = null;
+        Node<T> nodeAfter = head;
```
Sofasmile, line 38: do we need these variables?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,204 @@
...
     public T get(int index) {
-        return null;
+        if (isIndexInvalid(index)) {
+            throw new IndexOutOfBoundsException();
+        }
+
+        Node<T> result = head;
+        for (int i = 0; i < index; i++) {
+            result = result.getNext();
+        }
```
Sofasmile, line 79: can we move this logic to the separate private method? and use it in the methods: remove(), get(), set(), addByIndex()

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,204 @@
...
+            settedNode = settedNode.getNext();
+        }
+
+        T oldValue = settedNode.getValue();
+        settedNode.setValue(value);
+        return oldValue;
     }
 
     @Override
     public T remove(int index) {
```
Sofasmile, line 100: let's move part of the logic to the separate method delete/unlink and use this method in the methods remove by index and remove by value

Sofasmile (changes requested):
let's fix comments)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,164 @@
...
+        } else {
+            tail.next = newNode;
+        }
+        tail = newNode;
+        size++;
+        return true;
     }
 
     @Override
     public void add(T value, int index) {
```
Sofasmile, line 25: better remove all empty lines from method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,164 @@
...
     public T get(int index) {
-        return null;
+        if (isIndexInvalid(index)) {
+            throw new IndexOutOfBoundsException();
+        }
+        return findNodeByIndex(index).value;
     }
 
     @Override
     public T set(T value, int index) {
```
Sofasmile, line 68: let's remove all empty line from method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,164 @@
...
+
+        Node<T> settedNode = findNodeByIndex(index);
+
+        T oldValue = settedNode.value;
+        settedNode.value = value;
+        return oldValue;
     }
 
     @Override
     public T remove(int index) {
```
Sofasmile, line 81: how about this comment?
we can create method unlinkNode()/deleteNode()
```
    private T unlink(Node<T> currentNode){
        Node<T> prev = currentNode.prev;
        Node<T> next = currentNode.next;
        if (prev == null) {
            head = next;
        } else {
            prev.next = next;
        }
        if (next == null) {
            tail = prev;
        } else {
            next.prev = prev;
        }
        size--;
        return currentNode.value;
    }
```
and use it in the method remove by index:
```
        Node<T> removedNode = findNodeByIndex(index);
        return unlink(removedNode);
```
and method remove by value:
```
...
            if (currentValue == null ? currentValue == object : currentValue.equals(object)) {
                unlink(current);
                return true;
            }
...
```

Sofasmile (approved):

PR #206, opened 2020-12-14
implement MyLinkedList

IwalkInSilence:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,199 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    public MyLinkedList() {
+        size = 0;
```
Sofasmile, line 11: if you want you can remove initialization from the constructor because `int` value by default has 0 value.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,199 @@
...
     @Override
     public boolean isEmpty() {
+        if (head == null) {
+            return true;
+        }
         return false;
     }
+
+    private void linkFirst(T value) {
+        Node<T> h = head;
```
Sofasmile, line 93: h is not informative variable name

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,199 @@
...
-        return 0;
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
+        if (head == null) {
+            return true;
+        }
         return false;
```
Sofasmile, line 89:
```suggestion
        return size == 0;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,199 @@
...
+        if (h == null) {
+            tail = newNode;
+        } else {
+            h.prev = newNode;
+        }
+        size++;
+    }
+
+    private void linkLast(T value) {
+        Node<T> t = tail;
```
Sofasmile, line 105: the same here

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,199 @@
...
+        size++;
+    }
+
+    private void checkPositionIndex(int index) {
+        if (!isPositionIndex(index)) {
+            throw new IndexOutOfBoundsException("Index out of bounds " + index);
+        }
+    }
+
+    private boolean isPositionIndex(int index) {
```
Sofasmile, line 134: do we need this separate private method?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,199 @@
...
+        return index >= 0 && index <= size;
+    }
+
+    private void checkElementIndex(int index) {
+        if (!isElementIndex(index)) {
+            throw new IndexOutOfBoundsException("Index out of bounds " + index);
+        }
+    }
+
+    private boolean isElementIndex(int index) {
```
Sofasmile, line 144: the same with this method?

Sofasmile (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
Sofasmile: what about this comment?

Sofasmile (changes requested):
how about all comments?

Sofasmile (approved):

PR #217, opened 2021-04-06
Linked list implementation task done

mankokolya:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,179 @@
...
+    private int size;
+    private Node<T> head;
+    private Node<T> tail;
+
     @Override
     public boolean add(T value) {
-        return false;
+        if (head == null) {
+            head = new Node<>(null, value, null);
+            tail = new Node<>(null, value, null);
```
Max-Pochepets, line 14: here you create two different object, while in fact head and tail in list.size() == 1 are the same object

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,179 @@
...
+        } else {
+            if (head.next == null) {
```
Max-Pochepets, line 16: try to simplify this if else. you can check for size == 0 and assign head and tail due to situation

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,179 @@
...
+            addAtBeginning(value);
+            return;
+        }
+        addInMiddle(value, index);
     }
 
     @Override
     public boolean addAll(List<T> list) {
-        return false;
+        list.forEach(this::add);
```
Max-Pochepets, line 46: 👍

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,179 @@
...
+        return true;
     }
 
     @Override
     public void add(T value, int index) {
+        checkIndexIsValid(index, size, "Trying to add at nonexistent or negative index");
+        if (index == size) {
+            add(value);
+            return;
+        }
```
Max-Pochepets, line 36: empty line

Max-Pochepets (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,174 @@
...
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size;
+    private Node<T> head;
+    private Node<T> tail;
+
     @Override
     public boolean add(T value) {
-        return false;
+        Node<T> newNode = new Node<>(null, value, null);
```
KseniiaMakarova, line 12:
```suggestion
        Node<T> newNode = new Node<>(tail, value, null);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,174 @@
...
+        if (head == null) {
+            head = newNode;
+            tail = head;
+        } else {
+            tail.next = newNode;
+            newNode.previous = tail;
+            tail = newNode;
```
KseniiaMakarova, line 19: can we move `tail = newNode` from if-else? this line is the same in `if` and in `else`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,174 @@
...
+
+    private void checkIndexIsValid(int index, int size, String message) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException(message);
+        }
+    }
+
+    private Node<T> findNodeByIndex(int index) {
+        int count = 0;
+        Node<T> searchResult = head;
```
KseniiaMakarova, line 121: why are we always starting from head? let's start from tail if our index is closer to tail

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,174 @@
...
+        if (index == size - 1) {
+            return removeLastElement();
+        }
+        return removeFromMiddle(findNodeByIndex(index));
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> temp = head;
+        if (head.value == object || head.value.equals(object)) {
```
KseniiaMakarova, line 76: what if `head.value` is null and `object` is not null?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,174 @@
...
+            removeFirstElement();
+            return true;
+        }
+
+        while (temp != null) {
+            if (temp.value == null || temp.value.equals(object)) {
```
KseniiaMakarova, line 82: same here

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,174 @@
...
+        T oldValue = head.value;
+        if (nextToHead != null) {
+            nextToHead.previous = null;
+            head = nextToHead;
+        } else {
+            head = null;
+        }
+        size--;
+        return oldValue;
+    }
```
KseniiaMakarova, line 158: I would recommend to create a single `unlink()` method instead of these 3 methods

think about how we can do this - the method should work if we remove tail, or head, or the element from the middle

KseniiaMakarova (changes requested):

KseniiaMakarova (commented):
also, please create a PR from a separate branch, not from master

src/main/java/core/basesyntax/MyLinkedList.java
mankokolya: Here we will not get NullPointer as we check temp.value == null first

PR #223, opened 2021-04-07
LL implementation.

SakVladyslav:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,190 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    private static class Node<T> {
```
RomanDubovskyi, line 10: That's not critical but FYI:
usually private methods / inner classes are located at the end or near methods where you use them (if it's general/common and private -> to the end)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,190 @@
...
+            this.prev = prev;
+            this.item = item;
+            this.next = next;
+        }
+    }
+
     @Override
     public boolean add(T value) {
-        return false;
+        if (head == null) {
```
RomanDubovskyi, line 24: this method may be simplified, try smt like this: 

 - // create newNode object  like (head, value, null);
 - // if  your list is empty then tail = newNode, else head.next = newNode
 - // in any case head = newNode and size increase

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,190 @@
...
+        }
 
+        if (index == 0) {
+            addAtBeginning(value);
+        } else {
+            addByIndex(value, index);
+        }
+    }
+
+    private Node<T> findNodeByIndex(int index) {
```
RomanDubovskyi, line 58: for better performance it's better to iterate our list depending on index if it's between start and middle - begin from the start in other case begin from an end

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,190 @@
...
+            return removeLastElement();
+        }
+
+        return removeFromMiddle(findNodeByIndex(index));
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> headNode = head;
+        if (head.item == object || head.item.equals(object)) {
```
RomanDubovskyi, line 127: if head.item == null and object != null you will get NPE here. Let's take this case into consideration

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,190 @@
...
+
+        Node<T> nodeByIndex = findNodeByIndex(index);
+        T valueBeforeSet = nodeByIndex.item;
+        nodeByIndex.item = value;
+
+        return valueBeforeSet;
     }
 
     @Override
     public T remove(int index) {
```
RomanDubovskyi, line 112: create method unlink that will look like this and code will be simplified 
  private T unlink(Node<T> targetNode) {
        final T element = targetNode.item;
        Node<T> prev = targetNode.prev;
        Node<T> next = targetNode.next;

        if (prev == null) {
            first = next;
        } else {
            prev.next = next;
            targetNode.prev = null;
        }

        if (next == null) {
            last = prev;
        } else {
            next.prev = prev;
            targetNode.next = null;
        }

        size--;
        return element;
    }

RomanDubovskyi (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,50 +1,209 @@
...
+
     @Override
     public boolean add(T value) {
-        return false;
+        Node<T> newNode = new Node<>(tail, value, null);
+        if (head == null) {
+            head = newNode;
+        } else {
+            tail.next = newNode;
+            tail = newNode;
```
KseniiaMakarova, line 30: do you need this line in `else` (line 30)?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,50 +1,209 @@
...
+        }
+
+        if (index == 0) {
+            addAtBeginning(value);
+        } else {
+            addByIndex(value, index);
+        }
+    }
+
+    private Node<T> findNodeByIndex(int index) {
```
KseniiaMakarova, line 55: please move all private methods below public or at least below all usages of this private method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,50 +1,209 @@
...
+        if (index < size() / 2) {
```
KseniiaMakarova, line 56:
```suggestion
        if (index < size / 2) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,50 +1,209 @@
...
+        while (count < index) {
+            searchNode = searchNode.next;
+            count++;
+        }
+
+        return searchNode;
+    }
+
+    private Node<T> findNodeFromEnd(int index) {
+        int count = size() - 1;
```
KseniiaMakarova, line 75:
```suggestion
        int count = size - 1;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,50 +1,209 @@
...
+            return true;
+        }
+
+        while (headNode != null) {
+            if (headNode.item == null || headNode.item.equals(object)) {
+                removeFromMiddle(headNode);
+                return true;
+            }
+            headNode = headNode.next;
+        }
```
KseniiaMakarova, line 147:
```suggestion
        Node<T> headNode = head;
        while (headNode != null) {
            if (headNode.item == null || headNode.item.equals(object)) { // what if `headNode.item` is null and `object` is not null?
                unlink(headNode);
                return true;
            }
            headNode = headNode.next;
        }
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,50 +1,209 @@
...
+        size--;
+    }
+
+    private void removeFromMiddle(Node<T> nodeByIndex) {
+        Node<T> previous = nodeByIndex.prev;
+        Node<T> next = nodeByIndex.next;
+        previous.next = next;
+        next.prev = previous;
+        size--;
+    }
```
KseniiaMakarova, line 169: let's remove these methods and just use `unlink()` instead

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,50 +1,209 @@
...
+
+        if (next == null) {
+            tail = prev;
+        } else {
+            next.prev = prev;
+            targetNode.next = null;
+        }
+
+        size--;
+        return element;
```
KseniiaMakarova, line 201: set `targetNode.item` to null

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,50 +1,209 @@
...
+    }
+
+    private void checkIsIndexExist(int index, int size, String message) {
```
KseniiaMakarova, line 204: if the message is the same everywhere, maybe we should not pass it?
```suggestion
    private void checkIfIndexExists(int index, int size) { // or just `checkIndex()`
```

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):
great job)

PR #224, opened 2021-04-08
added solution

Dalai-S:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,185 @@
...
+            throw new IndexOutOfBoundsException("Index out of range");
+        }
+    }
+
+    private boolean isIndexInRange(int index) {
+        return index >= 0 && index < size;
+    }
+
+    private Node<T> getCurrentNode(int index) {
+        Node<T> current = head;
```
Sofasmile, line 136: let's firstly check which part of the list has an index
If the index is in the first part, then move from the first node to the index, otherwise - from the last node to the index.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,185 @@
...
+        return current.value;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        checkIndex(index);
+        T oldValue;
+        Node<T> current = getCurrentNode(index);
+        oldValue = current.value;
```
Sofasmile, line 72:
```suggestion
        T oldValue = current.value;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,185 @@
...
+
+        if (next == null) {
+            tail = previous;
+        } else {
+            next.previous = previous;
+            current.next = null;
+        }
+
+        current.value = null;
+        size--;
```
Sofasmile, line 101: let's move this code to the separate private method - unlink, and then we can use unlink method here and in the method remove(T object)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,185 @@
...
+        return deletedValue;
     }
 
     @Override
     public boolean remove(T object) {
+        int indexOfObject = findIndexByValue(object);
```
Sofasmile, line 107: do we need private method findIndexByValue()? better move all logic from method findIndexByValue() to method remove(T object)

Sofasmile (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,185 @@
...
+        final T deletedValue = current.value;
+        unlink(current);
+        return deletedValue;
     }
 
     @Override
     public boolean remove(T object) {
+        int i = 0;
+        Node<T> current = head;
+        while (i < size) {
```
Sofasmile, line 85:
```suggestion
        while (current != null) {
```

Sofasmile (changes requested):
good job)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -82,7 +82,7 @@ public T remove(int index) {
     public boolean remove(T object) {
         int i = 0;
```
Sofasmile, line 83: do we need int i variable?

Sofasmile (changes requested):

Sofasmile (approved):

PR #226, opened 2021-04-10
added first solution

IhorRome:


andriiromaniuk (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,146 @@
...
+        final Node<T> currentTail = last;
+        final Node<T> newNode = new Node<>(last, value, null);
+        last = newNode;
+        if (currentTail == null) {
+            first = newNode;
+        } else {
+            currentTail.next = newNode;
+        }
+        size++;
+        return true;
```
Sofasmile, line 34:
```suggestion
        final Node<T> newNode = new Node<>(last, value, null);
        if (last == null) {
            first = newNode;
        } else {
            last.next = newNode;
        }
        last = newNode;
        size++;
        return true;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,146 @@
...
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> lostNode = first;
+        while (lostNode != null) {
+            if (lostNode.item == object || lostNode.item != null && lostNode.item.equals(object)) {
+                unlink(lostNode);
+                return true;
+            } else {
```
Sofasmile, line 88: do we need else block here?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,146 @@
...
+        size++;
+        return true;
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index == size) {
+            add(value);
+        } else {
```
Sofasmile, line 41: we can add return after add() method and remove else block

Sofasmile (changes requested):
good job)

Sofasmile (approved):

KseniiaMakarova (approved):

PR #237, opened 2021-04-12
Overrided methods for LinkedList + implemented a few new ones.

Kharlanava:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,165 @@
...
-        return null;
+        checkElementIndex(index);
+        return unlink(getNode(index));
     }
 
     @Override
     public boolean remove(T object) {
+        for (Node<T> current = head; current != null; current = current.next) {
+            if (object == null && current.element == null
+                    || object != null && object.equals(current.element)) {
```
ritakisluhina, line 74: can we simplify it a little?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,165 @@
...
+                unlink(current);
+                return true;
+            }
+        }
```
freedmanMain, line 78:
```suggestion
        for (Node<T> current = head; current != null; current = current.next) {
            if (object == current.element
                    || object != null && object.equals(current.element)) {
                unlink(current);
                return true;
            }
        }
```Let's change this

File: src/main/java/core/basesyntax/MyLinkedList.java [hunk 3]
```diff
@@ -3,48 +3,165 @@
...
+            head = newNode;
+        }
+        size++;
+    }
+
+    private Node<T> getNode(int index) {
+        checkElementIndex(index);
+        if (index < (size / 2)) {
+            Node<T> newNode = head;
+            for (int i = 0; i < index; i++) {
```
ritakisluhina, line 120: we can reblace for with while

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,165 @@
...
 
     @Override
     public T remove(int index) {
-        return null;
+        checkElementIndex(index);
+        return unlink(getNode(index));
     }
 
     @Override
     public boolean remove(T object) {
```
freedmanMain, line 71:
```suggestion
    public boolean remove(T item) {
```
Rename this.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,165 @@
...
+        checkElementIndex(index);
+        if (index < (size / 2)) {
+            Node<T> newNode = head;
+            for (int i = 0; i < index; i++) {
+                newNode = newNode.next;
+            }
+            return newNode;
+        }
+        Node<T> newNode = tail;
+        for (int i = size - 1; i > index; i--) {
```
ritakisluhina, line 126: same here

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,165 @@
...
+            newNode = newNode.previous;
+        }
+        return newNode;
```
freedmanMain, line 129:
```suggestion
        checkElementIndex(index);
        Node<T> newNode;
        if (index < (size / 2)) {
            newNode = head;
            for (int i = 0; i < index; i++) {
                newNode = newNode.next;
            }
        } else {
        newNode = tail;
        for (int i = size - 1; i > index; i--) {
            newNode = newNode.previous;
        } 
      }
     return newNode;
```We can make it easier.

File: src/main/java/core/basesyntax/MyLinkedList.java [hunk 7]
```diff
@@ -3,48 +3,165 @@
...
+    }
+
+    private void checkPositionIndex(int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Index is out of bounds");
+        }
+    }
```
ritakisluhina, line 136: do we need this method?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,165 @@
...
+            return newNode;
+        }
+        Node<T> newNode = tail;
+        for (int i = size - 1; i > index; i--) {
+            newNode = newNode.previous;
+        }
+        return newNode;
+    }
+
+    private void checkPositionIndex(int index) {
```
freedmanMain, line 132: Try to combine these into one method. You can pass the size value as an argument.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,165 @@
...
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Index is out of bounds");
+        }
+    }
+
+    private void checkElementIndex(int index) {
```
freedmanMain, line 138: Same.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,165 @@
...
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Index is out of bounds");
+        }
+    }
+
+    private T unlink(Node<T> currentNode) {
+        final T element = currentNode.element;
+        final Node<T> next = currentNode.next;
+        final Node<T> previous = currentNode.previous;
```
ritakisluhina, line 147: do we need this variables?

ritakisluhina (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
Kharlanava: we can, but it will work the same, won't it?

ritakisluhina: in while we don't need to create a new variable and it looks more readable. 
if you disagree you can have it like it is (:

src/main/java/core/basesyntax/MyLinkedList.java
Kharlanava: Yes, because without we can not make removed Node null and it will exist somewhere in our memory. In JDK LinkedList after changing links we should nullify Node that we want to unlink.
+ With this variables code is easier to read.

src/main/java/core/basesyntax/MyLinkedList.java
Kharlanava: Changed

src/main/java/core/basesyntax/MyLinkedList.java
Kharlanava: Done

src/main/java/core/basesyntax/MyLinkedList.java
Kharlanava: Yes, definitely! Thanks

src/main/java/core/basesyntax/MyLinkedList.java
Kharlanava: Done

src/main/java/core/basesyntax/MyLinkedList.java
Kharlanava: +

src/main/java/core/basesyntax/MyLinkedList.java
Kharlanava: Yes, thanks :)

src/main/java/core/basesyntax/MyLinkedList.java
ritakisluhina: what about garbage collector?

src/main/java/core/basesyntax/MyLinkedList.java
Kharlanava: One check as a separate method is left

ritakisluhina (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,161 @@
...
+        for (T object : list) {
+            add(object);
+        }
+        return true;
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkElementIndex(index);
```
Sofasmile, line 53: should we call checkElementIndex() here? we have already called this method in the getNode() method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,161 @@
...
+        return getNode(index).element;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        checkElementIndex(index);
```
Sofasmile, line 59: same here

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,161 @@
...
+        Node<T> currentNode = getNode(index);
+        T oldValue = currentNode.element;
+        currentNode.element = value;
+        return oldValue;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkElementIndex(index);
```
Sofasmile, line 68: and here

Sofasmile (changes requested):
good job)

Sofasmile (approved):

PR #241, opened 2021-04-12
LinkedList hw v.01

Reastador:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,139 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size = 0;
```
Ihor-Nykyforov, line 6:
```suggestion
    private int size;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,139 @@
...
+        Node<T> removedValue = node(index);
+        if (removedValue.next == null) {
+            tail = removedValue.prev;
+        } else if (removedValue.prev == null) {
+            head = removedValue.next;
+        } else {
+            removedValue.prev.next = removedValue.next;
+            removedValue.next.prev = removedValue.prev;
+        }
+        size--;
```
Ihor-Nykyforov, line 91: You should create separate method unlink(Node node)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,139 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size = 0;
+    private Node<T> head;
+    private Node<T> tail;
+
+    class Node<T> {
```
Ihor-Nykyforov, line 10:
```suggestion
    private static class Node<T> {
```

Ihor-Nykyforov (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,143 @@
...
     public boolean add(T value) {
-        return false;
+        if (size == 0) {
+            tail = new Node<>(null,value,null);
+            head = tail;
+        } else {
+            Node<T> newNode = new Node<>(tail,value,null);
+            tail.next = newNode;
+            tail = newNode;
+        }
```
KseniiaMakarova, line 31:
```suggestion
        Node<T> newNode = new Node<>(tail, value, null);
        if (size == 0) {
            head = newNode;
        } else {
            tail.next = newNode;
        }
        tail = newNode;
```
always try to move out all the code that repeats both in `if` and `else`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,143 @@
...
     }
+
+    private boolean checkIndex(int index) {
+        if (index >= size || index < 0) {
+            throw new IndexOutOfBoundsException("Index out of bound exception");
+        }
+        return true;
+    }
+
+    Node<T> node(int index) {
```
KseniiaMakarova, line 117: what about access modifier? also, let's use verbs for method names

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,143 @@
...
+        checkIndex(index);
+        Node<T> removedValue = node(index);
+        return unlink(removedValue);
     }
 
     @Override
     public boolean remove(T object) {
+        for (int i = 0; i < size; i++) {
+            if (node(i).value == object || object != null && object.equals(node(i).value)) {
+                remove(i);
```
KseniiaMakarova, line 90: let's introduce a node from which we start iterating

and when we find a match - let's just use `unlink(currentNode)`

because we don't want to call `checkIndex(index)` and `node(index)` (inside `remove()`) when it is not needed

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,143 @@
...
-        return 0;
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
+        if (size == 0) {
+            return true;
+        }
         return false;
```
KseniiaMakarova, line 107: how can we make this simpler?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,143 @@
...
     }
+
+    private boolean checkIndex(int index) {
+        if (index >= size || index < 0) {
+            throw new IndexOutOfBoundsException("Index out of bound exception");
+        }
+        return true;
```
KseniiaMakarova, line 114: do we need to return true here?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,143 @@
...
+    }
+
+    Node<T> node(int index) {
+        if (index < (size * 3 / 2)) {
```
KseniiaMakarova, line 118: why `* 3`?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,143 @@
...
+
+    private T unlink(Node<T> removedValue) {
+        if (removedValue.next == null) {
+            tail = removedValue.prev;
+        } else if (removedValue.prev == null) {
+            head = removedValue.next;
+        } else {
+            removedValue.prev.next = removedValue.next;
+            removedValue.next.prev = removedValue.prev;
+        }
```
KseniiaMakarova, line 141: this is not a 100% correct implementation at the moment, you should unlink from both sides in each case

let’s try to refactor this a bit
create 2 variables for `nextNode` and `previousNode`

first check if `nextNode` is null - write `if-else` construction

then check if `previousNode` is null - write another `if-else` construction

let’s also set node’s value to null in the end

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
Reastador: IDK :)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -64,32 +63,35 @@ public boolean addAll(List<T> list) {
...
 
     @Override
     public boolean remove(T object) {
+        Node<T> current = head;
         for (int i = 0; i < size; i++) {
-            if (node(i).value == object || object != null && object.equals(node(i).value)) {
-                remove(i);
+            if (current.value == object || (object != null
+                    && object.equals(current.value))) {
+                unlink(current);
```
Reastador, line 91: Here JDK offers to use Object.equals() - it's correct to change to this checking?

KseniiaMakarova: at the moment - we should not use `Object.equals()`
we should learn to write it by hands

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,149 @@
...
+        return getNode(index).value;
+
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        checkIndex(index);
+        T returnedValue = getNode(index).value;
+        getNode(index).value = value;
```
KseniiaMakarova, line 74: is it effective that we call `getNode(index)` twice?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,149 @@
...
+    }
+
+    private void checkIndex(int index) {
+        if (index >= size || index < 0) {
+            throw new IndexOutOfBoundsException("Index out of bound exception");
+        }
+    }
+
+    private Node<T> getNode(int index) {
+        if (index <= (size * 2)) {
```
KseniiaMakarova, line 116: why `size * 2`? what are you checking in this `if`?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,149 @@
...
+    }
+
+    private T unlink(Node<T> removedValue) {
+        Node<T> nextNode = removedValue.next;
+        Node<T> prevNode = removedValue.prev;
+        if (nextNode == null) {
+            tail = prevNode;
+        } else {
+            nextNode.prev = prevNode;
+            removedValue.prev = null;
```
KseniiaMakarova, line 138:
```suggestion
            removedValue.next = null;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,149 @@
...
+        }
+        if (prevNode == null) {
+            head = nextNode;
+        } else {
+            prevNode.next = nextNode;
+            removedValue.next = null;
```
KseniiaMakarova, line 144:
```suggestion
            removedValue.prev = null;
```

KseniiaMakarova (changes requested):

KseniiaMakarova (changes requested):
please be more attentive to the comments (not everything is fixed) and fix travis

src/main/java/core/basesyntax/MyLinkedList.java
KseniiaMakarova: please write below what is the purpose of this check

src/main/java/core/basesyntax/MyLinkedList.java
Reastador: Sorry, I miss this one

KseniiaMakarova (approved):

PR #244, opened 2021-04-12
Mykola manko

mankokolya:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,164 @@
...
+        } else if (previous == null) {
+            next.previous = null;
+            head = next;
+        } else if (next == null) {
+            previous.next = null;
+            tail = temp.previous;
+        } else {
+            previous.next = next;
+            next.previous = previous;
+        }
```
KseniiaMakarova, line 146: let's try to simplify this a bit

first check if `next` is null - write `if-else` construction

then check if `previous `is null - write another `if-else` construction

let's also set node's value to null in the end

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,168 @@
...
+                previous.next = null;
+                tail = temp.previous;
+            }
+        } else if (previous == null) {
+            next.previous = null;
+            head = next;
+        } else {
+            previous.next = next;
+            next.previous = previous;
+        }
```
KseniiaMakarova, line 149:
```suggestion
        if (previous == null) {
            head = next;
        } else {
            previous.next = next;
            temp.previous = null;
        }

        if (next == null) {
            tail = previous;
        } else {
            next.previous = previous;
            temp.next = null;
        }
```

KseniiaMakarova (approved):

PR #248, opened 2021-04-12
First solution

Ihor-Nykyforov:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,148 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private Node<T> nodeByIndex(int index) {
```
Reastador, line 93: in my opinion, is better to create two different methods - one to check the index range and another to get Node

Reastador (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,149 @@
...
+        return size == 0;
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("No such index in data");
+        }
+    }
+
+    private Node<T> node(int index) {
```
KseniiaMakarova, line 103:
```suggestion
    private Node<T> findNode(int index) {
```
let's use verbs for method names

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,149 @@
...
+        return unlink(oldNode);
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> currentNode = head;
+        for (int i = 0; i < size; i++) {
+            if ((currentNode.item == object)
+                    || (object != null && object.equals(currentNode.item))) {
+                remove(i);
```
KseniiaMakarova, line 79: let's use unlink() here - we don't need to make calls to `checkIndex()` or `node()` methods here

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,149 @@
...
+        } else if (oldNode == head) {
+            nextNode.prev = null;
+            head = nextNode;
+        } else if (oldNode == tail) {
+            previousNode.next = null;
+            tail = previousNode;
+        } else {
+            nextNode.prev = previousNode;
+            previousNode.next = nextNode;
+        }
```
KseniiaMakarova, line 134: let's try to simplify this a bit

first check if `nextNode` is null - write `if-else` construction

then check if `previousNode` is null - write another `if-else` construction

let's also set node's value to null in the end

KseniiaMakarova (changes requested):
overall nice job!

KseniiaMakarova (approved):

PR #257, opened 2021-04-12
first solution

andriiromaniuk:


IhorRome (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,145 @@
...
+        Node<T> currentNode = tail;
+        Node<T> newNode = new Node<>(tail, value, null);
+        tail = newNode;
+        if (currentNode == null) {
+            head = newNode;
+        } else {
+            currentNode.next = newNode;
+        }
+        size++;
+        return true;
```
Sofasmile, line 33:
```suggestion
        Node<T> newNode = new Node<>(tail, value, null);
        if (tail == null) {
            head = newNode;
        } else {
            tail.next = newNode;
        }
        tail = newNode;
        size++;
        return true;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,145 @@
...
     }
 
     @Override
     public void add(T value, int index) {
-
+        if (index == size) {
+            add(value);
+        } else {
```
Sofasmile, line 40: we can add return after add() method and then we can remove else block

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,145 @@
...
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> tempNode = head;
+        while (tempNode != null) {
+            if (tempNode.item == object || tempNode.item != null && tempNode.item.equals(object)) {
+                unlink(tempNode);
+                return true;
+            } else {
```
Sofasmile, line 87: do we need else block?

Sofasmile (changes requested):

Sofasmile (approved):

PR #266, opened 2021-04-13
Add solution

AventusDev:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,6 +3,7 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private static final String exceptionText = "Index out of bound exception.";
```
KseniaObertii, line 6: constants write in uppercase, words divided by "_"

KseniaObertii (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
AventusDev: +

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,146 @@
...
-        return false;
+        if (size == 0) {
+            Node<T> first = new Node<>(null, value, null);
+            head = first;
+            tail = first;
+        } else {
+            Node<T> newNode = new Node<>(tail, value, null);
+            tail.next = newNode;
+            tail = newNode;
+        }
```
KseniiaMakarova, line 21:
```suggestion
        Node<T> newNode = new Node<>(tail, value, null);
        if (size == 0) {
            head = first;
        } else {
            tail.next = newNode;
        }
        tail = newNode;
```
let's always try to move out the lines that repeat in both `if` and `else`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,146 @@
...
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
+        return true;
     }
 
     @Override
     public T get(int index) {
-        return null;
+        indexChecker(index);
```
KseniiaMakarova, line 57:
```suggestion
        checkIndex(index);
```
use verbs for method names

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,146 @@
...
+        return unlink(getNode(index));
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> currentNode = head;
+        for (int i = 0; i < size; i++) {
+            if (currentNode.value == null || currentNode.value.equals(object)) {
```
KseniiaMakarova, line 80:
```suggestion
            if (currentNode.value == object (we should check if they are both null) || (what if currentNode is null here?) currentNode.value.equals(object)) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,146 @@
...
+        } else {
+            for (int i = size - 1; i > index; i--) {
+                searchedNode = searchedNode.prev;
+            }
+        }
+        return searchedNode;
+    }
+
+    private T unlink(Node<T> node) {
+        Node<T> oldNode = node;
```
KseniiaMakarova, line 120: do you need this variable?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,146 @@
...
+        } else {
+            prevNode.next = nextNode;
+        }
+        if (nextNode == null) {
+            tail = prevNode;
+        } else {
+            nextNode.prev = prevNode;
+        }
+        size--;
+        return node.value;
```
KseniiaMakarova, line 134:
```suggestion
        if (prevNode == null) {
            head = nextNode;
        } else {
            prevNode.next = nextNode;
            node.prev = null;
        }
        if (nextNode == null) {
            tail = prevNode;
        } else {
            nextNode.prev = prevNode;
            node.next = null;
        }
        T oldValue = node.value;
        node.value = null;
        size--;
        return oldValue;
```

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,149 @@
...
-        return false;
+        if (size == 0) {
+            Node<T> first = new Node<>(null, value, null);
+            head = first;
+            tail = first;
+        } else {
+            Node<T> newNode = new Node<>(tail, value, null);
+            tail.next = newNode;
+            tail = newNode;
+        }
```
KseniiaMakarova, line 21: what about the comment that I left earlier here?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,48 +3,149 @@
...
-        return null;
+        checkIndex(index);
+        return unlink(getNode(index));
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> currentNode = head;
+        for (int i = 0; i < size; i++) {
+            if (currentNode.value == object || currentNode.value.equals(object)) {
```
KseniiaMakarova, line 80: what if `currentNode` is null and `object` is not null?

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #313, opened 2021-08-23
Linked list homework

mishamolnar:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,172 @@
...
-        return null;
+        indexCheckNotSize(index);
+        indexCheck(index);
+        T buffer = getNodeByIndex(index).item;
+        getNodeByIndex(index).item = value;
+        return buffer;
     }
 
     @Override
     public T remove(int index) {
```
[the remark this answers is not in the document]
MaliukDaria: Checklist: Create a separate method unlink(Node node).

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,155 @@
...
+        }
+        return bufferNode;
+    }
+
+    private static class Node<E> {
+        private E item;
+        private Node<E> next;
+        private Node<E> prev;
+
+        Node(Node<E> prev, E element, Node<E> next) {
```
MaliukDaria, line 145: what about access modifiers?

MaliukDaria (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,147 @@
...
     }
 
     @Override
     public void add(T value, int index) {
+        if (index > size || index < 0) {
+            throw new IndexOutOfBoundsException("This index does not exist");
+        }
+        if (index == size) {
+            add(value);
+        } else if (index == 0) {
```
MaliukDaria, line 29:
```suggestion
        if (index == size) {
            add(value);
            return;
        } 
        if (index == 0) {
```

MaliukDaria (approved):

PR #315, opened 2021-08-25
Java LinkedList Homework

Sinelife:
add realisation to methods of LinkedList - size(), isEmpty(), add(elem), add(index), get(index), set(elem, index), remove(index), remove(elem), addAll(list)

### DIFF
--- a/src/main/java/core/basesyntax/MyLinkedList.java
+++ b/src/main/java/core/basesyntax/MyLinkedList.java
@@ -8,6 +8,7 @@ public class MyLinkedList<T> implements MyLinkedListInterface<T> {
     private static final String FROM_START_TYPE_OF_ITERATION = "fromStart";
     private static final String FROM_END_TYPE_OF_ITERATION = "fromEnd";
     private static final String COMMA = ",";
+    private static final String SPACE = " ";
     private static final String RIGHT_BRACKET = "]";
     private static final String LEFT_BRACKET = "[";
     private Node<T> head;
@@ -94,9 +95,7 @@ public class MyLinkedList<T> implements MyLinkedListInterface<T> {
 
     @Override
     public T get(int index) {
-        if (index < 0 || index > size - 1) {
-            throw new IndexOutOfBoundsException(INDEX_OF_BOUND_EXCEPTION_MESSAGE);
-        }
+        checkIndex(index);
         Node<T> current = head;
         int counter = 0;
         int step = 1;
@@ -120,9 +119,7 @@ public class MyLinkedList<T> implements MyLinkedListInterface<T> {
 
     @Override
     public T set(T value, int index) {
-        if (index < 0 || index > size - 1) {
-            throw new IndexOutOfBoundsException(INDEX_OF_BOUND_EXCEPTION_MESSAGE);
-        }
+        checkIndex(index);
         T oldValue = null;
         Node<T> current = head;
         int counter = 0;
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,284 @@
...
+            if (equals(current.value, o)) {
+                return counter;
+            }
+            counter++;
+            current = current.next;
+        }
+        return -1;
+    }
+
+    public boolean equals(T first, T second) {
```
Max-Pochepets, line 255: what is this equals? what are you comparing? why is it public?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,284 @@
...
+        unlink(current);
+        return current.value;
+    }
+
+    private void unlink(Node<T> current) {
+        if (current != null) {
+            Node<T> newNext = current.next;
+            Node<T> newPrev = current.prev;
+            newNext.prev = newPrev;
+            newPrev.next = newNext;
```
Max-Pochepets, line 228: can `newNext` or `newPrev` be null?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,284 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    public int indexOf(T o) {
```
Max-Pochepets, line 242:
```suggestion
    private int indexOf(T value) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,284 @@
...
+                    ? current.next : current.prev;
+            counter += step;
+        }
         return null;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        if (index < 0 || index > size - 1) {
```
Max-Pochepets, line 123:
```suggestion
        if (index < 0 || index >= size) {
```
and move it to a private method. it is used more than once

Max-Pochepets (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
Sinelife: make equals private. This quals check if two objects are equals even if them are null. The same thing did Objects,equals(a,b)

src/main/java/core/basesyntax/MyLinkedList.java
Sinelife: rework remove methods and delete methods removeElementFromEnd, removeElementFromBeginning, removeLastElement.
Add a null-check code in unlink() method

src/main/java/core/basesyntax/MyLinkedList.java
Sinelife: renamed variable

src/main/java/core/basesyntax/MyLinkedList.java
Sinelife: moved code to method checkSize() and fixed diapason

Max-Pochepets (approved):
this is a very large PR for this task. next time try to write as little code as possible

PR #316, opened 2021-08-26
implemented class MyLinkedList

VladGural:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,146 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
     @Override
     public void add(T value) {
+        Node<T> item = new Node<>(tail, value, null);
```
Max-Pochepets, line 12:
```suggestion
        Node<T> newNode = new Node<>(tail, value, null);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,146 @@
...
+        tail = item;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        checkPositionIndex(index);
+        if (index == size) {
+            add(value);
+        } else {
```
Max-Pochepets, line 27: no need for this else, we can `return` from prev if. don't complicate ifs

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,146 @@
...
-        return null;
+        checkElementIndex(index);
+        Node<T> indexNode = getNode(index);
+        if (head == indexNode && tail == indexNode) {
+            head = null;
+            tail = null;
+        } else if (head == indexNode) {
+            indexNode.next.prev = null;
+            head = indexNode.next;
+        } else if (tail == indexNode) {
```
Max-Pochepets, line 74: too much else if. try to think about two if-else constructions: if-else if-else, without `else if`

Max-Pochepets (changes requested):

VladGural:
renamed variable item to newNode
deleted unnecessary "else"
recoded if-else constructions

Max-Pochepets (approved):

PR #317, opened 2021-08-27
a resolved LinkedList task

Yelinskyi:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,146 @@
...
+        private Node<T> prev;
+
+        Node(Node<T> prev, T item, Node<T> next) {
+            this.prev = prev;
+            this.item = item;
+            this.next = next;
+        }
+    }
+
+    private T unlink(Node<T> deletedNode) { // Для зміни зв'язків
```
Max-Pochepets, line 22: private methods should be listed after public

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,146 @@
...
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Index is invalid!");
+        }
+    }
+
+    private Node<T> findNode(int index) {
+        Node<T> findNode;
+        if (index < size / 2) {
+            findNode = first;
+            for (int i = 1; i <= index; i++) {
```
Max-Pochepets, line 50:
```suggestion
            for (int i = 0; i < index; i++) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,146 @@
...
+                findNode = findNode.next;
+            }
+        } else {
+            findNode = last;
+            for (int i = size - 2; i >= index; i--) {
```
Max-Pochepets, line 55:
```suggestion
            for (int i = size - 1; i > index; i--) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,146 @@
...
+            first = newNode;
+        } else {
+            lastNode.next = newNode;
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        //checkPosition(index - 1);
```
Max-Pochepets, line 77: if we have 9 elements in list, why can't we insert on index 9?

Max-Pochepets (changes requested):
delete comments

src/main/java/core/basesyntax/MyLinkedList.java
Yelinskyi: In checkPosition() method, index check '>='.
I did the same checking, in add() it was working, but in checkPosition is wasn't working.

src/main/java/core/basesyntax/MyLinkedList.java
Yelinskyi: I rewrote it.

src/main/java/core/basesyntax/MyLinkedList.java
Yelinskyi: I wrirote it.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -145,4 +105,44 @@ public int size() {
     public boolean isEmpty() {
         return size == 0;
     }
+
+    private T unlink(Node<T> deletedNode) { // Для зміни зв'язків
```
Max-Pochepets, line 109:
```suggestion
    private T unlink(Node<T> deletedNode) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -145,4 +105,44 @@ public int size() {
...
+            last = prevNode;
+        } else {
+            nextNode.prev = prevNode;
+        }
+        size--;
+        return deletedNode.item;
+    }
+
+    private void checkPosition(int index) {
+        if (index < 0 || index >= size) {
```
Max-Pochepets, line 128: imagine we have 10 elements in list. indexes are 0-9. Can I insert new node on index 10? This check says I cannot. Think about it and try to add this method again where it didn't work

Max-Pochepets (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
Yelinskyi: I deleted it.

src/main/java/core/basesyntax/MyLinkedList.java
Yelinskyi: Checking allows you to put an item on the 10th index.

But you made me find my mistake. 
When I passed (index -1), it wasn't < 0.

I created new check method for add().

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -130,6 +125,12 @@ private void checkPosition(int index) {
         }
     }
 
+    private void checkPositionAdd(int index) {
+        if (index < -1 || index >= size) {
```
Max-Pochepets, line 129: now you're allowing a -1 index. also, from previous example this check still does not allow to place on index 10 with size 10, because if index = 10 - condition `index >= size` will be true and an exception will be thrown

Max-Pochepets (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
Yelinskyi: It isn't so.
Because I gave in the method (index -1), but I rewrote it.

Max-Pochepets (approved):

PR #326, opened 2021-08-31
Hw9 creation of linked list solution

Yehor-Zapara:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
+    private class Node<T> {
+        private T item;
+        private Node<T> prev;
+        private Node<T> next;
+
+        Node(Node<T> prev, T item, Node<T> next) {
+            this.item = item;
+            this.prev = prev;
+            this.next = next;
+        }
```
Koriastras, line 165: I'm not sure, but I suppose it might be better to put this class after we fill the fields of our main class, but before methods. And it should be static, I think.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
+        nodeByIndex.item = value;
+        return oldValue;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        Node<T> nodeByIndex = findByIndex(index);
+        return unlink(nodeByIndex);
```
Koriastras, line 53: here you can do without creating a new variable

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> first = null;
+    private Node<T> last = null;
```
Koriastras, line 7: initialization here is redundant

Koriastras (changes requested):

Koriastras (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
...
+    }
+
+    private void checkIndex(int index) {
+        if (index >= size || index < 0) {
+            throw new IndexOutOfBoundsException("Can't find element with index " + index);
+        }
+    }
+
+    private Node<T> findByIndex(int index) {
+        int middleOfList = (size - 1) / 2;
```
Max-Pochepets, line 98: we don't need this variable. also, why `size - 1`?

Max-Pochepets (changes requested):

Max-Pochepets (approved):

PR #327, opened 2021-08-31
first version with unlink method

Evgen-Krasnenkov:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,224 @@
...
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    private static class Node<T> {
+        private T item;
+        private Node<T> next;
+        private Node<T> previous;
+
+        private Node(T element) {
```
konstde00, line 16: Why we need 2 constructors here? Could we do the same only with second one?

konstde00, line 16: Why we need 2 constructors here? Could we do the same only with second one?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,224 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void unlinkNode(Node node) {
```
konstde00, line 193: To my mind it's not enough just to unlink input node, we have to connect previous one with the next one

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,224 @@
...
+            Node<T> addNode = new Node<>(null, value, currentNode);
+            addNode.previous = currentNode.previous;
+            currentNode.previous = addNode;
+            head = addNode;
+            size++;
+            return;
+        }
+
+        if (index <= 0 || index > size) {
+            throw new IndexOutOfBoundsException("Wrong Index.");
```
konstde00, line 70: It would be much better to create checkIndex(int index) method with logic which you used here and then just call it

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,224 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+
```
konstde00, line 6: We don't need empty line here, please remove it

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,224 @@
...
+            Node<T> newNode = new Node<>(tail, value, null);
+            tail.next = newNode;
+            tail = newNode;
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (size == 0 && index == size) {
```
konstde00, line 42: We don't need two if-else cases here, please, combine them in one with add(value)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,224 @@
...
+        return size == 0;
+    }
+
+    private void unlinkNode(Node node) {
+        node.next = null;
+        node.previous = null;
+    }
+
+    private Node<T> findNode(int index) {
+        if ((index < size) && index >= 0 && (index <= size / 2) && head != null) {
```
konstde00, line 199: Please, simplify this method using checkIndex

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,224 @@
...
+        if (index > 0 && index < size) {
+            Node<T> currentNode = findNode(index);
+            Node<T> addNode = new Node<>(null, value, currentNode);
+            addNode.previous = currentNode.previous;
+            currentNode.previous.next = addNode;
+            currentNode.previous = addNode;
+            size++;
+            return;
+        }
+        if (index == 0 && index < size) {
```
konstde00, line 59: Could you add this case just to "if()"  where index = 0 (and then it would be much better to combine all cases where index != size or index != 0)

src/main/java/core/basesyntax/MyLinkedList.java
Evgen-Krasnenkov: Fully agree. I deleted that.

src/main/java/core/basesyntax/MyLinkedList.java
Evgen-Krasnenkov: agree as you suggestion implied more improvement of my code. thanks

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,224 @@
...
+        currentNode.previous.next = setNode;
+        currentNode.next.previous = setNode;
+        unlinkNode(currentNode);
+        return currentNode.item;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        Node<T> currentNode = findNode(index);
```
konstde00, line 120: I'm not sure, but you can just use unlink() method to remove the node (we don't care what will happen with the value of node if we don't use it)

Evgen-Krasnenkov: я повністтю переписав remove by index and by value and unlinknode

src/main/java/core/basesyntax/MyLinkedList.java
Evgen-Krasnenkov: не згоден http://prntscr.com/1qw7ahg. mvn не пропускає

src/main/java/core/basesyntax/MyLinkedList.java
Evgen-Krasnenkov: згоден

src/main/java/core/basesyntax/MyLinkedList.java
Evgen-Krasnenkov: я зробив це.

src/main/java/core/basesyntax/MyLinkedList.java
konstde00: Ok, that's my mistake, I'm sorry

konstde00 (approved):

src/main/java/core/basesyntax/MyLinkedList.java
MaliukDaria: в скриншоте чекстайл ругается на отсутствие строки между классом и импортом. А пустую строку №6 нужно убрать

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,176 @@
...
     @Override
     public void add(T value) {
+        if (size == 0) {
+            Node<T> newNode = new Node<>(null, value, null);
+            head = tail = newNode;
+        } else {
+            Node<T> newNode = new Node<>(tail, value, null);
+            tail.next = newNode;
+            tail = newNode;
+        }
```
MaliukDaria, line 32:
```suggestion
		Node<T> newNode = new Node<>(tail, value, null);
        if (size == 0) {
            head = newNode;
        } else {
            tail.next = newNode;
        }
        tail = newNode;
```

It seems to me it looks simpler that way

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,176 @@
...
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Wrong Index.");
+        }
+        if (index == size) {
+            add(value);
+            return;
+        }
+        if (index != 0 && index != size) {
```
MaliukDaria, line 45: do we need to check `index != size` here? (look at the line 41)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,176 @@
...
+            size++;
+            return;
+        }
+        if (index == 0) {
+            Node<T> addNode = new Node<>(null, value, head);
+            head.previous = addNode;
+            head = addNode;
+            size++;
+            return;
+        }
```
MaliukDaria, line 60:
```suggestion
            Node<T> currentNode = findNode(index);
            Node<T> addNode = new Node<>(currentNode.previous, value, currentNode);
        if (index != 0) {
            currentNode.previous.next = addNode;  
        } else {
            head = addNode;
        }
        currentNode.previous = addNode;
        size++;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,176 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void unlinkNode(Node node) {
```
MaliukDaria, line 114: think about how could we simplify this method. In each of the `if` statements you have a lot of repeating logic, think about how to combine it

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,176 @@
...
+        node.previous.next = node.next;
+        node.next.previous = node.previous;
+        node.next = null;
+        node.previous = null;
+        size--;
+    }
+
+    private Node<T> findNode(int index) {
+        checkIndexValidity(index);
+        if ((index <= size / 2) && head != null) {
```
MaliukDaria, line 148: do we need this check `head != null` ?

MaliukDaria (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
Evgen-Krasnenkov: I agree.

src/main/java/core/basesyntax/MyLinkedList.java
Evgen-Krasnenkov: okey

src/main/java/core/basesyntax/MyLinkedList.java
Evgen-Krasnenkov: you are right once again

src/main/java/core/basesyntax/MyLinkedList.java
Evgen-Krasnenkov: thanks. Next if has checking up for head ! = null. It's not necessary

MaliukDaria (approved):
🔥

PR #330, opened 2021-08-31
Hw linked list

osa240:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,150 @@
...
+            this.prev = prev;
+            this.element = element;
+            this.next = next;
+        }
+    }
+
+    public MyLinkedList() {
+        first = new Node<>(null, null, null);
+        last = new Node<>(null, null, null);
+    }
```
AlexRozv, line 27: we don't need that constructor

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,150 @@
...
+
     @Override
     public void add(T value) {
+        if (size == 0) {
+            temporaryNode = new Node<>(first, value, last);
```
AlexRozv, line 32: we can use null as parameters instead of first and last

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,150 @@
...
+            first = temporaryNode;
+            last = temporaryNode;
+            size++;
+            return;
+        }
+        newNode = new Node<>(temporaryNode, value, null);
+        temporaryNode.next = newNode;
+        temporaryNode = newNode;
+        last = temporaryNode;
```
AlexRozv, line 41:
```suggestion
        newNode = new Node<>(last, value, null);
        last.next = newNode;
        last = newNode;
```
we can implement it without temporaryNode

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,150 @@
...
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index == size) {
+            add(value);
+            return;
+        }
+        if (index >= 0 && index < size) {
```
AlexRozv, line 51: we can create a separate method to check index and throw exception there if it's not suitable

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,150 @@
...
+            return;
+        }
+        throw new IndexOutOfBoundsException("It is not next index!");
     }
 
     @Override
     public void addAll(List<T> list) {
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
```
AlexRozv, line 70:
```suggestion
        for (T t : list) {
            add(t);
        }
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,150 @@
...
 
     @Override
     public boolean remove(T object) {
+        temporaryNode = first;
+        for (int i = 0; i < size; i++) {
+            if (temporaryNode.element == object || (temporaryNode.element != null
+                    && temporaryNode.element.equals(object))) {
+                remove(i);
+                return true;
+            }
```
AlexRozv, line 99: you can use indexToNode() to get the needed node and then just use unlink() on it

AlexRozv (changes requested):
student's review done

src/main/java/core/basesyntax/MyLinkedList.java
osa240: i will replace this code in other part of code. without initialize have NullPoinerException whe i try to remove first element.

src/main/java/core/basesyntax/MyLinkedList.java
osa240: logic of my code must to do only with that variables!

src/main/java/core/basesyntax/MyLinkedList.java
osa240: agree with you!

src/main/java/core/basesyntax/MyLinkedList.java
osa240: ok

AlexRozv (approved):
Good job!

src/main/java/core/basesyntax/MyLinkedList.java
Max-Pochepets: this if can be simplified up to 3 lines of code. or even more: try to think about checking not `== 0`, but `> 0`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,151 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void checkIndexIfInBound(T value, int index) {
```
Max-Pochepets, line 101: method named `checkIndexIfInBound` must only check if index is in bounds, no more

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,151 @@
...
+            last.next = null;
+            size--;
+            return;
+        }
+        (node.prev).next = node.next;
+        (node.next).prev = node.prev;
+        size--;
+    }
+
+    private Node<T> indexToNode(int index) {
```
Max-Pochepets, line 136:
```suggestion
    private Node<T> findNodeByIndex(int index) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,151 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size;
+    private Node<T> first;
+    private Node<T> last;
+    private Node<T> newNode;
+    private Node<T> temporaryNode;
```
Max-Pochepets, line 10: we don't need these fields. these can be local variables

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,151 @@
...
+            return;
+        }
+        if (index == size - 1) {
+            last = node.prev;
+            last.next = null;
+            size--;
+            return;
+        }
+        (node.prev).next = node.next;
+        (node.next).prev = node.prev;
```
Max-Pochepets, line 132: why do you need braces?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,151 @@
...
+            first = node.next;
+            first.prev = null;
+            size--;
+            return;
+        }
+        if (index == size - 1) {
+            last = node.prev;
+            last.next = null;
+            size--;
+            return;
```
Max-Pochepets, line 129: code duplication. can we simplify this somehow?

Max-Pochepets (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -23,18 +21,17 @@ protected Node(Node<T> prev, T element, Node<T> next) {
...
-            size++;
-            return;
+        if (size > 0) {
+            Node<T> newNode = new Node<>(last, value, null);
+            last.next = newNode;
+            last = newNode;
+        } else {
+            first = new Node<>(null, value, null);
+            last = new Node<>(null, value, null);
+            Node<T> newNode = new Node<>(last, value, first);
```
Max-Pochepets, line 31:
```suggestion
            Node<T> newNode = new Node<>(null, value, null);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -98,51 +107,39 @@ public boolean isEmpty() {
         return size == 0;
     }
 
-    private void checkIndexIfInBound(T value, int index) {
+    private void checkIndexIfInBound(int index) {
         if (index >= 0 && index < size) {
```
Max-Pochepets, line 111: it's a void method, and we don't actually need return here. how can we achieve it?

Max-Pochepets (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
osa240: I have had NPE so i have added one more condition in method unLink()

Max-Pochepets (approved):

PR #333, opened 2021-08-31
implemented LinkedList class

SergiiLonskyi:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,152 @@
...
+            Node<T> nodeCurrent = head;
+            for (int i = 1; i < index; i++) {
+                nodeCurrent = nodeCurrent.next;
+            }
+            newNode.next = nodeCurrent.next;
+            nodeCurrent.next = newNode;
+            newNode.previous = nodeCurrent;
+            newNode.next.previous = newNode;
+        }
+        size++;
```
blakh706, line 63: I think it can be simplified

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,152 @@
...
+        return null;
+    }
+
+    private Node<T> getNodeByIndex(int index) {
+        Node<T> node = head;
+        for (int i = 1; i <= index; i++) {
+            node = node.next;
+        }
+        return node;
+    }
```
blakh706, line 130: This we can do faster

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,152 @@
...
+
+    private Node<T> getNode(T value) {
+        for (int i = 0; i < size; i++) {
+            if ((getNodeByIndex(i).element == value)
+                    || (value != null && value.equals(getNodeByIndex(i).element))) {
+                return getNodeByIndex(i);
+            }
+        }
+        return null;
+    }
```
blakh706, line 122: We need this method ?

blakh706 (changes requested):

blakh706 (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
...
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("The index is wrong");
+        }
+        Node<T> newNode = new Node<>(null, value, null);
+        if (head == null) {
+            head = tail = newNode;
+        } else if (index == 0) {
```
Max-Pochepets, line 42: if (index == size) - does it look familiar? like `add(T value)` method? also, all this can be simplified. you can find node on index and node before it. and all you need to do check whenever previous node is null or not

Max-Pochepets (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
Max-Pochepets: what about this point? **also, all this can be simplified. you can find node on index and node before it. and all you need to do check whenever previous node is null or not**

Max-Pochepets (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
SergiiLonskyi: Hope, I understood you right! Please check last commit.

Max-Pochepets (approved):

PR #334, opened 2021-08-31
Implemented MyLinkedListInterface methods.

VladyslavLekhter:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,134 @@
...
     }
 
     @Override
     public boolean isEmpty() {
+        return size == 0;
+    }
+
+    private void checkIndex(int index) {
+        if ((index >= size && index != ARRAYS_ZERO_BORDER) || index < ARRAYS_ZERO_BORDER) {
+            throw new IndexOutOfBoundsException("Index out of bounds array size");
```
kvoytikh, line 96: make constant field with msg

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,134 @@
...
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private static final int ARRAYS_ZERO_BORDER = 0;
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    public MyLinkedList() {
+        head = new Node<>(null, null, null);
+        tail = new Node<>(null, null, null);
```
kvoytikh, line 13: i am not sure that you need this constructor :)

kvoytikh (changes requested):
good job :)

src/main/java/core/basesyntax/MyLinkedList.java
VladyslavLekhter: Thank you, i checked this) May be i could solve it in another way, but my code don't work with out it))

kvoytikh (approved):
Good job!)

src/main/java/core/basesyntax/MyLinkedList.java
Max-Pochepets: you don't need this. you can pass `null` as parameters in nodes you add

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,135 @@
...
+        } else {
+            currentNode = tail;
+            for (int i = size - 1; i >= index; i--) {
+                currentNode = currentNode.prev;
+            }
+        }
+        return currentNode;
+    }
+
+    private boolean addFirstNode(T value) {
```
Max-Pochepets, line 117: we don't need this method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,135 @@
...
+    public MyLinkedList() {
+        head = new Node<>(null, null, null);
+        tail = new Node<>(null, null, null);
+    }
+
     @Override
     public void add(T value) {
+        if (!addFirstNode(value)) {
+            tail.prev = tail.prev.next = new Node<>(tail.prev, value, tail);
+            size++;
```
Max-Pochepets, line 21: try to think of it this way: we can create a node, pointing straight to tail. and think about what can we do based on different size conditions?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,135 @@
...
+        currentNode.value = value;
+        return oldValue;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        Node<T> currentNode = findCurrentNode(index);
+        currentNode.prev.next = currentNode.next;
```
Max-Pochepets, line 64: what if `currentNode == head`?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,135 @@
...
+        currentNode.next.prev = currentNode.prev;
```
Max-Pochepets, line 65: what if `currentNode == tail`?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,135 @@
...
     @Override
     public boolean remove(T object) {
-        return false;
+        int countIndex = 0;
+        Node<T> currentNode = head.next;
+        T nodeValue = currentNode.value;
+        while (nodeValue != null && !nodeValue.equals(object)) {
+            nodeValue = currentNode.next.value;
+            currentNode = currentNode.next;
+            countIndex++;
```
Max-Pochepets, line 79: can we use `remove(int index)` here?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,135 @@
...
+    private boolean addFirstNode(T value) {
+        if (head.next == null && tail.prev == null) {
+            head.next = tail.prev = new Node<>(head, value, tail);
+            size++;
+            return true;
+        }
         return false;
     }
+
+    static class Node<T> {
```
Max-Pochepets, line 126:
```suggestion
    private static class Node<T> {
```

Max-Pochepets (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -61,25 +65,33 @@ public T set(T value, int index) {
...
+            currentNode.next.prev = null;
+            head = currentNode.next;
+        } else if (currentNode == tail || currentNode != null && currentNode.equals(tail)) {
+            currentNode.prev.next = null;
+            tail = currentNode.prev;
+        } else {
+            currentNode.prev.next = currentNode.next;
+            currentNode.next.prev = currentNode.prev;
+            currentNode.prev = currentNode.next = null;
+        }
```
Max-Pochepets, line 80: think of it this way: you can find node on index, and previous to it. between them you have to insert new node. all you need to check - is previous node null or not

Max-Pochepets (changes requested):

PR #335, opened 2021-08-31
added solution

kvoytikh:
added solution

### DIFF
--- a/src/main/java/core/basesyntax/MyLinkedList.java
+++ b/src/main/java/core/basesyntax/MyLinkedList.java
@@ -3,45 +3,150 @@ package core.basesyntax;
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private static final String OUT_OF_BOUNDS_EXCEPTION = "Index out of bounds!";
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    private class Node<T> {
+        private T value;
+        private Node<T> prev;
+        private Node<T> next;
+
+        public Node(Node<T> prev, T value, Node<T> next) {
+            this.prev = prev;
+            this.value = value;
+            this.next = next;
+        }
+
+    }
+
     @Override
     public void add(T value) {
+        Node<T> newNode = new Node<>(tail, value, null);
+        if (size == 0) {
+            head = newNode;
+        } else {
+            tail.next = newNode;
+        }
+        tail = newNode;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index == size) {
+            add(value);
+            return;
+        }
+
+        if (index == 0) {
+            Node<T> newNode = new Node<>(null, value, head);
+            head.prev = newNode;
+            head = newNode;
+        } else {
+            indexCheck(index);
+            Node<T> currentNode = getNodeByIndex(index);
+            Node<T> prevNode = currentNode.prev;
+            Node<T> newNode = new Node<>(prevNode, value, currentNode);
+            currentNode.prev = newNode;
+            prevNode.next = newNode;
+        }
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
+        for (T item : list) {
+            add(item);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        indexCheck(index);
+        return getNodeByIndex(index).value;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        indexCheck(index);
+        Node<T> currentNode = getNodeByIndex(index);
+        T previousNode = currentNode.value;
+        currentNode.value = value;
+        return previousNode;
     }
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,152 @@
...
+    }
+
+    @Override
+    public void add(T value, int index) {
+        if (index == size) {
+            add(value);
+            return;
+        }
+
+        if (index == 0) {
```
VladyslavLekhter, line 43: Redundant empty line.

VladyslavLekhter (changes requested):

VladyslavLekhter (approved):
Good job!)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,150 @@
...
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    private class Node<T> {
+        private T value;
+        private Node<T> prev;
+        private Node<T> next;
+
+        public Node(Node<T> prev, T value, Node<T> next) {
```
MaliukDaria, line 16: why public?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,150 @@
...
+    public void addAll(List<T> list) {
+        for (T item : list) {
+            add(item);
+        }
+    }
+
+    @Override
+    public T get(int index) {
+        indexCheck(index);
+        return getNodeByIndex(index).value;
```
MaliukDaria, line 66: we can add `indexCheck()` method into `getNodeByIndex()` method, this will reduce number of duplicated code

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,150 @@
...
+    @Override
+    public T remove(int index) {
+        indexCheck(index);
+        return unlink(getNodeByIndex(index));
+    }
+
+    @Override
+    public boolean remove(T object) {
+        Node<T> node = getNode(object);
+        if (node != null) {
```
MaliukDaria, line 87: could node be null here?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,150 @@
...
+    }
+
+    private Node<T> getNodeByIndex(int index) {
+        Node<T> bufferNode = head;
+        if (index <= size / 2) {
+            for (int i = 0; i < index; i++) {
+                bufferNode = bufferNode.next;
+            }
+        } else {
+            for (int j = size; j > size - index; j--) {
```
MaliukDaria, line 117: why `j > size - index`? We have to move from tail to previous nodes when index > size/2

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,150 @@
...
+
+    private Node<T> getNode(T value) {
+        for (int i = 0; i < size; i++) {
+            if ((getNodeByIndex(i).value == value)
+                    || (value != null && value.equals(getNodeByIndex(i).value))) {
+                return getNodeByIndex(i);
+            }
+        }
+        return null;
+    }
```
MaliukDaria, line 132: What is the complexity of this algorithm? How many times do you run through the list here? Let's start moving from the head through the nodes and compare the values, if we find match, we will return the current node

MaliukDaria (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
kvoytikh: yes, element can be no existed. We can get NullPointerException

kvoytikh: so we should return false, but no Exception (it was said in tests)

src/main/java/core/basesyntax/MyLinkedList.java
kvoytikh: this makes the if condition, thanks to this loop we go from the end to the specific index

src/main/java/core/basesyntax/MyLinkedList.java
kvoytikh: done)

src/main/java/core/basesyntax/MyLinkedList.java
kvoytikh: done)

src/main/java/core/basesyntax/MyLinkedList.java
kvoytikh: this is done as it is written in the description of your algorithm :)

MaliukDaria (approved):
🔥

PR #338, opened 2021-08-31
Linked list hw

VitaliiMYKH:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
...
+
+    private Node<T> elementFinder(int index) {
+        Node<T> placeFinder;
+        if ((size / 2) > index) {
+            placeFinder = head;
+            for (int i = 0; i < index; i++) {
+                placeFinder = placeFinder.next;
+            }
+            return placeFinder;
+        } else {
```
Yelinskyi, line 146: Do you need 'else' after 'return'?

Yelinskyi (changes requested):

Yelinskyi (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,162 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size = 0;
```
Max-Pochepets, line 6:
```suggestion
    private int size;
```
primitives have default values

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,162 @@
...
+            Node<T> newHead = new Node<>(null, value, head);
+            newHead.next.prev = newHead;
+            head = newHead;
+            size++;
+        } else if (size > 0 && index == size) {
+            Node<T> newTail = new Node<T>(tail, value, null);
+            newTail.prev.next = newTail;
+            tail = newTail;
+            size++;
+        } else {
```
Max-Pochepets, line 37: too complicated construction. we need to check if size = 0, but try to think this way: can we find nodes from both sides of inserted one? what can we do with them?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,162 @@
...
+        }
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("can not reach element with this index");
+        }
+    }
+
+    private Node<T> elementFinder(int index) {
```
Max-Pochepets, line 137:
```suggestion
    private Node<T> findNodeByIndex(int index) {
```
method names must be verbs

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,162 @@
...
+            for (Node<T> i = head; i != null; i = i.next) {
+                if (object == i.item) {
+                    unlink(i);
+                    return true;
+                }
+            }
+            return false;
+        }
+        for (Node<T> i = head; i != null; i = i.next) {
+            if (object.equals(i.item)) {
```
Max-Pochepets, line 86: this can be combined in one loop with one if

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,162 @@
...
+                unlink(i);
+                return true;
+            }
+        }
         return false;
     }
 
+    private T unlink(Node<T> nodeToUnlink) {
```
Max-Pochepets, line 94: all private methods should be listed after public

Max-Pochepets (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -24,20 +24,22 @@ public void add(T value, int index) {
...
+                tail = valueToAdd;
+            } else if (nodeInIndex.prev == null && size > 0) {
+                valueToAdd = new Node<>(null, value, nodeInIndex);
+                nodeInIndex.prev = valueToAdd;
+                head = valueToAdd;
+            } else {
+                valueToAdd = new Node<>(findNodeByIndex(index).prev, value, findNodeByIndex(index));
+                valueToAdd.prev.next = valueToAdd;
+                valueToAdd.next.prev = valueToAdd;
+            }
```
Max-Pochepets, line 42: still, it can be done without `else if`

Max-Pochepets (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -31,13 +31,13 @@ public void add(T value, int index) {
...
                 tail = valueToAdd;
-            } else if (nodeInIndex.prev == null && size > 0) {
-                valueToAdd = new Node<>(null, value, nodeInIndex);
-                nodeInIndex.prev = valueToAdd;
-                head = valueToAdd;
             } else {
-                valueToAdd = new Node<>(findNodeByIndex(index).prev, value, findNodeByIndex(index));
-                valueToAdd.prev.next = valueToAdd;
+                valueToAdd = new Node<>(nodeInIndex.prev, value, nodeInIndex);
+                if (nodeInIndex.prev == null && size > 0) {
```
Max-Pochepets, line 36:
```suggestion
                if (nodeInIndex.prev == null) {
```
we have checked `size == 0` and it's false, and size cannot be negative.

Max-Pochepets (approved):

PR #340, opened 2021-08-31
linked-list-hw

MiaHrynda:
…t), get(int index), set(T value, int index), remove(int index), remove(T object), size(), isEmpty() methods of LinkedList.

Created my own methods checkIndex(int index), getNode(int index), getNode(T value), linkBefore(T item, Node<T> nextNode), unlink(Node<T> node) methods.

### DIFF
--- a/src/main/java/core/basesyntax/MyLinkedList.java
+++ b/src/main/java/core/basesyntax/MyLinkedList.java
@@ -3,45 +3,156 @@ package core.basesyntax;
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    class Node<T> {
+        private T item;
+        private Node<T> prev;
+        private Node<T> next;
+
+        public Node(Node<T> prev, T item, Node<T> next) {
+            this.prev = prev;
+            this.item = item;
+            this.next = next;
+        }
+    }
+
     @Override
     public void add(T value) {
+        Node<T> newNode = new Node<>(tail, value, null);
+        if (size > 0) {
+            tail.next = newNode;
+        } else {
+            head = newNode;
+        }
+        tail = newNode;
+        size++;
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,156 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    class Node<T> {
```
mishamolnar, line 10: Default aсcess modifier?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,156 @@
...
+        private T item;
+        private Node<T> prev;
+        private Node<T> next;
+
+        public Node(Node<T> prev, T item, Node<T> next) {
```
mishamolnar, line 15: Are U sure about public?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,156 @@
...
     }
 
     @Override
     public void add(T value, int index) {
+        if (index > size || index < 0) {
+            throw new IndexOutOfBoundsException("Invalid index");
+        }
+        if (index == size) {
+            add(value);
+            return;
```
mishamolnar, line 41: You can do this without return, but i`m not sure that this is necessary

mishamolnar (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
MiaHrynda: It doesn't work without return

MaliukDaria: I think we could leave `return` here

MaliukDaria (approved):
Nice 🔥

PR #341, opened 2021-08-31
Implemented java-linked-list task.

KyryloDvornyk:


MAFIOZNIK1999 (approved):

MAFIOZNIK1999 (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,170 @@
...
+        } else if (index == size - 1) {
+            indexNode = tail;
+            tail = new Node<>(indexNode.previous, value, null);
+            indexNode.previous.next = tail;
+        } else {
+            indexNode = getNode(index);
+            indexNode.previous.next = new Node<>(indexNode.previous, value, indexNode.next);
+            indexNode.next.previous = indexNode.previous.next;
+        }
+        return indexNode.value;
```
MaliukDaria, line 120: how could we simplify this method? Maybe we can use `Node<T> getNode(int index)` method here?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,170 @@
...
     }
 
     @Override
     public T remove(int index) {
```
MaliukDaria, line 124: checklist: Create a separate method unlink(Node node).

MaliukDaria (changes requested):
lets place private methods after public

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,167 @@
...
+            removedValue = tail.value;
+            tail = tail.previous;
+            tail.next = null;
+        } else {
+            removedValue = node.value;
+            node.previous.next = node.next;
+            node.next.previous = node.previous;
+        }
+        size--;
+        return removedValue;
```
MaliukDaria, line 157:
```suggestion
        T removedValue = node.value;
        Node<T> prev = node.previous;
        Node<T> next = node.next;
        if (prev == null) {
            first = next;
        } else {
            prev.next = next;
            x.prev = null;
        }
        if (next == null) {
            last = prev;
        } else {
            next.prev = prev;
            x.next = null;
        }
        x.item = null;
        size--;
        return removedValue;
```

MaliukDaria (changes requested):
private methods should follow after public

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
+            head = new Node<>(null, value, indexNode.next);
+            indexNode.next.previous = head;
+        } else if (index == size - 1) {
+            tail = new Node<>(indexNode.previous, value, null);
+            indexNode.previous.next = tail;
+        } else {
+            indexNode.previous.next = new Node<>(indexNode.previous, value, indexNode.next);
+            indexNode.next.previous = indexNode.previous.next;
+        }
+        return indexNode.value;
```
MaliukDaria, line 69:
```suggestion
        checkIndex(index);
        Node<T> indexNode = getNode(index);
        T oldValue = indexNode.value;
        indexNode.value = value;
        return oldValue;
```

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #344, opened 2021-09-01
Linked list hw

DenysST:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,147 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private static final String EXCEPTION_MESSAGE = "Not valid index";
```
denyssmei, line 6: May we change that constant value to more informative

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,147 @@
...
     @Override
     public T remove(int index) {
-        return null;
+        return unlink(getNode(index));
     }
 
     @Override
     public boolean remove(T object) {
+        for (int i = 0; i < size(); i++) {
+            T currentElement = getNode(i).element;
```
denyssmei, line 79: Maybe put variable  anouncment out of for cycle. Mb after line 77

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,147 @@
...
+    }
+
+    private boolean indexIsValid(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException(EXCEPTION_MESSAGE);
+        }
+        return true;
+    }
+
+    private void addToTheMiddle(T value, int index) {
```
denyssmei, line 106: This method used only once. No reason to put this logick in separate method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,147 @@
...
+    private Node<T> getNode(int index) {
+        indexIsValid(index);
+        Node<T> current = head;
+        for (int i = 0; i < index; i++) {
+            current = current.next;
+        }
+        return current;
+    }
+
+    private T unlink(Node<T> someNode) {
```
denyssmei, line 126: Same. No reasone to make lock in separate method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,147 @@
...
+            Node<T> nextNode = new Node<>(tail, value, null);
+            tail.next = nextNode;
+            tail = nextNode;
+        }
     }
 
     @Override
     public void add(T value, int index) {
+        indexIsValidAdd(index);
+        if (index == size()) {
```
denyssmei, line 40: instead of size() use variable size

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,147 @@
...
 
     @Override
     public T remove(int index) {
-        return null;
+        return unlink(getNode(index));
     }
 
     @Override
     public boolean remove(T object) {
+        for (int i = 0; i < size(); i++) {
```
denyssmei, line 78: Same with size()

denyssmei (changes requested):

denyssmei (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,148 @@
...
+            tail = nextNode;
+        }
     }
 
     @Override
     public void add(T value, int index) {
+        indexIsValidAdd(index);
+        if (index == size) {
+            add(value);
+        } else if (index == 0) {
```
MaliukDaria, line 42:
```suggestion
            add(value);
            return;
        } 
        if (index == 0) {
```

it's hard to read complicated `if-else` constructions

src/main/java/core/basesyntax/MyLinkedList.java
MaliukDaria: I agree, you don't have separate methods for `addToTheEnd` and `addToTheBeginning`, so you have to implement all 3 methods and use them or don't use separate methods at all

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,148 @@
...
+        }
+        return true;
+    }
+
+    private void addToTheMiddle(T value, int index) {
+        Node<T> current = head;
+        int count = 0;
+        while (count++ != index) {
+            current = current.next;
+        }
```
MaliukDaria, line 112: meybe we can use our `getNode(int index)` method here?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,148 @@
...
     public boolean remove(T object) {
+        T currentElement;
+        for (int i = 0; i < size; i++) {
+            currentElement = getNode(i).element;
+            if ((currentElement == object)
+                    || (currentElement != null && currentElement.equals(object))) {
+                remove(i);
+                return true;
+            }
+        }
```
MaliukDaria, line 86: can we implement this method without using a method `getNode(i)`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,148 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private boolean indexIsValid(int index) {
```
MaliukDaria, line 100: lets make this method void

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,148 @@
...
+        int count = 0;
+        while (count++ != index) {
+            current = current.next;
+        }
+        Node<T> newNode = new Node<>(current.prev, value, current);
+        newNode.prev.next = newNode;
+        newNode.next.prev = newNode;
+    }
+
+    private Node<T> getNode(int index) {
```
MaliukDaria, line 118: checklist: Don’t iterate your nodes from the head to tail or vice versa.
Pay attention to the index you receive as an input, you can iterate only first or second half of the list, depending on the index value. That will boost your lists’ performance.

src/main/java/core/basesyntax/MyLinkedList.java
MaliukDaria: We should have `unlink` as separate method: Checklist: Create a separate method unlink(Node node).

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,148 @@
...
+        if (nextNode == null) {
+            tail = prevNode;
+        } else {
+            nextNode.prev = prevNode;
+        }
+        size--;
+        return someNode.element;
+    }
+
+    private boolean indexIsValidAdd(int index) {
```
MaliukDaria, line 144: lets make it void

MaliukDaria (changes requested):

MaliukDaria (approved):
🚀

PR #352, opened 2021-09-01
Task completed

TheReaper42:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,153 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size <= 0;
+    }
+
+    private boolean indexIsValidAdd(int index) {
```
ditroRV, line 113: This method is used only once, so we could check if the index is valid inside add method.

ditroRV (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,139 @@
...
+        if (index == size()) {
+            add(value);
+        } else if (index == 0) {
+            addToTheBegining(value);
+        } else {
+            addToTheMiddle(value, index);
+        }
+    }
+
+    private void addToTheBegining(T value) {
```
MaliukDaria, line 37: private methods should follow after public

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,139 @@
...
+        head.prev = newNode;
+        head = newNode;
+    }
+
+    private void addToTheMiddle(T value, int index) {
+        Node<T> current = head;
+        int count = 0;
+        while (count++ != index) {
+            current = current.next;
+        }
```
MaliukDaria, line 48: could we use here `getNode(int index)` method?

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #354, opened 2021-09-01
task completed

Andruwa808:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,140 @@
...
+        }
+        Node<T> node = getNode(index);
+        Node<T> newNode = new Node<>(node.prev, value, node);
+        if (index == 0) {
+            first = newNode;
+        } else {
+            node.prev.next = newNode;
+        }
+        node.prev = newNode;
+        size++;
```
andrii-demchenko, line 49: it is better to implement this part without else
```suggestion
        if (index == 0) {
            first = newNode;
            size++;
            return;
        }
        node.prev.next = newNode;
        node.prev = newNode;
        size++;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,140 @@
...
     public void addAll(List<T> list) {
+        for (T item : list) {
+            add(item);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkIndex(index);
```
andrii-demchenko, line 61: no need to use this method here. You have already implemented it in getNode method, which is called on the next line.
```suggestion
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,140 @@
...
+        T oldValue = getNode(index).item;
+        getNode(index).item = value;
+        return oldValue;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        Node<T> node = getNode(index);
+        return unlink(node);
```
andrii-demchenko, line 75:
```suggestion
        return unlink(getNode(index));
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,140 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void checkIndexForAdd(int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Index does not exist: " + index);
```
andrii-demchenko, line 101: Maybe it is better to make this message a CONSTANT?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,140 @@
...
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Index does not exist: " + index);
+        }
+    }
+
+    private Node<T> getNode(int index) {
+        checkIndex(index);
+        if (index > (size >> 1)) {
+            Node<T> tempNode = first;
```
andrii-demchenko, line 114: No need to declare two different nodes in both if and else methods. You can declare it once after checkIndex.
Also maybe it will be better to write size / 2 in if condition?

andrii-demchenko (changes requested):

andrii-demchenko (approved):
Good job

MaliukDaria (approved):
👍🏻

PR #361, opened 2021-09-26
added solution

YaroslavYarmoshyk:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,190 @@
...
+        public Node(Node<T> prev, T item, Node<T> next) {
+            this.prev = prev;
+            this.item = item;
+            this.next = next;
+        }
+    }
+
     @Override
     public void add(T value) {
+        linkLast(value);
```
RomanDubovskyi, line 24: currently add() method doesn't have any logic of its own. Better solution will be to move code from linkLast to add()

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,190 @@
...
     }
 
     @Override
     public void add(T value, int index) {
+        checkIndex(index);
+        if (index == size) {
+            linkLast(value);
```
RomanDubovskyi, line 31:
```suggestion
            add(value);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,190 @@
...
+        } else {
+            linkBefore(value, node(index));
+        }
     }
 
     @Override
     public void addAll(List<T> list) {
+        Object[] a = list.toArray();
```
RomanDubovskyi, line 39: just iterate list fir loop and call method `add()` for each element

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,190 @@
...
+            pred = newNode;
+        }
+        tail = pred;
+        size += numNew;
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index < 0) {
```
RomanDubovskyi, line 62: this if else logic is not ovbious, let's make more predictable behaviour
```suggestion
        checkindex(index);
        return node(index).item;
```
Use this check in all methods except add(index, value). In add write custom check or pass some kind of boolean to know whether to check index exclusively or not

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,190 @@
...
+            checkIndex(index);
+        } else {
+            checkIndex(index + 1);
+        }
+        return unlink(node(index));
     }
 
     @Override
     public boolean remove(T object) {
+        if (object == null) {
```
RomanDubovskyi, line 95: this method can be shortened, use single loop for iteration. Inside use the same check you used in arr list and storage

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,190 @@
...
+        tail = newNode;
+        if (t == null) {
+            head = newNode;
+        } else {
+            t.next = newNode;
+        }
+        size++;
+    }
+
+    void linkBefore(T e, Node<T> succ) {
```
RomanDubovskyi, line 135: why do we use default access modifier here?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,190 @@
...
+        } else {
+            Node<T> x = tail;
+            for (int i = size - 1; i > index; i--) {
+                x = x.prev;
+            }
+            return x;
+        }
+    }
+
+    T unlink(Node<T> x) {
```
RomanDubovskyi, line 169: same herer

RomanDubovskyi (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,153 @@
...
-        return null;
+        checkIndex(index, true);
+        return unlink(node(index));
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> x = head;
+        for (int i = 0; i < size; i++) {
+            if (x.item == object || x.item.equals(object)) {
```
RomanDubovskyi, line 77:
```suggestion
            if (x.item == object || x.item != null && x.item.equals(object)) {
```
it's better to add one more check in case item == null and object != null

RomanDubovskyi (approved):

PR #362, opened 2021-09-27
added solution

DmytroT-dev:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,197 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size;
+    private Node<T> first;
+    private Node<T> last;
+
+    class Node<T> {
```
RomanDubovskyi, line 10: why do we use default access modifier here?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,197 @@
...
+        private Node<T> prev;
+
+        public Node(Node<T> prev, T value, Node<T> next) {
+            this.prev = prev;
+            this.value = value;
+            this.next = next;
+        }
+    }
+
+    private void checkIndex(int index) {
```
RomanDubovskyi, line 22: private methods should be located after public ones

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,197 @@
...
+        if (index > size || index < 0) {
+            throw new IndexOutOfBoundsException("Invalid index. "
+                + "Index is out of bounds of array");
+        }
+    }
+
+    private void checkIndexWithEquals(int index) {
```
RomanDubovskyi, line 29:
```suggestion
    private void checkIndexInclusevily(int index) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,197 @@
...
+            for (int i = size - 1; i > index; i--) {
+                x = x.prev;
+            }
+            return x;
+        }
+    }
+
     @Override
     public void add(T value) {
+        Node<T> l = last;
```
RomanDubovskyi, line 56: we can remove this variable, last is enough

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,197 @@
...
+            add(value);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkIndexWithEquals(index);
+
+        if (index < (size >> 1)) {
```
RomanDubovskyi, line 100: let's use getElementByIndex here

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,197 @@
...
+                x = x.prev;
+            }
+            return x.value;
+        }
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        checkIndexWithEquals(index);
```
RomanDubovskyi, line 117: index is checked in getElement we don't need to do it twice

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,197 @@
...
+        current = null;
+        size--;
+
+        return oldValue;
     }
 
     @Override
     public boolean remove(T object) {
+        int k = 0;
+        if (object == null) {
```
RomanDubovskyi, line 169: let's have single for loop here. We just need to improve if check so it can compare both null and non null values

RomanDubovskyi (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
dmitriy-tverdokhlibov: Yeah, because I'don't write

src/main/java/core/basesyntax/MyLinkedList.java
dmitriy-tverdokhlibov: Ok

src/main/java/core/basesyntax/MyLinkedList.java
dmitriy-tverdokhlibov: Roman, thanks for review!

src/main/java/core/basesyntax/MyLinkedList.java
dmitriy-tverdokhlibov: We need to use encapsulation

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,160 @@
...
+
+        return oldValue;
     }
 
     @Override
     public boolean remove(T object) {
+        int k = 0;
+        for (Node<T> x = first; x != null; x = x.next) {
+            if ((object == null && x.value == null)
+                    || object != null && object.equals(x.value)) {
```
RomanDubovskyi, line 120:
```suggestion
            if (object == x.value || object != null && object.equals(x.value)) {
```

RomanDubovskyi (approved):

src/main/java/core/basesyntax/MyLinkedList.java
RomanDubovskyi: > Roman, thanks for review!

you're welcome 🤗

PR #363, opened 2021-10-02
Home_Work is done

romanNov-creator:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,177 @@
...
+            Node<T> node = new Node<>(null, value, next);
+            if (size == 0) {
+                tail = node;
+            }
+            head = node;
+            if (size > 0) {
+                next.prev = node;
+            }
+        } else if (index == size) {
+            Node<T> prev = tail;
```
RomanDubovskyi, line 51: we can call here `add(value)` and return after

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,177 @@
...
+            Node<T> nextNode = foundNode.next;
+            Node<T> node = new Node<>(foundNode, value, nextNode);
+            foundNode.next = node;
+            nextNode.prev = node;
+        }
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
```
RomanDubovskyi, line 66: you can simply call forEach and method `add()` inside

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,177 @@
...
-        return null;
+        checkIndex(index);
+        T element = getNodeByIndex(index).element;
+        unlink(getNodeByIndex(index));
+        return element;
     }
 
     @Override
     public boolean remove(T object) {
+        if (object == null) {
```
RomanDubovskyi, line 106: let's improve `if()` check inside of for loop and we wouldn't need 2 loops

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,177 @@
...
+
+    private void unlink(Node<T> node) {
+        Node<T> prevNode = node.prev;
+        Node<T> nextNode = node.next;
+        if (size == 1) {
+            head = null;
+            tail = null;
+            size--;
+            return;
+        } else if (prevNode == null) {
```
RomanDubovskyi, line 142: we don't need else block when our if block ends with return statement

RomanDubovskyi (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,162 @@
...
+        checkIndex(index);
+        T element = getNodeByIndex(index).element;
+        unlink(getNodeByIndex(index));
+        return element;
     }
 
     @Override
     public boolean remove(T object) {
+        for (Node<T> i = head; i != null; i = i.next) {
+            if (i.element == null || i.element.equals(object)) {
```
RomanDubovskyi, line 99:
```suggestion
            if (i.element == object || i.element != null && i.element.equals(object)) {
```
current check won't work correctly if i.element == null and object != null

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,162 @@
...
 
     @Override
     public void addAll(List<T> list) {
+        Object[] arr = list.toArray();
+        if (arr.length == 0) {
+            return;
+        }
+        for (Object obj : arr) {
+            add((T) obj);
+        }
```
RomanDubovskyi, line 71:
```suggestion
        for (T obj : list) {
            add(obj);
        }
```
this check is redundant, if length is 0 we will just skip the loop

RomanDubovskyi (changes requested):

RomanDubovskyi (approved):

PR #368, opened 2021-10-29
LinkedList jv-linked-list DzmitryKozel

DmitroKoz:


src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,23 @@
...
+        System.out.println(myLinkedList.get(2));
+        System.out.println(myLinkedList.get(3));
+        System.out.println(myLinkedList.get(4));
+        System.out.println(myLinkedList.remove(3));
+
+        System.out.println(myLinkedList.get(2));
+        System.out.println(myLinkedList.get(3));
+        System.out.println(myLinkedList.get(4));
+    }
+}
```
QQBratka, line 23: Don't use main method here

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,155 @@
...
+    private Node<T> tail;
+
+    private class Node<T> {
+        private T value;
+        private Node<T> prev;
+        private Node<T> next;
+
+        Node(Node<T> prev, T x, Node<T> next) {
+            this.prev = prev;
+            value = x;
```
QQBratka, line 17: use this

QQBratka (approved):

src/main/java/core/basesyntax/MyLinkedList.java
Sofasmile: @QQBratka  we should use this only if the name of parameters are the same, in this code, we have different variable names, so we can skip `this` keyword

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,155 @@
...
+        return oldValue;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index, "remove");
+        if (isEmpty()) {
+            return null;
+        }
```
Sofasmile, line 81: do we meed this check here?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,155 @@
...
+        Node<T> nodeToRemove = findNodeByIndex(index);
+        unLink(nodeToRemove);
+        return nodeToRemove.value;
     }
 
     @Override
     public boolean remove(T object) {
+        if (isEmpty()) {
+            return false;
+        }
```
Sofasmile, line 91: we can remove this check too

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,155 @@
...
+        if (node.prev == null && node.next == null) {
+            size--;
+        } else if (node.prev == null) {
+            head = node.next;
+            head.prev = null;
+            size--;
+        } else if (node.next == null) {
+            tail = node.prev;
+            tail.next = null;
+            size--;
```
Sofasmile, line 125: can we remove duplication of `size--` from each block and do it only once?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,155 @@
...
+            for (int i = size - 1; i > index; --i) {
+                soughtNode = soughtNode.prev;
+            }
+        }
+        return soughtNode;
+    }
+
+    private void checkIndex(int index, String operationType) throws IndexOutOfBoundsException {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Can't " + operationType
```
Sofasmile, line 153: better pass index to the exception instead of operationType, because operationType we can get from stacktrace

Sofasmile (changes requested):
great job)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,155 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size = 0;
```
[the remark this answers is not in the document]
DmitroKoz: Not necessary cause of its default is 0...

src/main/java/core/basesyntax/MyLinkedList.java
DmitroKoz: Seems nope cause of findNodeByIndex(index) may return null in some cases

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,149 @@
...
+        return size == 0;
+    }
+
+    private void unLink(Node<T> node) {
+        Node<T> previousNode;
+        Node<T> nextNode;
+        if (node.prev == null && node.next == null) {
+            size--;
+            return;
+        } else if (node.prev == null) {
```
Sofasmile, line 114: we can remove `else` because we have return in the previous block
```suggestion
        } 
        if (node.prev == null) {
```

Sofasmile (changes requested):
small comment))

Sofasmile (approved):

PR #384, opened 2021-11-02
In the MyLinkedList class methods add, addAll, get, set, remove, size…

MykolaVoronov:
…, isEmpty were implemented, findNodeByIndex and unlink methods and class Node<E> were added

### DIFF
--- a/src/main/java/core/basesyntax/MyLinkedList.java
+++ b/src/main/java/core/basesyntax/MyLinkedList.java
@@ -24,7 +24,7 @@ public class MyLinkedList<T> implements MyLinkedListInterface<T> {
 
     @Override
     public void add(T value, int index) {
-        if (index >= 0 && size == 0 || index == size) {
+        if (index == size) {
             add(value);
             return;
         }


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,142 @@
...
     @Override
     public void add(T value) {
+        if (size == 0) {
+            Node<T> firstNode = new Node<>(null, value, null);
+            head = firstNode;
+            tail = firstNode;
+            size++;
+            return;
+        }
+        tail.next = new Node<>(tail, value, null);
```
ShurionUA, line 22: Mb, better create new object for easy reading

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,142 @@
...
+        tail = tail.next;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index >= 0 && size == 0 || index == size) {
```
ShurionUA, line 29: If size = 0, and we want to add element on index = 5 - is this the correct index to add element?

MykolaVoronov: This is checked in the findNodeByIndex method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,142 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private Node<T> findNodeByIndex(int index) {
+        if (index < 0 || index >= size) {
```
ShurionUA, line 91: We have at least two index checks - mb we need separate method for index check?

src/main/java/core/basesyntax/MyLinkedList.java
ShurionUA: When we call method add, with idex = 5 and size = 0, value will be added without calling findNodeByIndex.

src/main/java/core/basesyntax/MyLinkedList.java
MykolaVoronov: In the first case, the index is checked for the next steps of the add method. In the second, the index is checked for valid values. So there is no sence in introducing a new method for checking the index

src/main/java/core/basesyntax/MyLinkedList.java
MykolaVoronov: image

ShurionUA (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,140 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    public MyLinkedList() {
```
Sofasmile, line 10: do we need empty constructor?

Sofasmile (approved):

src/main/java/core/basesyntax/MyLinkedList.java
MykolaVoronov: No, it's not necessary

PR #398, opened 2021-11-03
added solution

Oleksii-kho:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,144 @@
...
+    private void validIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Index " + index + " out off list length " + size);
+        }
+    }
+
+    private Node<T> findNodeByIndex(int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Index " + index + " out off list length " + size);
+        }
```
farfardesu, line 111: modify yout validIndex() method and use it instead of hardcoding
for example change method signature to "private Node<T> findNodeByIndex(int index, int limit)"

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,144 @@
...
+        private T item;
+        private Node<T> next;
+        private Node<T> prev;
+
+        Node(Node<T> prev, T element, Node<T> next) {
+            this.item = element;
+            this.next = next;
+            this.prev = prev;
+        }
+    }
```
farfardesu, line 100: may be it will be better if this class will be in separate file, but I am not sure

farfardesu (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
Oleksii-kho: Inner class should be used here

src/main/java/core/basesyntax/MyLinkedList.java
Oleksii-kho: Such a check is used only in this method and it is more visual. Then findNodeByIndex does not need to pass two parameters.

farfardesu (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,144 @@
...
     @Override
     public void add(T value) {
+        if (isEmpty()) {
+            head = new Node<>(null, value, null);
+            tail = head;
+        } else {
+            Node<T> newNode = new Node<>(tail, value, null);
+            tail.next = newNode;
+            tail = newNode;
+        }
```
Nick97-git, line 19:
```suggestion
        Node<T> newNode = new Node<>(tail, value, null);
        if (isEmpty()) {
            head = newNode;
        } else {
            tail.next = newNode;
        }
        tail = newNode;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,144 @@
...
+        private Node<T> prev;
+
+        Node(Node<T> prev, T element, Node<T> next) {
+            this.item = element;
+            this.next = next;
+            this.prev = prev;
+        }
+    }
+
+    private void validIndex(int index) {
```
Nick97-git, line 102:
```suggestion
    private void checkIndex(int index) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,144 @@
...
+                node.next.prev = null;
+                head = node.next;
+            } else if (node.equals(tail)) {
+                node.prev.next = null;
+                tail = node.prev;
+            } else {
+                node.prev.next = node.next;
+                node.next.prev = node.prev;
+            }
+        }
```
Nick97-git, line 142:
```suggestion
        if (size == 1) {
            head = null;
        } else if (node.equals(head)) {
                node.next.prev = null;
                head = node.next;
            } else if (node.equals(tail)) {
                node.prev.next = null;
                tail = node.prev;
            } else {
                node.prev.next = node.next;
                node.next.prev = node.prev;
            }
        }
```

Nick97-git (changes requested):

Nick97-git (approved):

PR #421, opened 2021-12-23
Created my own LinkedList with overriding and some new methods

YevhenStelmakh:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
 
     @Override
     public void add(T value, int index) {
+        checkIndexForAdding(index);
+        if (index == size) {
+            add(value);
+        } else {
+            final Node<T> beforeNewNode = findNodeOnIndex(index).prev;
+            final Node<T> newNode = new Node<>(beforeNewNode,value, findNodeOnIndex(index));
+            findNodeOnIndex(index).prev = newNode;
```
MaliukDaria, line 44: Can we use findNodeOnIndex(index) only ones here? findNodeOnIndex it's quite laborious operation, it should not be performed several times unnecessarily

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
+            } else {
+                beforeNewNode.next = newNode;
+            }
+            size++;
+        }
     }
 
     @Override
     public void addAll(List<T> list) {
+        Object[] objects = list.toArray();
```
MaliukDaria, line 56: Why do we need to convert list to array here?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
 
     @Override
     public T get(int index) {
-        return null;
+        checkIndexForSettingAndRemoving(index);
+        Node<T> value = head;
+        for (int i = 0; i < index; i++) {
+            value = value.next;
+        }
+        return (T)value.element;
```
MaliukDaria, line 69: Can we use your `findNodeOnIndex()` method here?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
+                }
+            }
+        } else {
+            for (Node<T> x = head; x != null; x = x.next) {
+                if (object.equals(x.element)) {
+                    unlink(x);
+                    return true;
+                }
+            }
+        }
```
MaliukDaria, line 103: can we do this without if-else construction? Just with one for loop?

MaliukDaria (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,151 @@
...
+                if (x.element == null) {
+                    unlink(x);
+                    return true;
+                }
+            }
+            if (object != null && object.equals(x.element)) {
+                unlink(x);
+                return true;
+            }
+        }
```
MaliukDaria, line 96:
```suggestion
        for (Node<T> x = head; x != null; x = x.next) {
            if (x.element == object || (object != null && object.equals(x.element))) {
                unlink(x);
                return true;
            }
        }
```

MaliukDaria (changes requested):

RomanDubovskyi (approved):

PR #438, opened 2022-01-26
realisation LinkedList (first version)

LinkClink:
realisation methods add, addAll, set, remove, get, size, IsEmpty

### DIFF
--- a/src/main/java/core/basesyntax/MyLinkedList.java
+++ b/src/main/java/core/basesyntax/MyLinkedList.java
@@ -3,45 +3,148 @@ package core.basesyntax;
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size;
+    private Node<T> last;
+    private Node<T> first;
+
     @Override
     public void add(T value) {
+        addNew(value);
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,148 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size;
+    private Node<T> last;
+    private Node<T> first;
+
     @Override
     public void add(T value) {
+        addNew(value);
```
Liubov-Chernohor, line 12: Maybe we don't need the additional method "addNew(T value)" if in method "add(T value)" we call only it. We can move logic from  "addNew(T value)" to "add(T value)".

Liubov-Chernohor (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
LinkClink: Yea, you right, thanks.)

Liubov-Chernohor (approved):

AlexRozv (approved):
great job 🔥

PR #441, opened 2022-01-28
created MyLinkedList class

liliia-ponomarenko:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,171 @@
...
     public boolean remove(T object) {
+        if (object == null) {
+            for (Node<T> x = first; x != null; x = x.next) {
+                if (x.value == null) {
+                    unlink(x);
+                    return true;
+                }
+            }
+        } else {
+            for (Node<T> x = first; x != null; x = x.next) {
```
anyutamysko, line 141: Node<T> result;
while(x != null) {
    if (x.value == object || (x.value != null && x.value.equals(object))) {
     result = x;
    }
}
result != null ? unlink(result) return true : false;

anyutamysko (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> first;
+    private Node<T> last;
+    private int size;
+
+    private void linkLast(T value) {
+        final Node<T> l = last;
+        final Node<T> newNode = new Node<>(l, value, null);
```
AlexRozv, line 12: instead of "l" let's use more informative naming

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
...
+    }
+
+    private void checkIndex(int index) {
+        if (index >= size || index < 0) {
+            throw new IndexOutOfBoundsException("Illegal index: " + index);
+        }
+    }
+
+    private Node<T> node(int index) {
+        checkIndex(index);
```
AlexRozv, line 41: let's also rename this method as well to make it more informative, for example "get(int index)"

AlexRozv (changes requested):

AlexRozv (approved):
well done 👍

PR #442, opened 2022-01-28
Implementhed own realisation of LinkedList. mvn build success. Need s…

DaniilOpryshko:


pikador10 (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,169 @@
...
 
     @Override
     public T set(T value, int index) {
-        return null;
+        if (checkIndex(index)) {
+            T oldValue = getNodeOfIndex(index).element;
+            getNodeOfIndex(index).element = value;
+            return oldValue;
+        }
+        throw new IndexOutOfBoundsException();
```
AlexRozv, line 74: let's throw IndexOutOfBoundsException inside checkIndex method, then you will be able to remove this if block
do the same in getNodeOfIndex() method

AlexRozv (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,156 @@
...
```
AlexRozv, line 73: nw when you're throwing exception inside the check index method we don't have to throw it here and we don't need this if block. this will work because if index is invalid the exception will be thrown and the method will be stopped

```suggestion
        checkIndex(index);
        T oldValue = getNodeOfIndex(index).element;
        getNodeOfIndex(index).element = value;
        return oldValue;
```

AlexRozv (changes requested):

AlexRozv (approved):
good job 👍

PR #445, opened 2022-01-29
hw9 solved add some methods

AprioriKS:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,157 @@
...
+        private Node(Node<T> prev, T element, Node<T> next) {
+            this.value = element;
+            this.prev = prev;
+            this.next = next;
+        }
+    }
+
     @Override
     public void add(T value) {
+        if (isEmpty()) {
```
Sasha-Bachynskyi, line 25: Does isEmpty() without object return always false?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,157 @@
...
 
     @Override
     public T set(T value, int index) {
-        return null;
+        if (checkingIndex(index)) {
+            T old = getNode(index).value;
+            getNode(index).value = value;
+            return old;
+        }
+        throw new IndexOutOfBoundsException();
```
Sasha-Bachynskyi, line 74: why don't you remove your exception into checkingIndex method?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,157 @@
...
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (index >= 0 && index < size) {
```
Sasha-Bachynskyi, line 79: it would be more logical to create check index method then remove a condition and an exception into there

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,157 @@
...
+            size--;
+            return removedElement;
+        }
+        throw new IndexOutOfBoundsException();
     }
 
     @Override
     public boolean remove(T object) {
-        return false;
+        try {
```
Sasha-Bachynskyi, line 90: i can't get, why you used try catch here)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,157 @@
...
+                for (int i = size - 1; i > index; i--) {
+                    node = node.prev;
+                }
+                return node;
+            }
+        }
+        throw new IndexOutOfBoundsException();
+    }
+
+    private int getIndex(T object) throws NoSuchElementException {
```
Sasha-Bachynskyi, line 127: you don't need to throw your exception up here, do you?

Sasha-Bachynskyi (changes requested):

Sasha-Bachynskyi (approved):
Good job)

src/main/java/core/basesyntax/MyLinkedList.java
AlexRozv: agree, we don't need it here

AlexRozv (changes requested):

AlexRozv (approved):
well done 👌

PR #447, opened 2022-01-30
Own implementation of LinkedList has been done

Liubov-Chernohor:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
+        } else if (index == 0) {
+            addFirst(value);
+        } else if (isIndexValid(index)) {
+            Node<T> currentNode = searchNodeByIndex(index);
+            Node<T> node = new Node<T>(currentNode.prev, value, currentNode);
+            currentNode.prev.next = node;
+            currentNode.prev = node;
+            size++;
+        } else {
+            throw new IndexOutOfBoundsException("Index is invalid");
```
Sasha-Bachynskyi, line 35: why don't you remove your exception in isIndexValid() method?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (isIndexValid(index)) {
+            return searchNodeByIndex(index).value;
+        } else {
+            throw new IndexOutOfBoundsException("Index is invalid");
```
Sasha-Bachynskyi, line 51: here, it seems logical, but I am not fully confident. Would you agree to remove this error in isIndexValid() method, so your exception wouldn't repeat?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
 
     @Override
     public T remove(int index) {
-        return null;
+        if (isIndexValid(index)) {
+            Node<T> currentNode = searchNodeByIndex(index);
+            unlink(currentNode);
+            return currentNode.value;
+        } else {
+            throw new IndexOutOfBoundsException("Index is invalid");
```
Sasha-Bachynskyi, line 70: the same question

Sasha-Bachynskyi (changes requested):
Good job, but let me know your thoughts about exceptions

src/main/java/core/basesyntax/MyLinkedList.java
Liubov-Chernohor: you are right!!! Thanks!

src/main/java/core/basesyntax/MyLinkedList.java
Liubov-Chernohor: done

Sasha-Bachynskyi (approved):
Looks very good =)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,159 @@
...
+            tail = node.prev;
+            node.prev.next = null;
+            node.prev = null;
+            size--;
+        } else {
+            node.prev.next = node.next;
+            node.next.prev = node.prev;
+            node.prev = null;
+            node.next = null;
+            size--;
```
AlexRozv, line 146: can we simplify this?
for example, you're always decreasing size, you can make it after all if-else blocks

AlexRozv (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
Liubov-Chernohor: Yes, you're right! Done!

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,157 @@
...
+        } else if (node == tail) {
+            tail = node.prev;
+            node.prev.next = null;
+            node.prev = null;
+        } else {
+            node.prev.next = node.next;
+            node.next.prev = node.prev;
+            node.prev = null;
+            node.next = null;
+        }
```
AlexRozv, line 144: it still looks a little too complicated, let's try to use two if-else blocks here
tip: we can create temporary nodes node.prev and node.next, then do two checks:
 if (prev == null) {...} 
 else {...}
 if (next == null) {...}
 else {...}

AlexRozv (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
Liubov-Chernohor: Yep, now it looks better))

AlexRozv (approved):
good job 👍

PR #449, opened 2022-01-31
Develop

OleksiyTitov:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,200 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private static class Node<T> {
```
dima-semeniuk, line 191: I think you should put class Node at the beginning of the code

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,200 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    Node<T> node(int index) {
```
dima-semeniuk, line 10: this method should be private

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,200 @@
...
+        } else {
+            Node<T> currentNode = tail;
+            for (int i = size - 1; i > index; i--) {
+                currentNode = currentNode.prev;
+            }
+            return currentNode;
+        }
+    }
+
+    void linkLast(T element) {
```
dima-semeniuk, line 26: These methods should be private

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,200 @@
...
+        final MyLinkedList.Node<T> newNode = new MyLinkedList.Node<>(tail, element, null);
```
dima-semeniuk, line 27: You can create nested class Node without calling outside class MyLinkedList

dima-semeniuk (changes requested):

dima-semeniuk (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,200 @@
...
+        private Node<T> prev;
+
+        Node(Node<T> prev, T element, Node<T> next) {
+            this.item = element;
+            this.next = next;
+            this.prev = prev;
+        }
+    }
+
+    private Node<T> node(int index) {
```
AlexRozv, line 22: I think it's better to move this logic to get() method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,200 @@
...
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (!isValidIndex(index)) {
+            throw new IndexOutOfBoundsException("Index: " + index + ", Size: " + size);
+        }
```
AlexRozv, line 144: it's will be better if you throw your exception inside the isValidIndex(index) method, this will allow us to remove this if check and we don't have to throw exception here as well, do it for all methods where you use isValidIndex(index)

AlexRozv (changes requested):

AlexRozv (approved):
well done 👍

PR #453, opened 2022-01-31
LinkedList was implemented

nastyaprokhorenko:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,142 @@
...
+            Node<T> newNode = new Node<T>(tail, value, null);
+            tail.next = newNode;
+            tail = newNode;
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if ((index == 0 && isEmpty()) || index == size) {
```
dima-semeniuk, line 37: I think you should remove (Index == 0 && isEmpty) and leave only check (index == size) 
```suggestion
        if(index == size) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,142 @@
...
+            add(value);
+        } else if (index == 0 && !isEmpty()) {
```
dima-semeniuk, line 39: here is enough to check index == 0

dima-semeniuk (approved):

AlexRozv (approved):
great job 🔥

PR #454, opened 2022-02-01
Impleament overriding LinkedList

ValeraLemberov:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,155 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Note<T> first;
```
DzhuzhaMichael, line 6: Maybe you should rename this and other same fields/variables "NoDe" not "NoTe"  (at the moment you have both Note and Node on your code). Also, according to checklist we should make NoDe as inner class.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,155 @@
...
+    private Note<T> last;
+    private int size;
+    private Note<T> boofNode;
+    private Note<T> newNode;
+
     @Override
     public void add(T value) {
+        if (size == 0) {
```
DzhuzhaMichael, line 14: You have method isEmpty() that returns you the same boolean result. You can use it here.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,155 @@
...
+        int count = 1;
+        while (count <= index) {
+            boofNode = boofNode.next;
+            count++;
+        }
+        return boofNode;
+    }
+
+    private void linkFirst(T value) {
+        Note<T> f = first;
```
DzhuzhaMichael, line 114: According to the checklist we shoul create informative variable names. Maybe you should change this and other single letter names to words.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,155 @@
...
+        return size == 0;
+    }
+
+    private Note<T> findeIndex(int index) {
+        boofNode = first;
+        if (index == 0) {
+            return boofNode;
+        }
+        int count = 1;
+        while (count <= index) {
```
DzhuzhaMichael, line 106: Depending to the index you receive as an input, you can iterate only first or second half of the list (checklist). Maybe it can be implemented in your case as well.

src/main/java/core/basesyntax/MyLinkedList.java
DzhuzhaMichael, line 114: According to the checklist we should create informative variable names. So, you can rename your single letter names here and in other parts of the code.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,155 @@
...
+        } else {
+            boofNode.prev.next = boofNode.next;
+            boofNode.next.prev = boofNode.prev;
+        }
+        size--;
+        return removedValue;
     }
 
     @Override
     public boolean remove(T object) {
```
DzhuzhaMichael, line 79: According to the checklist we should create a separate method unlink(Node node). This logic should be used in 2 methods: when we remove from list by index and by value. As far as I understand you doesn`t ulink this nodes. I don`t know is it critical or not for approwal.

DzhuzhaMichael, line 79: According to the checklist we should create a separate method unlink(Node node). This logic should be used in 2 methods: when we remove from list by index and by value. As far as I understand you do not ulink this nodes. I do not know is it critical or not for approwal.

DzhuzhaMichael (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,169 @@
...
+        if (index < halfSize) {
+            boofNode = first;
+            if (index == 0) {
+                return boofNode;
+            }
+            int count = 1;
+            while (count <= index) {
+                boofNode = boofNode.next;
+                count++;
+            }
```
AlexRozv, line 111: this if check is redundant
and instead of while loop with counter let's use for i = 0 ...

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,169 @@
...
+        } else {
+            boofNode = last;
+            if (index == size - 1) {
+                return boofNode;
+            }
+            int count = size - 1;
+            while (count > index) {
+                boofNode = boofNode.prev;
+                count--;
+            }
```
AlexRozv, line 122: same here

AlexRozv (changes requested):

AlexRozv (approved):
well done 👌

PR #457, opened 2022-02-01
Classes implemented. Tests passed. Codestyle checked.

lugburz-hr:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,196 @@
...
+        return currentNode;
+    }
+
+    private void checkIndex(int index, boolean isStrict) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("IndexOutOfBoundsException " + index);
+        }
+        if (!isStrict && index == size) {
+            throw new IndexOutOfBoundsException("IndexOutOfBoundsException " + index);
+        }
```
sviat-h, line 79: I think we can combine these two 'if' blocks 
```suggestion
        if (index < 0 || index > size || !isStrict && index == size) {
            throw new IndexOutOfBoundsException("IndexOutOfBoundsException " + index);
        }
```

sviat-h (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,193 @@
...
+    private int size;
+
+    public Node<T> getFirst() {
+        return first;
+    }
+
+    public Node<T> getLast() {
+        return last;
+    }
+
```
AlexRozv, line 17: w don't need these getters because we don't want to give access to our nodes

AlexRozv (changes requested):

AlexRozv (approved):
well done ✅

PR #458, opened 2022-02-01
Realised LinkedList

DenTima:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,168 @@
...
+        }
+    }
+
+    public void checkElementIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Index is invalid");
+        }
+    }
+
+    void linkLast(T t) {
```
Drapikovsky, line 34: Change local variable names. They must be more informative. Replace 't' to "value" here and another place. 
```suggestion
    void linkLast(T value) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,168 @@
...
+        Node<T> l = last;
+        Node<T> newNode = new Node<>(l, t, null);
+        last = newNode;
+        if (l == null) {
+            first = newNode;
+        } else {
+            l.next = newNode;
```
Drapikovsky, line 41: make variable name l more informative. for example
```suggestion
            temporaryLast.next = newNode;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,168 @@
...
+                }
+            }
+        } else {
+            for (Node<T> node = first; node != null; node = node.next) {
+                if (object.equals(node.item)) {
+                    unlink(node);
+                    return true;
+                }
+            }
+        }
```
AlexRozv, line 157: we dont need this "if-else" statement if we change your "if" statement inside the for loop like this:
if (object == node.item || (object != null && object.equals(node.item))) {
	unlink(node);
	return true;
}

AlexRozv (changes requested):

AlexRozv (approved):
well done 👌

PR #459, opened 2022-02-01
Custom LinkedList implemented

karma-o:
Not the best solution, but I spend too much time on this and had none of it, at all, to properly refactor it.

### DIFF
--- a/src/main/java/core/basesyntax/MyLinkedList.java
+++ b/src/main/java/core/basesyntax/MyLinkedList.java
@@ -3,45 +3,148 @@ package core.basesyntax;
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+
+    static class Node<E> {
+        private Node<E> prev;
+        private E value;
+        private Node<E> next;
+
+        public Node(Node<E> prev, E value, Node<E> next) {
+            this.prev = prev;
+            this.value = value;
+            this.next = next;
+        }
+    }
+
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    public MyLinkedList() {
+        this.size = 0;
+    }
+
+    private void checkIndex(int index) {
+        if (!(index >= 0 && index < size || index >= 0 && size == 0)) {
+            throw new IndexOutOfBoundsException(
+                    "Index must be: 0 > index > size for size = " + size
+            );
+        }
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,144 @@
...
     @Override
     public T remove(int index) {
-        return null;
+        Node<T> nodeAtIndex = getNodeAtIndex(index);
+        if (index == 0) {
+            head = head.next;
+        }
+        if (index > 0 && index < size - 1) {
+            nodeAtIndex.next.prev = nodeAtIndex.prev;
+            nodeAtIndex.prev.next = nodeAtIndex.next;
```
NazariiYatsiuk, line 114: You should create separate method /unlink/ instead

https://mate-academy.github.io/jv-program-common-mistakes/java-core/collections/linked-list.html#create-a-separate-method-unlinknode-node

NazariiYatsiuk (changes requested):

NazariiYatsiuk (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,148 @@
...
+    }
+
+    private Node<T> getNodeAtIndex(int index) {
+        checkIndex(index);
+        Node<T> currentNode = head;
+        int currentIndex = 0;
+        while (currentIndex < index) {
+            currentNode = currentNode.next;
+            currentIndex++;
+        }
```
AlexRozv, line 38: it is optional, but we could just use "for i" loop here

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,148 @@
...
+            Node<T> tempNext = new Node<>(tail, value, null);
+            tail.next = tempNext;
+            tail = tempNext;
+            size++;
+        }
+        if (isEmpty()) {
+            head = new Node<>(null, value, null);
+            tail = head;
+            size++;
+        }
```
AlexRozv, line 74: this one is also optional, but instead of the second "if" here it's better to use "else"

AlexRozv (approved):
left some comments, overall well done 👌

PR #463, opened 2022-02-02
 implement all the required methods

Viktoriia-Zinchenko:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,157 @@
...
+            targetNode.prev.next = node;
+            targetNode.prev = node;
+        }
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
+        for (T t : list) {
+            add(t);
```
DenTima, line 57: You can change local variable name. They must be more informative. Replace 't' to "value".
```suggestion
        for (T value : list) {
            add(value);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,157 @@
...
+            }
+        } else {
+            while (node != null) {
+                if (node.value.equals(object)) {
+                    unlink(node);
+                    return true;
+                }
+                node = node.next;
+            }
+        }
```
DenTima, line 102: You can avoid if-else construction using this code 
Node<T> node;
        for (node = first; node != null; node = node.next) {
            if (object != null && object.equals(node.data) || node.data == object) {
                unlink(node);
                return true;
            }
        }

It looks better
```suggestion
Node<T> node;
        for (node = first; node != null; node = node.next) {
            if (object != null && object.equals(node.data) || node.data == object) {
                unlink(node);
                return true;
            }
        }
```

DenTima (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
Viktoriia-Zinchenko: Agree

src/main/java/core/basesyntax/MyLinkedList.java
Viktoriia-Zinchenko: agree, thanks!)

DenTima (approved):
👍👍👍👍

PR #464, opened 2022-02-02
added solution

Navijava:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,149 @@
...
     public void add(T value, int index) {
+        if (index == size) {
+            add(value);
+            return;
+        } else if (index == 0) {
+            Node<T> nextNode = new Node<>(null, value, head);
+            head.prev = nextNode;
+            head = nextNode;
+        } else {
+            checkIndex(index);
```
pikador10, line 45: you can put checkIndex on the top of this method and check the index at once

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,149 @@
...
+            Node<T> targetNode = getNode(index);
+            Node<T> newnode = new Node<>(targetNode.prev, value, targetNode);
```
pikador10, line 47: better write newNode

pikador10 (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,149 @@
...
+            }
+        } else {
+            while (node != null) {
+                if (node.value.equals(object)) {
+                    unlink(node);
+                    return true;
+                }
+                node = node.next;
+            }
+        }
```
AlexRozv, line 124: we don't need this if-else construct and we can leave only one while loop if we change our if statement inside this while loop like this:

if (object == node.value || (object != null && object.equals(node.value)))

AlexRozv (changes requested):

AlexRozv (approved):
well done ✅

PR #477, opened 2022-03-16
HW-LinkedList

savchenkoarts:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,163 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
     @Override
     public void add(T value) {
+        final Node<T> t = tail;
```
KseniiaMakarova, line 12: be careful with naming
```suggestion
       Node<T> oldTail = tail;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,163 @@
...
+    }
+
+    private void checkElementIndex(int index) {
+        if (!(index >= 0 && index < size)) {
+            throw new IndexOutOfBoundsException("Index: " + index + ", Size: " + size);
+        }
+
+    }
+
+    private void checkPositionIndex(int index) {
```
KseniiaMakarova, line 148:
```suggestion
    private void checkElementIndexForAdd(int index) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,163 @@
...
-        return null;
+        checkElementIndex(index);
+        return find(index).item;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        checkElementIndex(index);
+        Node<T> x = find(index);
```
KseniiaMakarova, line 58: naming

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,163 @@
...
+        T oldVal = x.item;
```
KseniiaMakarova, line 59: naming

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,163 @@
...
+                }
+            }
+        } else {
+            for (Node<T> x = head; x != null; x = x.next) {
+                if (x.item.equals(object)) {
+                    unlink(x);
+                    return true;
+                }
+            }
+        }
```
KseniiaMakarova, line 87: can we use only 1 loop here? check previous tasks - how to check 2 objects for equality given that if they are 2 nulls we should also return true?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,163 @@
...
         return false;
     }
 
+    T unlink(Node<T> x) {
+        final T element = x.item;
+        final Node<T> next = x.next;
+        final Node<T> prev = x.prev;
```
KseniiaMakarova, line 94:
```suggestion
        Node<T> next = x.next;
        Node<T> prev = x.prev;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,163 @@
...
+                if (x.item.equals(object)) {
+                    unlink(x);
+                    return true;
+                }
+            }
+        }
         return false;
     }
 
+    T unlink(Node<T> x) {
```
KseniiaMakarova, line 91: access modifier + naming of parameter

KseniiaMakarova (changes requested):
overall good job!

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -70,48 +71,15 @@ public T remove(int index) {
...
-                    return true;
-                }
-            }
-        } else {
-            for (Node<T> x = head; x != null; x = x.next) {
-                if (x.item.equals(object)) {
-                    unlink(x);
-                    return true;
-                }
+        for (Node<T> x = head; x != null; x = x.next) {
```
KseniiaMakarova, line 74: `x` is bad naming

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -138,14 +106,38 @@ public boolean isEmpty() {
         return x;
     }
 
+    private T unlink(Node<T> x) {
```
KseniiaMakarova, line 109: `x` is bad naming

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,154 @@
...
-        return null;
+        checkElementIndex(index);
+        Node<T> oldNode = find(index);
+        return unlink(oldNode);
     }
 
     @Override
     public boolean remove(T object) {
+        for (Node<T> node = head; node != null; node = node.next) {
+            if (object == node.item || object != null && node.item.equals(object)) {
```
KseniiaMakarova, line 74:
```suggestion
            if (object == node.item || object != null && object.equals(node.item)) {
```
before calling `equals()` on object we should make sure that it is not null

KseniiaMakarova (changes requested):
1 small fix left

KseniiaMakarova (approved):

PR #481, opened 2022-03-22
my solution is ready

DenysShl:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,150 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> nodeHead;
+    private Node<T> nodeTail;
```
MaliukDaria, line 7:
```suggestion
    private Node<T> head;
    private Node<T> tail;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,150 @@
...
+    private int size;
+
+    public MyLinkedList() {
+        this.nodeHead = this.nodeTail = new Node(null, null, null);
+        this.size = 0;
+    }
```
MaliukDaria, line 13: we dont need to do this, we should not have a single node from the start

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,150 @@
...
+            this.next = next;
+        }
+    }
+
     @Override
     public void add(T value) {
+        Node<T> currentNode = nodeHead;
+        for (int i = 0; i < size; i++) {
+            currentNode = currentNode.next;
+        }
```
MaliukDaria, line 32: why do we need this if we have tail node?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,150 @@
...
+        Node<T> newNode = new Node<T>(value, currentNode, null);
+        if (nodeHead == null) {
+            nodeHead = newNode;
+            nodeTail = newNode;
```
MaliukDaria, line 36: we could not have `nodeHead == null`  if we initialize it by `new Node(null, null, null);` from scratch. If we add first node, why we dont increment size variable?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,150 @@
...
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("No such index " + index);
+        }
```
MaliukDaria, line 73: we could move this check in separate private method as it repeated several times in code

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,150 @@
...
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("No such index " + index);
+        }
+        Node<T> currentNode = nodeHead;
+        for (int i = 0; i < index; i++) {
+            currentNode = currentNode.next;
+        }
```
MaliukDaria, line 53: lets move logic where we get node by index in separate private method as we use it several times in code

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,150 @@
...
+            nodeHead = currentNode.next;
+            size--;
+        } else if (currentNode.next == null) {
+            nodeTail = currentNode.prev;
+            size--;
+        } else {
+            currentNode.next.prev = currentNode.prev;
+            currentNode.prev.next = currentNode.next;
+            size--;
+        }
```
MaliukDaria, line 116: lets move logic of unlinking node to separate method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,150 @@
...
     @Override
     public int size() {
-        return 0;
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0 ? true : false;
```
MaliukDaria, line 150:
```suggestion
        return size == 0;
```

src/main/java/core/basesyntax/MyLinkedListApp.java
```diff
@@ -0,0 +1,19 @@
+package core.basesyntax;
+
+public class MyLinkedListApp {
```
MaliukDaria, line 3: lets remove this class from PR

MaliukDaria (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,145 @@
...
+        }
+        if (currentNode.prev == null) {
+            head = currentNode.next;
+        } else if (currentNode.next == null) {
+            tail = currentNode.prev;
+        } else {
+            currentNode.prev.next = currentNode.next;
+            currentNode.next.prev = currentNode.prev;
+        }
+        size--;
```
MaliukDaria, line 137:
```suggestion
        if (currentNode.prev == null) {
            head = currentNode.next;
        } else {
            currentNode.prev.next = currentNode.next;
        }
        if (currentNode.next == null) {
            tail = currentNode.prev;
        } else {
            currentNode.next.prev = currentNode.prev;
        }
        size--;
```

we could simplify this

MaliukDaria (changes requested):

PR #482, opened 2022-03-25
Custom linked list are implemented. All goals completed

JadeRowan:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,196 @@
...
     @Override
     public T get(int index) {
+        checkIndex(index);
+        Node<T> foundedNode = findNode(index);
+        return foundedNode == null ? null : foundedNode.getValue();
+    }
+
+    private Node<T> findNode(T content) {
+        Node<T> iterationNode = head;
+        while (iterationNode != null) {
```
KseniiaMakarova, line 75: same

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,196 @@
...
+        return neededNode;
+    }
+
+    private Node<T> searchFromTail(int index) {
+        Node<T> neededNode = tail;
+        int whileIndex = size - 1;
+        while (index < whileIndex) {
+            neededNode = neededNode.getPrevious();
+            whileIndex--;
+        }
```
KseniiaMakarova, line 103: let's write `for` loop as you did in `searchFromHead(int index)`

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
JadeRowan: Yes head can be null. Also if i reach tail.next which is null - loop safily ends

PR #483, opened 2022-03-26
feat: add my own LinkedList implementation

mikenikolov:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,151 @@
...
+            tail = head;
+            size++;
+            return;
+        }
+
+        Node<T> newNode = new Node<>(tail, value, null);
+        if (tail.next != null) {
+            tail.prev = tail;
+        }
+        tail.next = newNode;
```
Nick97-git, line 24: think about how to simplify this logic. Can you rewrite so you don't need to use return; in first if and write size++ twice in this method?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,151 @@
...
+        }
+        Node<T> newNode = new Node<>(null, value, null);
+        Node<T> requestedNode = getNode(index);
+        link(requestedNode, newNode);
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
+        if (list != null) {
```
Nick97-git, line 43: check for null is redundant

Nick97-git (changes requested):

Nick97-git (approved):

PR #485, opened 2022-03-29
LinkedList HW Skarzhynets

dSkarzh:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    private static class Node<T> {
```
KseniiaMakarova, line 10: let's move this inner class to the end of file

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
...
+
     @Override
     public void add(T value) {
+        if (size == 0) {
+            createFirstNode(new Node<>(value));
+            size++;
+        } else {
+            linkAsLast(new Node<>(value));
+            size++;
+        }
```
KseniiaMakarova, line 34:
```suggestion
        if (size == 0) {
            createFirstNode(new Node<>(value));
        } else {
            linkAsLast(new Node<>(value));
        }
        size++;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
...
     }
 
     @Override
-    public void add(T value, int index) {
+    public void add(T value, int index) throws IndexOutOfBoundsException {
+        if (index == 0 && size == 0) {
+            createFirstNode(new Node<>(value));
```
KseniiaMakarova, line 40: can we move `new Node<>(value)` into local variable and reuse it?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
...
+            for (int i = size - 1; i > index; i--) {
+                currentNode = currentNode.prev;
+            }
+            return currentNode;
+        }
+    }
+
+    private void validateIndex(int index) throws IndexOutOfBoundsException {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Index " + index
```
KseniiaMakarova, line 159: is it RuntimeException? if yes, do we need to write `throws` in method signature?

KseniiaMakarova (changes requested):

PR #489, opened 2022-04-03
added-solution

GerasimovTA-92:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,153 @@
...
+            this.prev = prev;
+            this.value = value;
+            this.next = next;
+        }
+    }
+
     @Override
     public void add(T value) {
+        Node<T> current = tail;
+        Node<T> newNode = new Node<>(current, value,null);
```
Sofasmile, line 26:
```suggestion
        Node<T> newNode = new Node<>(current, value, null);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,153 @@
...
+        tail = newNode;
+        if (current == null) {
+            head = newNode;
+        } else {
+            current.next = newNode;
+        }
+        size = size + 1;
```
Sofasmile, line 33:
```suggestion
        size++;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,153 @@
...
+        Node<T> current = findNodeByIndex(index);
+        removeNode(current);
+        return current.value;
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> current = head;
+        while (current.next != null) {
+            if ((current.value == object) || (current.value != null && current.value.equals(object))) {
```
dSkarzh, line 113:
```suggestion
            if (current.value == object || current.value != null && current.value.equals(object)) {
```
redundant parentheses, it will work without them because && is evaluated before ||

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,153 @@
...
+                break;
+            }
+            current = current.next;
+        }
+        if (Objects.equals(current.value, object)) {
```
dSkarzh, line 118: don't use class Objects, in common mistakes it's explicitly said that you should not use that
instead replace it with equivalent manual check (as you did earlier)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,153 @@
...
 
 import java.util.List;
+import java.util.Objects;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    private static class Node<T> {
```
dSkarzh, line 11: let's move this private class to the end of MyLinkedList class

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,153 @@
 package core.basesyntax;
 
 import java.util.List;
+import java.util.Objects;
```
dSkarzh, line 4: see comment below and common mistakes page

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,153 @@
...
+            current = current.next;
+        }
+        if (Objects.equals(current.value, object)) {
+            removeNode(current);
+            return true;
+        }
         return false;
     }
 
+    private void removeNode(Node<T> current) {
```
dSkarzh, line 125:
```suggestion
    private void unlink(Node<T> current) {
```
removeNode when it is called from remove method is really not informative name

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,153 @@
...
     public boolean remove(T object) {
+        Node<T> current = head;
+        while (current.next != null) {
+            if ((current.value == object) || (current.value != null && current.value.equals(object))) {
+                break;
+            }
+            current = current.next;
+        }
+        if (Objects.equals(current.value, object)) {
+            removeNode(current);
```
dSkarzh, line 119:
```suggestion
            unlink(current);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,153 @@
...
+            if (prev == null) {
+                head = newNode;
+            } else {
+                prev.next = newNode;
+            }
+            size++;
+        }
+    }
+
+    private Node<T> findNodeByIndex(int index) {
```
dSkarzh, line 57: let's move this method below all public methods

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,153 @@
...
+        return findNodeByIndex(index).value;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        if (index >= size || index < 0) {
+            throw new IndexOutOfBoundsException("Invalid index");
+        }
+        Node<T> x = findNodeByIndex(index);
```
dSkarzh, line 93:
```suggestion
        Node<T> nodeWithValueToReplace = findNodeByIndex(index);
```
x is not an informative name, think of something else
better long name, but understandable, than short but confusing

dSkarzh (commented):
Core logic looks good, left comments on minor problems: things like naming etc

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,150 @@
...
+        while (current.next != null) {
+            if (current.value == object || current.value != null && current.value.equals(object)) {
+                break;
+            }
+            current = current.next;
+        }
+        if (current.value == object || current.value != null && current.value.equals(object)) {
+            unlink(current);
+            return true;
+        }
```
Sofasmile, line 92:
```suggestion
        while (current.next != null) {
            if (current.value == object || current.value != null && current.value.equals(object)) {
                unlink(current);
                return true;
            }
            current = current.next;
        }
```

Sofasmile (changes requested):
great job)

Sofasmile (approved):

PR #499, opened 2022-04-08
created myLinkedList

DLyudmila:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,185 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size;
+    private Node<T>[] linkedList;
+    private int defaultCapacity = 1;
```
MaliukDaria, line 8: Whe should not use arrays here

MaliukDaria (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,191 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size;
+    private Node<T> head;
+    private Node<T> tail;
+
+    private static class Node<T> {
```
Nick97-git, line 10: let's move this class to the end of this one

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,191 @@
...
     public void add(T value) {
+        if (size == 0) {
+            head = (Node<T>) new Node<>(null, value, null);
+            tail = head;
+        }
+        if (size > 0) {
+            Node<T> tempNode = tail;
+            tail = (Node<T>) new Node<>(tempNode, value, null);
+            tempNode.next = tail;
+        }
```
Nick97-git, line 32: you can rewrite your logic so you won't need two ifs

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,191 @@
...
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
+        for (int i = 0; i < list.size(); i++) {
+            Node<T> tempNode = tail;
+            tail = (Node<T>) new Node<>(tempNode, list.get(i), null);
+            tempNode.next = tail;
+            size++;
```
Nick97-git, line 86: you can use add method inside of for loop here instead of all this logic

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,191 @@
...
+        prevNode.next = tempNode;
+        tempNode.next = nextNode;
+        nextNode.prev = tempNode;
+
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
+        for (int i = 0; i < list.size(); i++) {
```
Nick97-git, line 82: let's use foreach

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,191 @@
...
     @Override
     public T get(int index) {
-        return null;
+        if (size > index && index >= 0) {
+            Node<T> tempNode = head;
+            for (int i = 0; i < index; i++) {
+                tempNode = tempNode.next;
+            }
+            return tempNode.item;
+        } else {
```
Nick97-git, line 98: let's make check of index in the beginning of the method, the same in other similar cases

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,191 @@
...
+            throw new IndexOutOfBoundsException("Invalid Index");
+        }
     }
 
     @Override
     public T set(T value, int index) {
```
Nick97-git, line 104: too complex logic, this is how this method logic should look like:
1. Find node by index.
2. Create a variable that will store the current value of found node.
3. Update the value of the current node with a new one.
4. Return old value of a node

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,191 @@
...
+                head = nodeToDelete.next;
+            } else if (index == size - 1) {
+                nodeToDelete.prev.next = null;
+                tail = nodeToDelete.prev;
+            } else {
+                nodeToDelete.prev.next = nodeToDelete.next;
+                nodeToDelete.next.prev = nodeToDelete.prev;
+            }
+            size--;
+            return nodeToDelete.item;
```
Nick97-git, line 148: too complex logic, let's first find a node that will be removed, after that I guess you could see the ways of simplifying of your logic

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,191 @@
...
+        if (size > 0) {
+            Node<T> tempNode = tail;
+            tail = (Node<T>) new Node<>(tempNode, value, null);
+            tempNode.next = tail;
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
```
Nick97-git, line 37: too complex logic. There are two cases in here, after check for IndexOutOfBoundsException:
1. If index == 0 then you should use add method
2. If an index is valid then you should find the current node by this index and change on the passed one, shifting the current one to the right

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,191 @@
...
+            return nodeToDelete.item;
+        } else {
+            throw new IndexOutOfBoundsException("Invalid Index");
+        }
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> nodeToDelete = head;
+        for (int i = 0; i < size; i++) {
```
Nick97-git, line 157: the same as previous one

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,191 @@
...
-        return 0;
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
+        if (size == 0) {
+            return true;
+        }
         return false;
```
Nick97-git, line 191:
```suggestion
        return size == 0;
```

Nick97-git (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
DLyudmila: I have done it but add method doesn't work when index == 0 but size is not empty. I need to find another solution

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,156 @@
...
+    private int size;
+    private Node<T> head;
+    private Node<T> tail;
+
     @Override
     public void add(T value) {
+        Node<T> node = new Node<>(tail, value, null);
+        if (isEmpty()) {
+            head = node;
+            tail = node;
```
Nick97-git, line 15: you can put this line after else and remove duplicate below

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,156 @@
...
+            add(element);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index >= size || index < 0) {
+            throw new IndexOutOfBoundsException("Invalid Index");
+        } else {
```
Nick97-git, line 67:
```suggestion
        } 
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,156 @@
...
+            return tempNode.item;
+        }
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        if (index >= size || index < 0) {
+            throw new IndexOutOfBoundsException("Invalid Index");
+        } else {
```
Nick97-git, line 80:
```suggestion
        }
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,156 @@
...
+            return returnValue;
+        }
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (index >= size || index < 0) {
+            throw new IndexOutOfBoundsException("Invalid Index");
+        } else {
```
Nick97-git, line 95:
```suggestion
        } 
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,156 @@
...
+            }
+            return tempNode.item;
+        }
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        if (index >= size || index < 0) {
+            throw new IndexOutOfBoundsException("Invalid Index");
```
Nick97-git, line 79: let's create checkIndex of this and use it here and in the method below

Nick97-git (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,156 @@
...
+            tempNode = tempNode.next;
+        }
+        return tempNode.item;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        if (checkIndex(index) == false) {
+            throw new IndexOutOfBoundsException("Invalid Index");
```
Nick97-git, line 77: you should throw an exception in checkIndex method too, here in this method you just call checkIndex method, it should be void not boolean

Nick97-git (changes requested):

Nick97-git (approved):

PR #500, opened 2022-04-08
Override all method

dmytro-koliada:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,163 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void checkSize(int index) {
```
NesstorFedak, line 129: in this method you are checking index, rename please

NesstorFedak (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size;
+    private Node<T> head = new Node<>(null, null, null);
+    private Node<T> tail = head;
+    private Node<T> previous = tail;
```
KseniiaMakarova, line 9: let's remove `previous` field
and make `head` and `tail` null by default

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
...
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkIndex(index);
+        Node<T> newNode = tail;
+        for (int i = 0; i < index; i++) {
```
KseniiaMakarova, line 62: please move logic of finding element by index to a private method
you can reuse it in many public methods

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
...
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> newNode = tail;
+        for (int i = 0; i < size; i++) {
+            if (newNode.item == object || newNode.item != null && newNode.item.equals(object)) {
+                unlink(newNode);
+                return true;
+            } else if (newNode.next != null) {
```
KseniiaMakarova, line 110: do we need else after return?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
...
+        if (index >= size || index < 0) {
+            throw new IndexOutOfBoundsException(
+                    "Wrong index, " + size);
+        }
+    }
+
+    private T unlink(Node<T> newNode) {
+        T result = newNode.item;
+        if (newNode.prev == head) {
+            tail = newNode.next;
```
KseniiaMakarova, line 137: I can't understand this logic(

let's refactor code please
head and tail should be null by default (not empty nodes)
head is the first element, tail is the last element

head and tail should not be linked directly (nodes are NOT forming any circle)

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,141 @@
...
+            add(value);
+        } else if (index == 0) {
+            Node<T> newNode = new Node<>(null, value, tail);
+            newNode.next = head;
+            head.prev = newNode;
+            head = newNode;
+            size++;
+        } else {
+            Node<T> prevNode = findElementByIndex(index).prev;
+            Node<T> nextNode = findElementByIndex(index);
```
KseniiaMakarova, line 35:
```suggestion
            Node<T> nextNode = findElementByIndex(index);
            Node<T> prevNode = nextNode.prev;
```
don't call same method 2 times

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #503, opened 2022-04-10
Implement `MyLinkedList` class` methods

lexzane:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    private static class Node<T> {
```
dSkarzh, line 10: let's move this inner class to the bottom of MyLinkedList class
general structure is as follows: fields, constructors, getters&setters, public methods, private helper methods, inner classes

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
+            for (int i = size - 1; i > index; i--) {
+                indexedNode = indexedNode.prev;
+            }
+        }
+        return indexedNode;
+    }
+
     @Override
     public void add(T value) {
+        Node<T> checkTail = tail;
```
dSkarzh, line 40:
```suggestion
        Node<T> oldTail = tail;
```
just better naming

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
+        private Node<T> next;
+
+        public Node(Node<T> prev, T element, Node<T> next) {
+            this.prev = prev;
+            this.item = element;
+            this.next = next;
+        }
+    }
+
+    private Node<T> getIndexedNode(int index) {
```
dSkarzh, line 22:
```suggestion
    private Node<T> getNodeByIndex(int index) {
```
same thing about moving to the bottom, and better naming

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    private static class Node<T> {
+        private T item;
+        private Node<T> prev;
+        private Node<T> next;
+
+        public Node(Node<T> prev, T element, Node<T> next) {
```
dSkarzh, line 15:
```suggestion
        private Node(Node<T> prev, T element, Node<T> next) {
```
no need for this constructor to be public

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
+                }
+            }
+        } else {
+            for (Node<T> node = head; node != null; node = node.next) {
+                if (object.equals(node.item)) {
+                    unlink(node);
+                    return true;
+                }
+            }
+        }
```
dSkarzh, line 123: you can refactor and get rid of this ugly if else construction only by writing proper checks
```suggestion
        for (Node<T> node = head; node != null; node = node.next) {
            if ( /* new check here */ ) {
                unlink(node);
                return true;
            }
        }
        return false;
```

dSkarzh: if you feel stuck: check equals hashcode lessons once more, or just text me

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
+            tail = prev;
+        } else {
+            next.prev = prev;
+            node.next = null;
+        }
+
+        node.item = null;
+        size--;
+        return item;
+    }
```
dSkarzh, line 149:
```suggestion
    private void unlink(Node<T> node) {
        Node<T> prev = node.prev;
        Node<T> next = node.next;
        if (prev == null) {
            head = next;
        } else {
            prev.next = next;
        }
        if (next == null) {
            tail = prev;
        } else {
            next.prev = prev;
        }
        size--;
    }
```
please remove redundant empty lines, also there is no need to remove links, that are pointing FROM the node we want to remove, it's enough to remove all links that point TO the node we want to remove
if some object (in our case Node) doesn't have a reference it will be automatically removed by garbage collector
why do you return item from this methods? maybe it's better to make this method void?

dSkarzh (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,145 @@
...
+        }
+        throw new IndexOutOfBoundsException("Can't set value. Pass correct input index");
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (index >= 0 && index < size) {
+            T removeNode = getNodeByIndex(index).item;
+            unlink(getNodeByIndex(index));
```
KseniiaMakarova, line 75: use local variable - do not call same method 2 times

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,145 @@
...
+            targetNode.item = value;
+            return oldValue;
+        }
+        throw new IndexOutOfBoundsException("Can't set value. Pass correct input index");
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        Node<T> removeNode = getNodeByIndex(index);
```
KseniiaMakarova, line 73: а если тут индекс невалидный?

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #506, opened 2022-04-10
Added homework solution

ZeroNeiron:


DimaPilipenko (approved):
it seems to be ok)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,143 @@
...
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Index " + index + " out of bounds");
+        }
+        if (index == size) {
+            add(value);
+            return;
+        }
+        Node<T> node = getNodeByIndex(index);
+        Node<T> prevNode = node == null ? null : node.prev;
```
KseniiaMakarova, line 32: can it be null?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,143 @@
...
+        return node.value;
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> node = head;
+        int index = 0;
+        while (node != null) {
+            if (node.value == object || node.value.equals(object)) {
+                remove(index);
```
KseniiaMakarova, line 80: use unlink() instead

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,143 @@
...
+        unlink(node);
+        return node.value;
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> node = head;
+        int index = 0;
+        while (node != null) {
+            if (node.value == object || node.value.equals(object)) {
```
KseniiaMakarova, line 79: what if node.value == null && object != null?

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,143 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private Node<T> getNodeByIndex(int index) {
+        Node<T> node = head;
```
KseniiaMakarova, line 99: why are we always moving from head?

KseniiaMakarova (changes requested):

PR #527, opened 2022-05-01
the solution was implemented

Dima-Davydenko:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,166 @@
...
+            head = newValue;
+        } else if (index == size - 1) {
+            Node<T> newValue = new Node<>(tail.prev, value, tail.next);
+            tail.prev.next = newValue;
+            tail = newValue;
+        } else {
+            Node<T> newValue = new Node<>(oldNode.prev, value, oldNode.next);
+            oldNode.next.prev = newValue;
+            oldNode.prev.next = newValue;
+        }
```
MaliukDaria, line 82: Do we really need all that? Could we just set new value to Node<T> oldNode?

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #534, opened 2022-05-11
tests and stylechecks passed, self-refactoring done

drapail:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,176 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size;
+    private Node<T> first;
+    private Node<T> last;
+
+    private static class Node<T> {
```
Nick97-git, line 10: private classes and methods should go after public methods

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,176 @@
...
+        }
+    }
+
+    private void checkIndexSet(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Invalid index");
+        }
+    }
+
+    private void checkIndexRemove(int index) {
```
Nick97-git, line 34: remove one of them and change the name to checkIndex

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,176 @@
...
+            this.next = next;
+            this.prev = prev;
+        }
+    }
+
+    private void checkIndexAdd(int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Invalid index");
+        }
+    }
```
Nick97-git, line 26: remove this method and use its logic directly in method where it's used

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,176 @@
...
+
+    private void checkIndexRemove(int index) {
+        if (index < 0 || index == size) {
+            throw new IndexOutOfBoundsException("Invalid index");
+        }
+    }
+
+    private Node iterateMyLinkedList(int index) {
```
Nick97-git, line 40:
```suggestion
    private Node getNodeByIndex(int index) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,176 @@
...
+        }
+        return null;
+    }
+
+    private void unlink(Node nodeToUnlink) {
+        if (nodeToUnlink == first && nodeToUnlink == last) {
+            first = null;
+            last = null;
+            return;
+        }
```
Nick97-git, line 66: remove this check

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,176 @@
...
+            first = nodeToUnlink.next;
+            return;
+        }
+        if (nodeToUnlink == last) {
+            nodeToUnlink.prev.next = null;
+            last = nodeToUnlink.prev;
+            return;
+        }
+        nodeToUnlink.prev.next = nodeToUnlink.next;
+        nodeToUnlink.next.prev = nodeToUnlink.prev;
```
Nick97-git, line 78:
```suggestion
        if (nodeToUnlink == first) {
            nodeToUnlink.next.prev = null;
            first = nodeToUnlink.next;
        } else if (nodeToUnlink == last) {
            nodeToUnlink.prev.next = null;
            last = nodeToUnlink.prev;
            return;
        } else {
        nodeToUnlink.prev.next = nodeToUnlink.next;
        nodeToUnlink.next.prev = nodeToUnlink.prev;
       }
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,176 @@
...
+    }
+
     @Override
     public void add(T value) {
+        if (size == 0) {
+            Node node = new Node<>(value, null, null);
+            first = node;
+            last = node;
+        }
+        if (size > 0) {
```
Nick97-git, line 88:
```suggestion
        } else {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,176 @@
...
+            Node node = new Node<>(value, null, null);
+            Node currentNode = last;
+            currentNode.next = node;
+            node.prev = currentNode;
+            last = node;
```
Nick97-git, line 93: think about hot to simplify this code and avoid duplication

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,176 @@
...
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        checkIndexAdd(index);
+        if (size == 0 || index == size) {
+            this.add(value);
```
Nick97-git, line 102:
```suggestion
            add(value);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,176 @@
...
+        Node currentNode = iterateMyLinkedList(index);
+        return (T) currentNode.item;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        checkIndexSet(index);
+        Node currentNode = iterateMyLinkedList(index);
+        Node returnNode = new Node(currentNode.item, currentNode.next, currentNode.prev);
```
Nick97-git, line 140:
```suggestion
        T oldValue = currentNode.item;
```

Nick97-git (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
drapail: I agree with the change, but have one question. I saw in one tutorial that it is preferred to use multiple if statements instead of if-elseif-else, because "else" case tends to collect various limit conditions (read "mistakes") and generally multiple if statements are easier to read. So, why is it better to use if-elseif-else?

src/main/java/core/basesyntax/MyLinkedList.java
drapail: I moved Node creation out from if statements directly to method, moved "last = node" to the end, as we always add to the end.
I can't think of other simplifications, it looks pretty simple to me already. If I am missing something obvious - do not hesitate to point that to me, that's interesting

src/main/java/core/basesyntax/MyLinkedList.java
Nick97-git: to be honest, both options are bad, but choosing between them it's better to use if else in such cases in my opinion. I don't have some specific reasons for using of this approach, just from my experience it's a more readable option than the previous one

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,162 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size;
+    private Node<T> first;
+    private Node<T> last;
+
+    private static class Node<T> {
```
Nick97-git, line 10: private classes and methods should go after public methods

src/main/java/core/basesyntax/MyLinkedList.java
drapail: Understood, thank you!
What do you mean "both options are bad"? How can you avoid if-else constructions at all and in this case in particular?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,162 @@
...
+
     @Override
     public void add(T value) {
+        Node node = new Node<>(value, null, null);
+        if (size == 0) {
+            first = node;
+        } else {
+            Node currentNode = last;
+            currentNode.next = node;
+            node.prev = currentNode;
```
Nick97-git, line 36: do you need this line?

Nick97-git (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
drapail: yes, I update links to the previous and next nodes when add a node to the end of the linked list. Without it a lot of tests fail

src/main/java/core/basesyntax/MyLinkedList.java
Nick97-git: I didn't mean, that you can avoid these options here. I meant that they look both not so good in terms of readability)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,163 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size;
+    private Node<T> first;
+    private Node<T> last;
+
+    private void checkIndex(int index) {
```
Nick97-git, line 10: private methods should go after public ones

Nick97-git (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -150,7 +144,14 @@ private void unlink(Node nodeToUnlink) {
...
     }
 
+    private void checkIndex(int index) {
+        if (index < 0 || index == size) {
+            throw new IndexOutOfBoundsException("Invalid index");
+        }
+    }
+
     private static class Node<T> {
+
```
Nick97-git, line 154: remove empty line

Nick97-git (approved):

PR #548, opened 2022-05-25
Hw jv linked list

zim10093:


liliia-ponomarenko (changes requested):
Good solution! Let’s improve it 😉

liliia-ponomarenko (approved):
good job 👍

PR #550, opened 2022-05-27
implemented MyLinkedListInterface and all the required methods

Zinchuk-Yevhenii-Mate:


liliia-ponomarenko (changes requested):
Good solution! Let’s improve it 😉

PR #555, opened 2022-05-31
MyLinkedList implemented

oleksii-kordysh:


nazarbaklanov (changes requested):
Добавил комменты

nazarbaklanov (approved):
Ok)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,181 @@
...
 
     @Override
     public boolean remove(T object) {
-        return false;
+        try {
+            unlink(findNode(object));
+            return true;
+        } catch (NoSuchElementException e) {
+            return false;
+        }
```
KostyaNikiforov, line 63: We don't have to catch unchecked exceptions. Let's throw checked or you can just move the logic from that method here. Anyway you don't use it anywhere else

oleksii-kordysh: moved logic from findNode by element in remove by element

KostyaNikiforov (approved):
🔥

PR #556, opened 2022-05-31
All tests were passed. CheckStyle corrected.

MKaramavrov:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,145 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Your index is not valid.");
```
Cosmic-M, line 98: it will be better if you create a constant with message for your exception

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,145 @@
...
+        Node<T> current = getNode(index);
+        T removedValue = current.value;
+        unlink(current);
+        return removedValue;
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> current = head;
+        for (int i = 0; i < size; i++) {
```
Cosmic-M, line 75: you may replace with: while (current != null) {
(it's just my proposal, I think it looks a bit better)

PR #557, opened 2022-05-31
hw-linked-array

TymurTsvietkov:


Vasyl-Bevz (changes requested):

Vasyl-Bevz (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,152 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> firstNode;
+    private Node<T> lastNode;
+    private int size;
+
+    public MyLinkedList() {
+        size = 0;
```
KostyaNikiforov, line 11: What is the default value of int?)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,152 @@
...
+            Node<T> newNode = new Node<>(lastNode, value, null);
+            lastNode.next = newNode;
+            lastNode = newNode;
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        validateIndex(index, 1);
```
KostyaNikiforov, line 29: The fewer parameters a method has, the easier it is to work with.
In this case, it is not very clear why the second int parameter is needed.
What if I pass 5 to it?
There are 3 options to make it clearer:
1. Create two constants.
2. Create an enum with two parameters.
3. Use boolean instead of int.
And also you can try to avoid using the second argument.
What if you move this method after your next if block?

KostyaNikiforov (changes requested):

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):

PR #559, opened 2022-05-31
methods were implemented

mykola8426:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,157 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
     @Override
     public void add(T value) {
```
olyaklipa, line 11: I think you can add "size++" only once in the end of the method (after all your if-else blocks).

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,157 @@
...
-        return null;
+        indexCheck(index);
+        return getNodeByIndex(index).item;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        Node<T> nodeWillReplaced = getNodeByIndex(index);
+        T a = nodeWillReplaced.item;
```
olyaklipa, line 69: "a" - could you please make the variable name more informative

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,157 @@
...
 
     @Override
     public T get(int index) {
-        return null;
+        indexCheck(index);
+        return getNodeByIndex(index).item;
     }
 
     @Override
     public T set(T value, int index) {
```
olyaklipa, line 67: Alternatively, you can use your "remove(index) and "add(value, index) " methods here

olyaklipa (commented):
Nice solution!

olyaklipa (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,155 @@
...
+
+        } else if (index < size) {
+            Node<T> nodeOnIndex = getNodeByIndex(index);
+            Node<T> newNode = new Node<>(nodeOnIndex, value, nodeOnIndex.prev);
+            nodeOnIndex.prev.next = newNode;
+            nodeOnIndex.prev = newNode;
+            size++;
+        } else {
+            add(value);
+        }
```
KostyaNikiforov, line 48: It seems like you can remove your first if block because if the size equals 0 or if the index equal size you do the same things

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,155 @@
...
+        } else {
+            actual = tail;
+            for (int j = size - 1; j > index; j--) {
+                actual = actual.prev;
+            }
+        }
+        return actual;
+    }
+
+    private void indexCheck(int index) {
```
KostyaNikiforov, line 134:
```suggestion
    private void checkIndex(int index) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,155 @@
...
     public void addAll(List<T> list) {
+        for (T element : list) {
+            add(element);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        indexCheck(index);
```
KostyaNikiforov, line 60: You can remove this line, because you already check the index in the `getNodeByIndex` method

KostyaNikiforov (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
mykola8426: Метод checkIndex() включає index >= size , тобто якщо я видалю і користувач захоче додати елемент на 0 позицію , то викине Exception

KostyaNikiforov (approved):

PR #561, opened 2022-06-01
created linkedlist

KarinaTanasiichuk:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,146 @@
...
+        return oldNodeValue;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        validIndex(index);
+        Node<T> oldNode = findNode(index);
+        T oldNodeItem = oldNode.item;
+        unlink(oldNode);
```
zim10093, line 81: Recommendation. You can return an item in the unlink() method. It will save several lines.
It may look like this, return unlink(findNode(index));
So, you can change it if you want,  the last word for the mentor.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -70,8 +70,8 @@ public T remove(int index) {
     public boolean remove(T object) {
         Node<T> node = first;
         for (int i = 0; i < size; i++) {
-            if (node.item == object
-                    || node.item.equals(object)) {
+            if (object == null ? node.item == null
+                    : node.item.equals(object)) {
```
zim10093, line 74: Imagine:
object   ==  3; 
node == null;
what will happend?
So, you should invoke the equals method in the nonNull method. Now item can be Null and you try to invoke equals in the null object.

zim10093 (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,143 @@
...
+        Node<T> node = first;
+        if (node != null) {
+            for (int i = 0; i < size; i++) {
+                if (node.item == object || node.item.equals(object)) {
+                    unlink(node);
+                    return true;
+                }
+                node = node.next;
+            }
+        }
```
zim10093, line 80: In this case, you can`t remove a first null element. You should check in the loop.
```suggestion
            for (int i = 0; i < size; i++) {
                if (object == null ? node.item == null : object.equals(node.item)) {
                    unlink(node);
                    return true;
                }
                node = node.next;
        }
```

zim10093 (changes requested):

zim10093 (approved):
At last. Please make commit messages more informative

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,141 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void validIndex(int index) {
```
KostyaNikiforov, line 92:
```suggestion
    private void validateIndex(int index) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,141 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size;
+    private Node<T> first;
+    private Node<T> last;
+
     @Override
     public void add(T value) {
+        Node<T> node = new Node<>(last, value,null);
```
KostyaNikiforov, line 12:
```suggestion
        Node<T> node = new Node<>(last, value, null);
```

KostyaNikiforov (approved):
Good job)

PR #562, opened 2022-06-01
LinkedList methods were implemented

YaroslavLa:


viktoriia-haraniuk (approved):
Good solution, I like it :wink:

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):

PR #565, opened 2022-06-01
created MyLinkedList class

DimaPodakin:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,140 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private Node<T> findNodeByIndex(int index) {
+        checkIndexException(index);
```
ChumakAndrey, line 91: There is redundant check, because you additional check index in each method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,140 @@
...
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("This index is invalid");
+        }
+    }
+
+    private T unlink(Node<T> removedNode) {
+        if (size == 1) {
+            head = null;
+            tail = null;
+        } else if (removedNode == head) {
```
ChumakAndrey, line 117: In this case it works, but I think it's not good practice to compare objects with `==`

ChumakAndrey (changes requested):

ChumakAndrey (approved):

KostyaNikiforov (approved):
LGTM:fire:

PR #586, opened 2022-07-13
Hw120722

sergeiburya:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,212 @@
...
+
+    public MyLinkedList() {
+
+    }
+
+    public MyLinkedList(int size, Node<T> head, Node<T> tail) {
+        this.size = size;
+        this.head = head;
+        this.tail = tail;
+    }
```
MaliukDaria, line 18: Why do we need these constructors?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,212 @@
...
     public T remove(int index) {
-        return null;
+        checkIndex(index, size - 1);
+        if (index == 0) {
+            return removeFirst();
+        } else if (index == size - 1) {
+            return removeLast();
+        } else {
+            Node<T> oldNode = findElement(index);
+            unlink(oldNode);
```
MaliukDaria, line 70: could we implement logic of unlinking node, considering that we could unlink first node or last node and just call unlink method here and in `boolean remove(T object)` method?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,212 @@
...
+    private void unlink(Node<T> node) {
+        node.next.prev = node.prev;
+        node.prev.next = node.next;
+        node.prev = null;
+        node.next = null;
+    }
+
+    private void checkIndex(int index, int limit) {
+        if (index < 0 || index > limit) {
+            throw new IndexOutOfBoundsException("Index is outside");
```
MaliukDaria, line 185:
```suggestion
            throw new IndexOutOfBoundsException("Index " + index + " is out or range for limit " + limit);
```

lets make exception messages more informative

MaliukDaria (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,167 @@
...
+                    return true;
+                }
+            }
+        } else {
+            for (Node<T> currentNode = head; currentNode != null; currentNode = currentNode.next) {
+                if (object.equals(currentNode.value)) {
+                    unlink(currentNode);
+                    return true;
+                }
+            }
```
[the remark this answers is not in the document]
sergeiburya: I think that it is possible, but we need to avoid NullPointerException when we call the object.equals in the empty object . I haven't found a solution yet.

MaliukDaria (approved):

PR #596, opened 2022-07-29
Solution created.

tumanUk:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,140 @@
...
+        } else {
+            current = tail;
+            for (int i = size - 1; i > index; i--) {
+                current = current.prev;
+            }
+        }
+        return current;
+    }
+
+    private void checkIndexException(int index) {
```
MaliukDaria, line 107:
```suggestion
    private void checkIndex(int index) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,140 @@
...
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("This index is invalid");
```
MaliukDaria, line 109:
```suggestion
            throw new IndexOutOfBoundsException("This index "+index+"is invalid for size " + size);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,140 @@
...
+    }
+
+    private T unlink(Node<T> removedNode) {
+        if (size == 1) {
+            head = null;
+            tail = null;
+        } else if (removedNode.equals(head)) {
+            head.next.prev = null;
+            head = removedNode.next;
+        } else if (removedNode.equals(tail)) {
```
MaliukDaria, line 120:
```suggestion
        } else if (removedNode.prev == null) {
            head.next.prev = null;
            head = removedNode.next;
        } else if (removedNode.next == null) {
```

I think it is better not to use equals here

MaliukDaria (changes requested):

PR #597, opened 2022-08-01
task completed v.1.0

ghost1975:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,176 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    public MyLinkedList() {
+    }
```
Nick97-git, line 11: do you need to add this constructor explicitly?

Nick97-git (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,176 @@
...
+        }
+        if (tail == null) {
+            tail = new Node<>(value, head, null);
+            head.next = tail;
+            size++;
+            return;
+        }
+        tail.next = new Node<>(value, tail, null);
+        tail = tail.next;
+        size++;
```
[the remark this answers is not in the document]
ghost1975: Changed the logic a little. In the case when there is only one Node head, the initial value "previous" field of the newly created Node tail = "null". It needs to be initialized. The same as the value of the "next" field for the "head".

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,174 @@
...
+        } else {
+            if (tail == null) {
+                tail = newNode;
+                tail.previous = head;
+                head.next = tail;
+            } else {
+                tail.next = newNode;
+                tail = newNode;
+            }
+        }
```
Nick97-git, line 24: code below doesn't work? If it works, I don't see any reason not to use it 
Node newNode = new Node<>(tail, value, null);
if (size == 0) {
head = newNode;
} else {
tail.next = newNode;
}
tail = newNode;
size++;

Nick97-git (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
ghost1975: with the proposed implementation, I receive the error "Cannot assign field "next" because "this.tail" is null."
That's why I added field initialization head.next and tail.previous.  
The corresponding code fragment is used only if there is only one head node.

Nick97-git (approved):

PR #612, opened 2022-08-07
added solution

Nesterenko-Andrey:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,144 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size = 0;
```
justpopova, line 6:
```suggestion
    private int size = 0;
```
default value is 0

justpopova, line 6:
```suggestion
    private int size;
```
by default is already 0

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,144 @@
...
+    private Node<T> head;
+    private Node<T> tail;
+
```
justpopova, line 9: also i think here we need to have constructor with our initialized fields head , tail

justpopova (changes requested):
:)

justpopova (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,144 @@
...
+            Node<T> node = new Node<>(current.prev, value, current);
+            current.prev.next = node;
+            current.prev = node;
+        }
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
+        for (T t : list) {
```
Sofasmile, line 44:
```suggestion
        for (T element : list) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,144 @@
...
     @Override
     public T get(int index) {
-        return null;
+        return getNode(index).value;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        checkIndex(index);
```
Sofasmile, line 56: should I call checkIndex() here? we have already called this method in getNode().

Sofasmile (changes requested):
great work

Sofasmile (approved):

PR #617, opened 2022-08-08
add solution

AndrewHudzenko:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,160 @@
...
 import java.util.List;
+import java.util.Objects;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    public class Node<T> {
+        private T element;
+        private Node<T> next;
+        private Node<T> prev;
+
+        public Node(Node<T> prev, T value, Node<T> next) {
```
max-bolotin, line 12: Is it obligatory to create the constructor with prev and next parameters? Probably only value might be enough

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,160 @@
...
+                tail = prev;
+            }
+        }
+        size--;
+        return removedElement;
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> current = head;
```
max-bolotin, line 102: In remove by object method it is usually enough to find the required object by simple iteration through nodes, and when found  - call the remove by index method with the obtained index. Such implementation will be much simplier

max-bolotin (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
AndrewHudzenko: in my case yes / i tried first time use only contructor with value , but i got exceptions like - "u have first element null...something" and i have to initialize in each case "head" = null

src/main/java/core/basesyntax/MyLinkedList.java
AndrewHudzenko: good point

src/main/java/core/basesyntax/MyLinkedList.java
max-bolotin: ok, got it

max-bolotin (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,147 @@
 package core.basesyntax;
 
 import java.util.List;
+import java.util.Objects;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    public class Node<T> {
```
Sofasmile, line 7: usually, inner class is created at the end of the file

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,147 @@
...
+            tail = tail.next;
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        //here we can't use method checkElementIndex(index)
+        // because we add +1 to size
+        //useless to create new method only for this case
```
Sofasmile, line 39: let's remove these comments

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,147 @@
...
+        checkElementIndex(index);
+        return getNodeByIndex(index).element;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        checkElementIndex(index);
+        Node<T> node = getNodeByIndex(index);
+        T oldVal = node.element;
```
Sofasmile, line 73:
```suggestion
        T oldValue = node.element;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,147 @@
...
+        } else {
+            Node<T> prev = getNodeByIndex(index - 1);
+            removedElement = prev.next.element;
+            prev.next = prev.next.next;
+            if (index == size - 1) {
+                tail = prev;
+            }
+        }
+        size--;
+        return removedElement;
```
Sofasmile, line 97: let's refactor this method in the following way:
- check index;
- get node by index
- call another method, like unlink() or removeNode() where this node will be removed from the list

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,147 @@
...
     }
 
     @Override
     public boolean remove(T object) {
+        int index = 0;
+        Node<T> current = head;
+        while (current != null) {
+            if (object == current.element || object != null && object.equals(current.element)) {
+                remove(index);
+                return true;
```
Sofasmile, line 107: here you can call method unlink()/removeNode() too

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,147 @@
...
     public boolean isEmpty() {
-        return false;
+        return head == null;
+    }
+
+    private Node<T> getNodeByIndex(int index) {
+        Node<T> current = head;
+        for (int i = 0; i < index; i++) {
+            current = current.next;
+        }
```
Sofasmile, line 129: we can improve performance here
if the index is in the first part of the list, you can start the movement from the beginning, otherwise you can start the movement from the end

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,147 @@
...
+        return current;
+    }
+
+    private String outOfBoundsMsg(int index) {
+        return "Index: " + index + ", Size: " + size;
+    }
```
Sofasmile, line 135: better move logic from this method to the checkElementIndex()

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,147 @@
...
+
+    private boolean isElementIndex(int index) {
```
Sofasmile, line 137:
```suggestion
    private boolean isIndexCorrect(int index) {
```

Sofasmile (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,156 @@
...
     @Override
     public T remove(int index) {
-        return null;
+        checkElementIndex(index);
+        return removeNode(getNodeByIndex(index));
     }
 
     @Override
     public boolean remove(T object) {
+        int index = 0;
```
Sofasmile, line 71: do we need index variable?

Sofasmile (changes requested):

Sofasmile (approved):

PR #621, opened 2022-08-09
implemented my own LinkedList and checked common mistakes

GolDenik:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,145 @@
...
+    }
+
+    private Node<T> findNodeByIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Index " + index + " is invalid");
+        }
+        Node<T> node = head;
+        for (int i = 1; i <= index; i++) {
+            node = node.next;
+        }
```
JustElvis, line 129: i think you need find index from head to half and from tail to half. its faster.

JustElvis (changes requested):

JustElvis (approved):
great job

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,153 @@
...
+            this.next = next;
+        }
+    }
+
+    private void checkIndexForAddByIndex(int index) {
+        if (index >= 0 && index <= size) {
+            return;
+        }
+        throw new IndexOutOfBoundsException("Cannot add element , because index "
+                + index + " is invalid");
```
Sofasmile, line 104:
```suggestion
        if (index < 0 || index > size) {
            throw new IndexOutOfBoundsException("Cannot add element , because index "
                + index + " is invalid");
        }
```

Sofasmile (approved):
🔥

PR #630, opened 2022-08-10
Home work

fliarkovskyi:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,159 @@
...
+            node.prev.next = null;
+        }
+        if (node.prev == null && node.next != null) {
+            head = node.next;
+            node.next.prev = null;
+        }
+        if (node.prev != null && node.next != null) {
+            node.prev.next = node.next;
+            node.next.prev = node.prev;
+        }
```
victoriadobryden, line 138: let's make it through operator case of for one of conditions: node.prev == null or node.next == null.

victoriadobryden (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,156 @@
...
     }
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException(
+                    "LinkedList index: " + index + " out of bounds exception");
+        }
+        if (index == size) {
+            add(value);
```
Noct2000, line 29:
```suggestion
        if (index == size) {
            add(value); 
            return;
       }
       checkIndex(index)
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,156 @@
...
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void addFirst(Node<T> node) {
+        head = node;
+        tail = node;
+        size++;
+    }
```
Noct2000, line 103: I think you can use this logic directly in the `public void add(T value)`

Noct2000 (changes requested):

Noct2000 (approved):

PR #639, opened 2022-08-10
Was created own LinkedList, was implemented all the required methods

valentyn-slutskyi:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,174 @@
...
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Index " + " is not exist");
+        }
+    }
+
+    private void checkIndexBigger(int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Index " + " is not exist");
+        }
+    }
```
SofiiaKazanivska, line 127: you can combine these two checks, maybe that would be better

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,174 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    static class Node<T> {
```
SofiiaKazanivska, line 105: I think it would be better if this class was at the beginning, because it is a static class

SofiiaKazanivska (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,174 @@
...
+
+    private Node<T> findNodeByValue(T object) {
+        Node<T> deletedNode = first;
+        while (deletedNode != null) {
+            if (deletedNode.value == object
+                    || deletedNode != null && deletedNode.value.equals(object)) {
+                return deletedNode;
+            }
+            deletedNode = deletedNode.next;
+        }
```
Noct2000, line 152: check if index < size / 2 then move from first to last, else move from last to first

Noct2000 (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
valentyn-slutskyi: There is no any index. This method is looking for Node by value. But in another method I use looking for from both side, see findNodeByIndex() method.

Noct2000 (approved):

PR #649, opened 2022-08-11
Home work

oleh-satoshi:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,149 @@
...
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    public MyLinkedList() {
+        tail = new Node<>(null, null, null);
+        head = new Node<>(null, null, null);
+    }
```
justpopova, line 13:
```suggestion

```
by default tail and head default, and you can remove all this constructor, cause it's empty and we don't need to initialize here anything

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,149 @@
...
+            tail.next = node;
+            node.prev = tail;
+            tail = node;
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        checkIndex(index);
```
justpopova, line 30:
```suggestion
        checkIndex(index);
```
i think maybe remove this method at all and write checking condition here, because you use this method just one time in your code. i dont think if its necessary to have it

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,149 @@
...
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Wrong index" + index);
+        }
+    }
```
justpopova, line 108:
```suggestion
       
```
maybe better remove it and move all logic to add(T value, int index) {
            throw new IndexOutOfBoundsException("Wrong index" + index);
        }
        
     
}

justpopova (changes requested):
good job!

justpopova (approved):
🔥

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,141 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void checkEqualsIndex(int index) {
```
Sofasmile, line 102:
```suggestion
    private void checkIndex(int index) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,141 @@
...
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Wrong index" + index);
+        }
+    }
+
+    private Node<T> getNodeByIndex(int index) {
+        Node<T> current = head;
+        for (int i = 0; i < index; i++) {
```
Sofasmile, line 110: if the index is in the first part of the list, you should start movement from the head, otherwise from the tail
it will improve performance

Sofasmile (changes requested):

Sofasmile (approved):

PR #651, opened 2022-08-11
Linked list was created

KosatskiyKirill:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,149 @@
...
+                requiredNode = requiredNode.next;
+            }
+        } else {
+            requiredNode = tail;
+            for (int i = size - 1; i > index; i--) {
+                requiredNode = requiredNode.prev;
+            }
+        }
+        return requiredNode;
+    }
```
Rossoline, line 129: You can simplify it to one loop.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,149 @@
...
+            }
+        } else if (requiredNode == tail) {
+            tail = tail.prev;
+            tail.next = null;
+            size--;
+        } else {
+            requiredNode.next.prev = requiredNode.prev;
+            requiredNode.prev.next = requiredNode.next;
+            size--;
+        }
```
Rossoline, line 149: make only one size -- in the end of method

Rossoline (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
Rossoline: all method `searchNodeByIndex`

KosatskiyKirill: Yes, but then i must always started of head and i don't wont it

Rossoline (approved):

Sofasmile (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -26,9 +26,10 @@ public void add(T value, int index) {
         }
         if (index == size) {
             add(value);
+            return;
         } else {
```
Sofasmile, line 30: we can remove else block because previous block contains `return`

Sofasmile (changes requested):

PR #665, opened 2022-08-31
First solution of LinkedList HM

slvkobzar:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,146 @@
 package core.basesyntax;
 
 import java.util.List;
+import java.util.Objects;
```
aaananas, line 4: image

aaananas (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,144 @@
...
 
     @Override
     public boolean remove(T object) {
+        for (int i = 0; i < size; i++) {
+            if (object == findNode(i).element || object != null
+                    && object.equals(findNode(i).element)) {
+                remove(i);
+                return true;
+            }
+        }
```
aaananas, line 80:
```suggestion
        for (Node<T> node = head; head != null; node = node.next) {
            if (node.value == value || value != null && value.equals(node.value)) {
               unlink(node);
               return true;
            }
        }
        return false;
```
using findNode(i) like this is not a good idea, because you invoke this method and iterate almost over the whole list 2 times for each current iteration
there is no need for this, because you can iterate over the actual nodes, and pass the node with the value that is to be removed to a unlink method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,144 @@
...
+            return;
+        }
+        Node currentNode = findNode(index);
+        if (index == 0) {
+            Node newNode = new Node(null, value, currentNode);
+            currentNode.prev = newNode;
+            head = newNode;
+        } else {
+            Node newNode = new Node(currentNode.prev, value, currentNode);
+            currentNode.prev.next = newNode;
```
aaananas, line 39:
```suggestion
        if (index == 0) {
            Node newNode = new Node(null, value, head);
            head.prev = newNode;
            head = newNode;
        } else {
            Node currentNode = findNode(index);
            Node newNode = new Node(currentNode.prev, value, currentNode);
            currentNode.prev.next = newNode;
```
no need to look for a node if you know that its head

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,144 @@
...
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Index out of range");
+        }
+        if (index == size) {
+            add(value);
+            return;
+        }
```
aaananas, line 31:
```suggestion
        if (index == size) {
            add(value);
            return;
        }
        checkIndex(index);
```

aaananas: it will work perfectly

aaananas (changes requested):

aaananas (approved):

PR #667, opened 2022-09-04
added solution

roman680:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,138 @@
...
+        } else {
+            current = tail;
+            for (int i = size - 1; i > index; i--) {
+                current = current.prev;
+            }
+        }
+        return current;
+    }
+
+    private void removeNode(Node<T> current) {
```
zachar-chorny, line 111:
```suggestion
    private void unlink(Node<T> current) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,138 @@
...
+                current = current.next;
+            }
+        } else {
+            current = tail;
+            for (int i = size - 1; i > index; i--) {
+                current = current.prev;
+            }
+        }
+        return current;
+    }
```
zachar-chorny, line 109: It's better to transfer this logic to get() method, because otherwise there is only one line, and use get method instead of findNode method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,138 @@
...
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private Node<T> findNode(int index) {
+        if (index >= size || index < 0) {
+            throw new ArrayIndexOutOfBoundsException("Index of bounds: " + index);
+        }
```
zachar-chorny, line 95: Put this check in a separate method and use it in all methods where you work with the index

zachar-chorny (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
roman680: Why did I need to do this if every time when I'm working with index I'm using findNode method that include this check?

kozhukhovsky (approved):

PR #675, opened 2022-09-12
Implement all class methods

dogmax296:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
...
+        }
+        size--;
+        return node.value;
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> node = start;
+        for (int i = 0; i < size; i++) {
+            if ((object == null && object == node.value) || node.value.equals(object)) {
```
Rommelua, line 72: You will get NPE if **object != null** and **node.value == null**
just use null-safe equals expression here.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
...
+            start = newNode;
+        } else {
+            newNode = new Node<>(value, node, node.prev);
+            node.prev.next = newNode;
+            node.prev = newNode;
+        }
+        size++;
+    }
+
+    private static class Node<E> {
```
Rommelua, line 155:
```suggestion
    private static class Node<T> {
```
You can use **T** as well in this class.

Rommelua (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
dogmax296: if I delete (object == null && object == node.value) test testRemoveByItemPositive() is not passed
if you talking about Objects class - we can't use it in this task

src/main/java/core/basesyntax/MyLinkedList.java
dogmax296: OK

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> start;
+    private Node<T> end;
+    private int size = 0;
```
aaananas, line 8:
```suggestion
    private int size;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> start;
+    private Node<T> end;
```
```suggestion
    private Node<T> head;
    private Node<T> tail;
```
you have methods addToTail, so lets call fields ahead and tail accordingly

src/main/java/core/basesyntax/MyLinkedList.java
aaananas, line 72:
```suggestion
            if ((object == node.value) || node.value != null && node.value.equals(object)) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
...
+            }
+        }
+        node = start;
+        for (int i = 0; i < size; i++) {
+            if (index == i) {
+                break;
+            }
+            node = node.next;
+        }
+        return node;
```
aaananas, line 110:
```suggestion
        if (index > size >>> 1) {
            node = end;
            for (int i = size - 1; i >= 0; i--) {
                if (index == i) {
                    break;
                }
                node = node.prev;
            }
        } else {
            node = start;
            for (int i = 0; i < size; i++) {
                if (index == i) {
                    break;
                }
                node = node.next;
            }
        }
        return node;
```

aaananas (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
dogmax296: OK

src/main/java/core/basesyntax/MyLinkedList.java
dogmax296: OK

src/main/java/core/basesyntax/MyLinkedList.java
dogmax296: OK

src/main/java/core/basesyntax/MyLinkedList.java
dogmax296: OK

PR #682, opened 2022-09-25
v.1.0 made all methods work like default ones

yadiver82:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,166 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size;
+    private Node<T> first;
+    private Node<T> last;
+
+    private static class Node<T> {
```
Rommelua, line 10: Private class should be located after public methods. Or even after private ones.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,166 @@
...
+            linkBefore(value, node(index));
+        }
     }
 
     @Override
     public void addAll(List<T> list) {
+        @SuppressWarnings("unchecked") T[] listArray = (T[]) list.toArray();
+        if (listArray.length == 0) {
+            throw new RuntimeException("The length of input list is 0");
+        }
```
Rommelua, line 42: We do not need to convert to the array.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,166 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    Node<T> node(int index) {
```
Rommelua, line 98:
```suggestion
    private Node<T> getNode(int index) {
```
**All** non-public methods must be **private** by default! You should have some reason for making them protected or package-private.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,166 @@
...
     }
 
     @Override
     public void add(T value, int index) {
+        checkPositionIndex(index);
+        if (index == size) {
+            linkLast(value);
+        } else {
+            linkBefore(value, node(index));
+        }
```
Rommelua, line 34: Better remove checking here and add it into `getNode(index)` method.

Rommelua (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
yadiver82: Super! Thanks, I deleted checkPositionIndex() at all! I left only checkElementIndex() for getNode!

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,153 @@
...
+            }
+        } else {
+            for (Node<T> node = first; node != null; node = node.next) {
+                if (object.equals(node.item)) {
+                    unlink(node);
+                    return true;
+                }
+            }
+        }
         return false;
```
aaananas, line 65:
```suggestion
        for (Node<T> node = first; node != null; node = node.next) {
                if (node.item == object || object != null && object.equals(node.item)) {
                    unlink(node);
                    return true;
                }
            }
        }
        return false;
```

aaananas (changes requested):
great job

src/main/java/core/basesyntax/MyLinkedList.java
yadiver82: genius!!!

yadiver82: I rewrited this part of code many times, and did not find such compact solution!

aaananas (approved):

PR #693, opened 2022-10-07
implemented task

vov4ik89:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,154 @@
...
+            }
+            return node;
+        } else {
+            Node<T> node = tail;
+            for (int i = size - 1; i > index; i--) {
+                node = node.prev;
+            }
+            return node;
+        }
+    }
```
CyberDogFK, line 124: I hear in programming practice recommended avoid few return point`s in method. Some time we really need this. But in this case, i think, we can use variable for saving value, and return it in the end of method, in one return point

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,154 @@
...
+        } else {
+            for (Node<T> i = head; i != null; i = i.next) {
+                if (equalsObjects(object, i.value)) {
+                    unlink(i);
+                    return true;
+                }
+            }
+        }
         return false;
     }
```
CyberDogFK, line 98: Good decision with cycles but, can we in some way make less return point`s from method?

CyberDogFK (changes requested):
Good solution, just one little code style approving

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,152 @@
...
     @Override
     public T remove(int index) {
-        return null;
+        return unlink(getNode(index));
     }
 
     @Override
     public boolean remove(T object) {
-        return false;
+        boolean remove = false;
```
ChumakAn, line 78: you can avoid declaring it here

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,152 @@
...
+        if (object == null) {
+            for (Node<T> i = head; i != null; i = i.next) {
+                if (i.value == null) {
+                    unlink(i);
+                    remove = true;
```
ChumakAn, line 83: and just return true here

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,152 @@
...
+                }
+            }
+        } else {
+            for (Node<T> i = head; i != null; i = i.next) {
+                if (equalsObjects(object, i.value)) {
+                    unlink(i);
+                    remove = true;
```
ChumakAn, line 90: same for this

ChumakAn (changes requested):
Good! One minor comment

PR #694, opened 2022-10-07
Implement Node and LinkedList classes

anastasiavasylyk:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,189 @@
...
+        checkPositionIndex(index);
+        if (index == size) {
+            linkLast(value);
+        } else {
+            linkBefore(value, node(index));
+        }
     }
 
     @Override
     public void addAll(List<T> list) {
```
Bodddd, line 113: the solution is much simpler than it seems, you already have a method that adds to the list (add() line 98). so you only need to get all the elements and put them in one by one

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,189 @@
...
+
+    T unlink(Node<T> removedElement) {
+        final T element = removedElement.value;
+        final Node<T> next = removedElement.next;
+        final Node<T> prev = removedElement.prev;
+        if (prev == null) {
+            first = next;
+        } else {
+            prev.next = next;
+            removedElement.prev = null;
```
Bodddd, line 84: line 84, 90 and 92 are not needed, since we remove the ability to get into this node at line 83 and 89

Bodddd (changes requested):
Good job. Let’s try to simplify a little

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,189 @@
...
     @Override
     public T remove(int index) {
-        return null;
+        checkElementIndex(index);
+        return unlink(node(index));
     }
 
     @Override
     public boolean remove(T object) {
+        if (object == null) {
```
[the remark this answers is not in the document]
anastasiavasylyk: in Try to avoid common mistakes file states that it cannot be used Object class, so I can`t use Object.equals

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private static class Node<T> {
```
Bodddd, line 6: usually private nested classes are placed at the end of the main class (after unlink() method)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
+
+    private void checkElementIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Invalid index");
+        }
+    }
+
+    private Node<T> node(int index) {
+        // assert isElementIndex(index);
+
```
Bodddd, line 110: empty line

src/main/java/core/basesyntax/MyLinkedList.java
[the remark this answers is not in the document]
Bodddd: Yes, but you can create your own implementation of Object.equals

Bodddd (changes requested):
near finish

Bodddd (approved):
Good job🔥

liliia-ponomarenko (changes requested):
Great job🔥 Let’s improve your solution 😉

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,157 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size;
+    private Node<T> first;
+    private Node<T> last;
+
+    public MyLinkedList() {
+        this.size = 0;
+    }
```
liliia-ponomarenko, line 12:
```suggestion
```
you don’t  need to initialize this variable (default  value 0)

liliia-ponomarenko (approved):
great 🚀

src/main/java/core/basesyntax/MyLinkedList.java
anastasiavasylyk: okay, I get it

PR #700, opened 2022-10-09
Homework

alex-anofriev:


src/test/java/core/basesyntax/MyLinkedListTest.java
```diff
@@ -338,23 +341,70 @@ public void removeObjectValueByItemPositive() {
         cats.add(FIRST_CAT);
         cats.add(SECOND_CAT);
         cats.add(THIRD_CAT);
-        Assert.assertEquals("Expected size is incorrect", 3, cats.size());
```
Pavlo-Gook, line 341: We cannot change the test code

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,209 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private static final int DEFAULT_INDEX = 0;
+    private static final int DIVIDER = 2;
+    private Node<T> first;
+    private Node<T> last;
+    private Node<T> current;
+    private Node<T> temp;
+    private Node<T> secondTemp;
```
Pavlo-Gook, line 12: There is no need to create temporary variables as fields

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,209 @@
...
+        temp.prevElement = secondTemp;
+        temp.nextElement = current;
+        secondTemp.nextElement = temp;
+        current.prevElement = temp;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
+        for (T t : list) {
```
Pavlo-Gook, line 56: It would be better to name the element like 'item' instead 't'

Pavlo-Gook (changes requested):

Pavlo-Gook (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,172 @@
...
+            return;
+        }
+        Node<T> current = getNodeByIndex(index);
+        Node<T> newNode = new Node<>(value);
+        newNode.prevElement = current.prevElement;
+        newNode.nextElement = current;
+        current.prevElement.nextElement = newNode;
+        current.prevElement = newNode;
+        size++;
     }
```
hronograf, line 51: I will not request changes if you leave it as it is

and at lines 44-50 code is more readable because you wrote `Node<T> newNode = new Node<>(value);` 
and then it is clear what we do when we write `newNode.prevElement = current.prevElement;`

but at 33-36 lines, you wrote `first = new Node<>(value);` 
and when we write `first.nextElement = current;` it is harder to read and understand what is going on

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,172 @@
...
+            add(element);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        indexChecker(index);
+        return getNodeByIndex(index).value;
     }
```
hronograf, line 64: I think we should always call `indexChecker` method before `getNodeByIndex`,  
so can we move `indexChecker` inside `getNodeByIndex`?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,172 @@
...
     public int size() {
-        return 0;
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size() == DEFAULT_INDEX;
+    }
```
hronograf, line 105: same about `== 0`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,172 @@
...
+
+    private void indexCheckerForAdding(int index) {
+        if (index < DEFAULT_INDEX || index > size()) {
+            throw new IndexOutOfBoundsException("Index is out of bounds for size "
+                    + size() + " with element " + index);
+        }
+    }
+
+    private Node<T> getNodeByIndex(int index) {
+        if (index <= size() / DIVIDER) {
```
hronograf, line 122: we also can use `2` instead of `DIVIDER` 
it is clear that we check if index in the first or second half of our list

hronograf (changes requested):
Good!

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,172 @@
...
+        } else {
+            prevNode.nextElement = nextNode;
+        }
+        if (nextNode == null) {
+            last = prevNode;
+        } else {
+            nextNode.prevElement = prevNode;
+        }
+        size--;
+    }
```
hronograf, line 163: image
after 153 line we should write `nextNode.prevElement = null` and similar for other cases

hronograf (changes requested):

alex-anofriev:
There was error on github after I was adding a pull request because tests were changed in mate repository a week ago. So i changed them too and then checks passed.

hronograf (changes requested):
looks good
and we mustn't change tests, so revert your changes and if tests fail, look at errors and try to resolve them

hronograf (approved):
Nice!

PR #701, opened 2022-10-10
Created MyLinkedList class.

PavloPolovyi:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,185 @@
...
     public T get(int index) {
-        return null;
+        checkIndex(index);
+        if (index == 0) {
+            return head.item;
+        } else if (index == size - 1) {
+            return tail.item;
+        } else {
+            return findNode(index).item;
+        }
```
nazariipertsak, line 62: Do we need this else block? If you are returning in if blocks

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,185 @@
...
+            prev.next = next;
+            node.prev = null;
+        }
+        if (next == null) {
+            tail = prev;
+        } else {
+            next.prev = prev;
+            node.next = null;
+        }
+        node.item = null;
```
nazariipertsak, line 184: Do we need to set item of unlinked to null?

nazariipertsak (changes requested):

nazariipertsak (approved):
Great

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,183 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    public MyLinkedList() {
+    }
```
liliia-ponomarenko, line 11: Do you really need this constructor?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,183 @@
...
+            }
+        } else {
+            for (Node<T> x = head; x != null; x = x.next) {
+                if (object.equals(x.item)) {
+                    unlink(x);
+                    return true;
+                }
+            }
+        }
         return false;
```
liliia-ponomarenko, line 87: Can you combine these loops?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,183 @@
...
+        size++;
+    }
+
+    private void linkBefore(T value, Node<T> before) {
+        Node<T> newNode = new Node<>(before.prev, value, before);
+        newNode.prev.next = newNode;
+        before.prev = newNode;
+        newNode.next = before;
+        size++;
+    }
```
liliia-ponomarenko, line 126: these methods are similar, can you rewrite your code and use only one?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,183 @@
...
 
     @Override
     public T get(int index) {
-        return null;
+        checkIndex(index);
+        if (index == 0) {
+            return head.item;
+        } else if (index == size - 1) {
+            return tail.item;
+        }
```
liliia-ponomarenko, line 48: Do you really need these conditionals?  method findNode() also return these values

liliia-ponomarenko (changes requested):
I like your solution 👍🏻 Let's improve it 😉

liliia-ponomarenko (approved):
Good job! 🔥😎

PR #702, opened 2022-10-10
Added body to methods in MyLinkedList, all tests were passed

arthur4655:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,174 @@
...
+    private void linkMiddle(T value, int index) {
+        indexValidation(index);
+        currentNode = findByIndex(index);
+        newNode = new Node<>(currentNode.prev, value, currentNode);
+        currentNode.prev.next = newNode;
+        currentNode.prev = newNode;
+        currentSize++;
+    }
+
+    private void unlink(Node<T> node) {
```
AndriiMatveev, line 149: Can we return something to simplify another methods?

AndriiMatveev (changes requested):

AndriiMatveev (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,169 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int currentSize;
+    private Node<T> first;
+    private Node<T> last;
+    private Node<T> newNode;
+    private Node<T> currentNode;
```
hronograf, line 10: this two fields isn't a state of `MyLinkedList` object, 
it is a state of methods, so let's move them to methods

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,169 @@
...
+
+    static class Node<T> {
```
hronograf, line 12: image

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,169 @@
...
+        }
+        currentSize++;
+    }
+
+    private void linkLast(T value) {
+        newNode = new Node<>(last, value, null);
+        last.next = newNode;
+        last = newNode;
+        currentSize++;
+    }
```
hronograf, line 131: it this method should `linkLast` it shouldn't increase `currentSize`
it should just `link`
let's move `currentSize++;` to `add` methods

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,169 @@
...
+        for (T t : list) {
+            add(t);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        indexValidation(index);
+        return findByIndex(index).item;
```
hronograf, line 57: looks like we always call `indexValidation` before `findByIndex`
so let's move `indexValidation` in `findByIndex`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,169 @@
...
+            }
+        } else {
+            while (currentNode != null) {
+                if (object.equals(currentNode.item)) {
+                    unlink(currentNode);
+                    return true;
+                }
+                currentNode = currentNode.next;
+            }
+        }
```
hronograf, line 94: we can combine situations when `object == null` and `object != null`
(check it inside if-statement, implementation of `Objects.equals` may help you)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,169 @@
...
+            first = node.next;
+        } else if (node.next == null && node.prev != null) {
+            node.prev.next = null;
+        } else if (node.next != null) {
+            node.prev.next = node.next;
+            node.next.prev = node.prev;
+        }
+        currentSize--;
+        return value;
+    }
```
hronograf, line 155: what if `node.prev == null && node.next == null`? (case when we have only one node)

hronograf (changes requested):
Good!
I like that you have separate methods `linkFirst`, `linkLast` and `linkMiddle`
logic is divided and easy to read

hronograf (approved):
Good job!

PR #703, opened 2022-10-10
homework has been done

AndriiMatveev:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,186 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size = 0;
+    private Node<T> first;
+    private Node<T> last;
+
+    private void checkPositionIndex(int index) {
+        if (!(index >= 0 && index <= size)) {
+            throw new IndexOutOfBoundsException("This index is out of bound " + index);
```
arthur4655, line 12: It is recommended to use constants over literals

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,170 @@
...
 
     @Override
     public void addAll(List<T> list) {
+        Object[] a = list.toArray();
+        if (a.length == 0) {
+            return;
+        }
+        for (Object object : a) {
+            linkLast((T) object);
+        }
```
arthur4655, line 107: We can use foreach with List<T> without creating Object[] array, and also we don't need to check for empty list, because if it is empty, cycle won't start

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,166 @@
...
+    }
+
+    private void checkElementIndex(int index) {
+        if (!(index >= 0 && index < size)) {
+            throw new IndexOutOfBoundsException(EXCEPTION_MESSAGE + index);
+        }
+    }
+
+    Node<T> node(int index) {
```
ChumakAn, line 23: let's rename it to make it more clear what's the purpose of the method is

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,166 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private static final String EXCEPTION_MESSAGE = "This index is out of bound ";
+    private int size;
+    private Node<T> first;
+    private Node<T> last;
```
ChumakAn, line 9: what about renaming to head and tail? Optional, but helps to be consistent with LL impl

ChumakAn (changes requested):
Good! Small comments

ChumakAn (approved):

PR #704, opened 2022-10-11
implemented LinkedList class

nazariipertsak:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,217 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void checkIndex(int index) {
+        if (index > size || index < 0) {
+            throw new IndexOutOfBoundsException("Index out of bounds");
```
PavloPolovyi, line 124: You can make this exception more informative, e.g. "index 2 out of bounds".

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,217 @@
...
+        }
+    }
+
+    private void checkIndexGetSet(int index) {
+        if (index + 1 > size || index < 0) {
+            throw new IndexOutOfBoundsException("Index out of bounds");
```
PavloPolovyi, line 130: You can make this exception more informative, e.g. "index 2 out of bounds".

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,217 @@
...
+            for (int i = 0; i < index; i++) {
+                node = node.next;
+            }
+        }
+        node = tail;
+        for (int i = size - 1; i > index; i--) {
+            node = node.prev;
+        }
+        return node;
+    }
```
PavloPolovyi, line 155: It seemed that you reassign value of node variable after first loop in second loop. Maybe add else clause or one more return in if clause? Because if this statement (if (index < (size >> 1))) is true, it does job, but after that second loop do it again and change value of node.

PavloPolovyi (changes requested):

PavloPolovyi (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,219 @@
...
+        Node<T> newNode = new Node<>(null, value, null);
+        Node<T> oldNode = getIndexNode(index);
+        if (size == 0) {
+            addFirstElement(newNode);
+            return;
+        }
+        if (index == size || index == 0) {
+            addTopBottom(newNode, index);
+            return;
+        }
```
liliia-ponomarenko, line 46: can we avoid using the custom method for addFirstElement addTopBottom? For  ```index == size``` you can use method  add(T value) for another case you can use something  similar: 
```
        Node<T> next = getIndexNode(index);
        Node<T> prev = next.prev;
        Node<T> newNode = new Node<>(prev, value, next);
        if (prev == null) {
            first = newNode;
        } else {
            prev.next = newNode;
        }
        next.prev = newNode;
        size++;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,219 @@
...
+        if (index == 0) {
+            removeBottom();
+            return removedItem;
+        }
+        if (index == size - 1) {
+            removeTop();
+            return removedItem;
+        }
+        removeNode(node);
+        return removedItem;
```
liliia-ponomarenko, line 91: the same is for method remove() is it really necessary to have different methods for these actions? Can we simplify this logic?

liliia-ponomarenko (changes requested):
Good job 👍  Let’s  improve your solution 😉

liliia-ponomarenko (changes requested):
Let’s simplify your code😉

liliia-ponomarenko (approved):
great! 😊

PR #709, opened 2022-10-12
Hw linked list solution

Bohdan20b:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,137 @@
...
+        private T value;
+        private Node<T> prev;
+        private Node<T> next;
+
+        public Node(Node<T> prev,T value, Node<T> next) {
+            this.value = value;
+            this.prev = prev;
+            this.next = next;
+        }
     }
```
yehbreus, line 138: i think nested class should be above class methods

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,137 @@
...
+        return foundNode;
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayIndexOutOfBoundsException();
+        }
+    }
+
+    private T unlink(Node<T> node) {
```
yehbreus, line 109: i think this method should return void

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,137 @@
...
+        Node<T> previous = node.prev;
+        Node<T> next = node.next;
+        if (previous == null && next == null) {
+            head = tail = null;
```
yehbreus, line 113: i think this check is redundant

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,137 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private Node<T> findIndex(int index) {
```
yehbreus, line 93: method returns Node<T>. Maybe we need to change his name?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,137 @@
...
+            i++;
+        }
+        return foundNode;
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new ArrayIndexOutOfBoundsException();
+        }
+    }
```
yehbreus, line 107: you can merge this part with findIndex() method, and you won't need to call it before calling findIndex everytime

yehbreus (changes requested):
Check comments, please)

yehbreus (approved):

src/main/java/core/basesyntax/MyLinkedList.java
liliia-ponomarenko: usually, private nested classes (along with their methods) are placed after the upper-class methods

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,133 @@
...
+            tail.next = newNode;
+        }
+        tail = newNode;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new ArrayIndexOutOfBoundsException();
```
liliia-ponomarenko, line 37: you can use ```checkIndex(int index)```  after the next conditional block instead this (but add return statement after ```add()```)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,133 @@
...
+            newNode.prev = newNode;
+            head = newNode;
+            size++;
+        } else {
+            Node<T> searchIndex = findNode(index);
+            Node<T> newNode = new Node<>(searchIndex.prev, value, searchIndex);
+            searchIndex.prev.next = newNode;
+            searchIndex.prev = newNode;
+            size++;
+        }
```
liliia-ponomarenko, line 51:
```suggestion
            Node<T> next = findNode(index);
            Node<T> prev = next.prev;
            Node<T> newNode = new Node<>(prev, value, next);
            if (prev == null) {
               head = newNode;
            } else {
               prev.next = newNode;
            }
            next.prev = newNode;
            size++;
```
you can simplify  this code and avoid code duplication

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,133 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private Node<T> findNode(int index) {
```
liliia-ponomarenko, line 102: It’s correct but you can improve the performance of your LinkedList (find a hint in the documentation😉).

liliia-ponomarenko (changes requested):
Great job🔥 Let’s improve your solution 😉

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,139 @@
...
+        private T value;
+        private Node<T> prev;
+        private Node<T> next;
+
+        public Node(Node<T> prev,T value, Node<T> next) {
+            this.value = value;
+            this.prev = prev;
+            this.next = next;
+        }
+    }
```
KostyaNikiforov, line 20: Usually inner classes are placed at the end of the file

KostyaNikiforov (approved):

PR #711, opened 2022-10-12
Implemented task

Veronika-Tretiak:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,132 @@
...
+        for (T value: list) {
+            add(value);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        Node<T> newNode = getNodeByIndex(index);
+        return newNode.value;
```
yehbreus, line 63: i think we can use "return getNodeByIndex(index).value"

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,132 @@
...
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        T valueNode = getNodeByIndex(index).value;
+        getNodeByIndex(index).value = value;
```
yehbreus, line 69: are we realy need to call this method here? or better will be use valueNode.value = value?)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,132 @@
...
+
+    public Node<T> getNodeByIndex(int index) {
+        if (index < 0 && index >= size) {
+            throw new IndexOutOfBoundsException("Index is not correct: "
+                            + index + ", for Size: " + size);
+        }
+        if (!(index >= 0 && index < size)) {
+            throw new IndexOutOfBoundsException("Index is not correct: "
+                    + index + ", for Size: " + size);
+        }
```
yehbreus, line 113: are we realy need this?)

yehbreus (approved):
Approved, but check comments, please)

liliia-ponomarenko (changes requested):
Great job! 🔥 Let's improve your solution😉

liliia-ponomarenko (approved):
LGTM😎🔥

PR #712, opened 2022-10-12
solution

vyshnevsky:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,145 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private Node<T> findByIndex(int index) {
+        if (index < 0 || index >= size) {
```
Bodddd, line 94: It would be better to create a separate method for checking the correctness of index

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,145 @@
...
+            throw new IndexOutOfBoundsException("Index " + index + " is not valid, "
+                    + "for size: " + size);
+        }
+        int currentIndex;
+        Node<T> currentNode;
+        if (index < size / 2) {
```
Bodddd, line 100: in my opinion, it is better to use the ternary operator here, then the code size will be smaller and not look like a duplication

Bodddd (changes requested):
Almost perfect!

src/main/java/core/basesyntax/MyLinkedList.java
vyshnevsky: Thanks, it's a good idea, but this solution is the most understandable for me

Bodddd (approved):
Good job🔥

liliia-ponomarenko (changes requested):
Great job🔥 let's improve your solution a little😉

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,149 @@
...
     }
 
     @Override
     public void add(T value, int index) {
+        if (index == size) {
+            add(value);
+            return;
+        }
+        Node<T> newNode = new Node<>(null, value, null);
+        if (index == 0) {
```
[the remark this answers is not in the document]
vyshnevsky: > do you really need custom handling for this case? Can you combine this and the next block? avoid code duplication

I tried to do this, but I can't.  Could you please tell me how to do this?
https://pastenow.ru/J7BS7

KostyaNikiforov (approved):

PR #713, opened 2022-10-12
implemented LinkedList class

yehbreus:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,153 @@
...
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    public MyLinkedList() {
+        this.head = null;
+        this.tail = null;
+    }
```
Bohdan20b, line 13: why do you need this?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,153 @@
...
+    private class Node<T> {
+        private T value;
+        private Node<T> next;
+        private Node<T> prev;
+
+        public Node(T value) {
+            this.value = value;
+            this.next = null;
+            this.prev = null;
+        }
```
Bohdan20b, line 24: i would rather use this constructor:
public Node(Node<T> prev,T value, Node<T> next) {
            this.value = value;
            this.prev = prev;
            this.next = next;

Bohdan20b (approved):

src/main/java/core/basesyntax/MyLinkedList.java
yehbreus: yep, my bad, already done

src/main/java/core/basesyntax/MyLinkedList.java
yehbreus: I think my constructors is better, because we don't have to pass null values ​​to the method. Maybe menthor help us with this)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,153 @@
...
+    }
+
     @Override
     public void add(T value) {
+        Node<T> newNode = new Node<>(value);
+        Node<T> currentNode = head;
```
ChumakAn, line 30: it's redundant variable

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,153 @@
...
+            head.prev = new Node<>(value);
+            head.prev.next = head;
+            head = head.prev;
+        } else {
+            Node<T> currentNode = getNode(index);
+            Node<T> newNode = new Node<>(value);
+            newNode.prev = currentNode.prev;
+            newNode.next = currentNode;
+            currentNode.prev.next = newNode;
+            currentNode.prev = newNode;
```
ChumakAn, line 57: you could make this block less if you not set tail and head null in constructor

src/main/java/core/basesyntax/MyLinkedList.java
ChumakAn: you don't need this too

ChumakAn (changes requested):
Good! Let's improve

ChumakAn (approved):

PR #716, opened 2022-10-13
My LinkList

anton-pushkar:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,129 @@
...
-        return false;
+        return size == 0;
+    }
+
+    private Node<T> getNodeByIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("index " + index + " out of bounds " + size);
+        }
+        Node<T> currentNode = head;
+        for (int i = 0; i < index; i++) {
```
Serhii-Prykhodko, line 112: Common mistake. You have to move from tail to center, if index > size / 2. It can increase LinkedList performance.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,129 @@
...
+        for (T t : list) {
+            add(t);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        Node<T> node = getNodeByIndex(index);
+        return node.value;
```
Serhii-Prykhodko, line 64: Do you really need to create a new Node?  I think 
return getNodeByIndex(index).value;
is better solution

Serhii-Prykhodko (changes requested):

Serhii-Prykhodko (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,128 @@
...
+        private Node<T> prev;
+        private T value;
+        private Node<T> next;
+
+        private Node(Node<T> prev, T value, Node<T> next) {
+            this.prev = prev;
+            this.value = value;
+            this.next = next;
+        }
+    }
```
KostyaNikiforov, line 20: It is a good practice to place inner classes at the end of the file

KostyaNikiforov (approved):
🔥

PR #718, opened 2022-10-13
LinkedList implementation

AlinaOryshak:


Danyl-Kniaziev (commented):
Good job! Let's fix some minor mistakes

Danyl-Kniaziev (changes requested):

Danyl-Kniaziev (approved):
LGFM🔥!

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,182 @@
...
+        size++;
+    }
+
+    private void insert(T value, Node<T> afterNewNode) {
+        Node<T> beforeNewNode = afterNewNode.prev;
+        Node<T> newNode = new Node<>(beforeNewNode, value, afterNewNode);
+        beforeNewNode.next = newNode;
+        afterNewNode.prev = newNode;
+        size++;
+    }
```
hronograf, line 110: I like this naming

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,182 @@
...
 
     @Override
     public boolean remove(T object) {
+        int index = getIndex(object);
+        if (isElementIndex(index)) {
+            remove(index);
+            return true;
+        }
         return false;
     }
```
hronograf, line 80: if we rename `getIndex` to `getNode` and it will return `Node<T>`, then we can just unlink returned node
because now we go through our list in `getIndex` method and after that in `getObjByIndex` (inside `remove` method)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,182 @@
...
+            add(item);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkElementIndex(index);
+        return getObjByIndex(index).value;
     }
```
hronograf, line 53: looks like we always call `checkElementIndex` before `getObjByIndex`
so let's move `checkElementIndex` to `getObjByIndex`

hronograf (changes requested):
Good job!

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,183 @@
...
     public T get(int index) {
-        return null;
+        return getObjByIndex(index).value;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        checkElementIndex(index);
+        Node<T> current = getObjByIndex(index);
```
hronograf, line 57: we already moved `checkElementIndex` to `getObjByIndex`, so do we need this check at 56 line?
(you should check it in other places too)

hronograf (changes requested):

hronograf (approved):
Great! 🚀

PR #723, opened 2022-10-14
implemented task

Iryna-a:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,144 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private static final int DEFAULT_SIZE = 0;
```
ana-rom, line 6: Do you think we need a default size for LL? Maybe we can omit it?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,144 @@
...
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    public MyLinkedList() {
+        size = DEFAULT_SIZE;
```
ana-rom, line 12: and here you will have an empty constructor then

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,144 @@
...
+            }
+            element = element.next;
+        }
         return false;
     }
 
     @Override
     public int size() {
-        return 0;
+        return size;
```
ana-rom, line 94: I'm not sure, but it seems to me when you add 2 "return" statements, you will never go through the second return  because the method will finish its work with the first one "return"
correct me if I'm wrong please

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,144 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
```
ana-rom, line 99: the same question
I would do just the last one "return size == 0;", it will give you true if size is empty and false if it's not.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,144 @@
...
+        }
+        tail = node;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException();
+        }
```
ana-rom, line 32: just an idea, maybe if it makes sense for you, you can create one more method as you did for checkIndex(int index), and here just call it
btw, should we write some message about exceptions, shouldn't we?

Iryna-a: I am not sure about putting this in a saparate method because this kind of condition occurs only once

ana-rom (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,142 @@
...
+        if (element == head) {
+            head = element.next;
+        } else if (element == tail) {
+            tail = element.prev;
+        } else {
+            element.prev.next = element.next;
+            element.next.prev = element.prev;
+        }
+        size--;
+    }
```
hronograf, line 131: image
(as example, after 123 line if `head != null` we should write `head.prev = null`)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,142 @@
...
+        for (T item : list) {
+            add(item);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkIndex(index);
+        Node<T> element = findElement(index);
```
hronograf, line 56: looks like we always call `checkIndex ` before `findElement `
so let's move `checkIndex` in `findElement` method

hronograf (changes requested):
Good!

ChumakAn (approved):

PR #725, opened 2022-10-14
Task complited.

ZoreslavT:
Also three methods getNode(), unLink() & chekindex() was added;

### DIFF
--- a/src/main/java/core/basesyntax/MyLinkedList.java
+++ b/src/main/java/core/basesyntax/MyLinkedList.java
@@ -119,7 +119,7 @@ public class MyLinkedList<T> implements MyLinkedListInterface<T> {
     }
 
     private Node<T> getNode(int index) {
-        checkPositionIndex(index);
+        checkElementIndex(index);
         if (size == 1) {
             return head;
         }
@@ -130,7 +130,6 @@ public class MyLinkedList<T> implements MyLinkedListInterface<T> {
             }
             return result;
         } else {
-            checkElementIndex(index);
             Node<T> result = tail;
             for (int i = size - 1; i > index; i--) {
                 result = result.prev;


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,186 @@
...
+            return result.value;
+        } else if (index == size) {
+            throw new IndexOutOfBoundsException("Invalid index " + index);
+        } else {
+            Node<T> result = tail;
+            for (int i = size - 1; i > index; i--) {
+                result = result.prev;
+            }
+            return result.value;
+        }
```
Veronika-Tretiak, line 92: You have a method getNode. Can you use it here to make you code shorter?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,186 @@
...
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        if (index < 0 || index >= size && head != null) {
+            throw new IndexOutOfBoundsException("Invalid index " + index);
+        }
```
Veronika-Tretiak, line 99: Don’t create repeating code.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,186 @@
...
+        getNode(index).value = value;
+        return removedValue;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (index < 0 || index >= size && head != null) {
+            throw new IndexOutOfBoundsException("Invalid index " + index);
+        }
```
Veronika-Tretiak, line 109: Don’t create repeating code.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,186 @@
...
+        if (next == null) {
+            tail = prev;
+        } else {
+            next.prev = prev;
+            currentNode.next = null;
+        }
+        size--;
+        return currentNode.value;
+    }
```
Veronika-Tretiak, line 166: Is it good practice to set the value to null?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,186 @@
...
+        public Node(T value) {
+            this.value = value;
+        }
+    }
+
     @Override
     public void add(T value) {
+        currentNode = new Node<>(value);
+        Node<T> t = tail;
+        if (head == null) {
```
Veronika-Tretiak, line 31: it is better to rely on size == 0 in this case

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,186 @@
...
+        private Node<T> next;
+
+        public Node(T value) {
+            this.value = value;
+        }
+    }
+
     @Override
     public void add(T value) {
+        currentNode = new Node<>(value);
```
Veronika-Tretiak, line 29: You create currentNode, but you use it only one time. And create a lot of newNode

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,171 @@
...
+    private Node<T> head;
+    private Node<T> tail;
+    private Node<T> currentNode;
+    private int size;
+
+    public MyLinkedList() {
+        this.head = null;
+        this.tail = null;
+        size = 0;
+    }
```
KostyaNikiforov, line 15: All of these are default values. You can remove the constructor

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,171 @@
...
+        if (object == null) {
+            for (Node<T> n = head; n != null; n = n.next) {
+                if (n.value == null) {
+                    unLink(n);
+                    return true;
+                }
+            }
+        } else {
+            for (Node<T> n = head; n != null; n = n.next) {
+                if (n.value == object || n.value != null && n.value.equals(object)) {
```
KostyaNikiforov, line 94: If the `object` is null and n.value is null, this expression will return true

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,171 @@
...
+    }
+
+    private void checkPositionIndex(int index) {
+        if (index < 0 || index > size()) {
+            throw new IndexOutOfBoundsException("Invalid index " + index);
+        }
+    }
+
+    private void checkElementIndex(int index) {
+        if (index < 0 || index >= size && head != null) {
```
KostyaNikiforov, line 120: Why do you check the `head` here?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,171 @@
...
+        }
+        if (index < (size / 2)) {
+            Node<T> result = head;
+            for (int i = 0; i < index; i++) {
+                result = result.next;
+            }
+            return result;
+        } else if (index == size) {
+            throw new IndexOutOfBoundsException("Invalid index " + index);
+        } else {
```
KostyaNikiforov, line 155: Can you use `checkElementIndex` method instead of this if-block?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,171 @@
...
+            tail = currentNode;
+            head.next = tail;
+            tail.prev = head;
+        } else {
+            Node<T> lastNode = tail;
+            tail = currentNode;
+            tail.prev = lastNode;
+            tail.prev.prev.next = lastNode;
+            lastNode.next = tail;
+        }
```
KostyaNikiforov, line 32: If the size if not 0, it would be enough to just find the last element and set the `next` field with the new node

KostyaNikiforov (changes requested):

ZoreslavT:
1) Changed method add();
2) Constructor was removed
3) Deleted one of the blocks if in method remove(T object).
4) Used checkElementIndex() in method getNode(int index).

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,149 @@
...
+        if (index < (size / 2)) {
+            Node<T> result = head;
+            for (int i = 0; i < index; i++) {
+                result = result.next;
+            }
+            return result;
+        } else {
+            checkElementIndex(index);
+            Node<T> result = tail;
```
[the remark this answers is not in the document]
ZoreslavT: Agree, i need only checkElementIndex().

PR #728, opened 2022-10-14
created own linked list

GHSpeedWagon:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,144 @@
...
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkPositionIndex(index);
+        T oldValue = findByIndex(index).value;
+        node = findByIndex(index);
+        unlink(node);
+        return oldValue;
```
HryhorashPavlo, line 76: i think you can just return node.value
and delete T old value = findByIndex(index).value

HryhorashPavlo (changes requested):

HryhorashPavlo (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,154 @@
...
+        if (node == null) {
+            add(value);
+            return;
+        } else if (node.prev == null) {
+            head.prev = new Node<>(null, value, head);
+            head = head.prev;
+        } else {
+            Node<T> newNode = new Node<>(null, value, null);
+            newNode.next = node;
+            newNode.prev = node.prev;
```
ChumakAn, line 41: you can pass it as a constructor params

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,154 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private static final String INDEX_EXCEPTION = "Index is incorrect";
+    private Node<T> node;
```
ChumakAn, line 7: why do you need this variable? Don't you think it's redundant? You already have a tail, so you can just use local variables in method instead of making it a class variable

ChumakAn (changes requested):
Good, let's go through common mistakes and avoid creating redundant variables

ChumakAn (approved):
LGTM👍

PR #731, opened 2022-10-15
Realize methods

YuliaTsariv:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,150 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    private class Node<T> {
```
Veronika-Tretiak, line 10: Usually private nested classes (along with their methods) are placed after the upper-class methods, so let's move the Node class to the end

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,150 @@
...
+
+        public Node(T value) {
+            this.value = value;
+        }
+    }
+
     @Override
     public void add(T value) {
+        Node<T> node = new Node<>(value);
+        if (head == null) {
```
Veronika-Tretiak, line 23: it is better to rely on size == 0 in this case

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,150 @@
...
     @Override
     public T get(int index) {
-        return null;
+        return findByIndex(index).value;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+
```
Veronika-Tretiak, line 66: empty line, check all code for empty lines

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,150 @@
...
+        } else if (node.next == null) {
+            node.prev.next = null;
+            tail = node.prev;
+        } else if (node.prev == null) {
+            node.next.prev = null;
+            head = node.next;
+        } else {
+            node.next.prev = node.prev;
+            node.prev.next = node.next;
+        }
```
Veronika-Tretiak, line 148: Is it good practice to set the value to null? Think how you can improve your method

KostyaNikiforov (approved):
Cool solution! 🔥

PR #734, opened 2022-10-15
solved jv linked list task

HryhorashPavlo:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,146 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    private static class Node<T> {
```
Veronika-Tretiak, line 10: Usually private nested classes (along with their methods) are placed after the upper-class methods, so let's move the Node class to the end

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,146 @@
...
+        }
+        if (index == 0) {
+            Node<T> newNode = new Node<T>(null, head, value);
+            head.prev = newNode;
+            head = newNode;
+        } else {
+            Node<T> currentNode = getNodeByIndex(index);
+            Node<T> newNode = new Node<T>(currentNode.prev, currentNode, value);
+            currentNode.prev.next = newNode;
+            currentNode.prev = newNode;
```
Veronika-Tretiak, line 94: do you really need custom handling for this case? Can you combine this and the next block? avoid code duplication

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,146 @@
...
+        if (currentNode.next != null) {
+            currentNode.next.prev = currentNode.prev;
+        } else {
+            tail = currentNode.prev;
+        }
+        if (currentNode.prev != null) {
+            currentNode.prev.next = currentNode.next;
+        } else {
+            head = currentNode.next;
+        }
```
Veronika-Tretiak, line 64: Is it good practice to set the value to null?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,145 @@
...
+            if (currentNode.value == value || currentNode.value != null
+                    && currentNode.value.equals(value)) {
+                return currentNode;
+            }
+            currentNode = currentNode.next;
+        }
+        return null;
+    }
+
+    private void removeNode(Node<T> currentNode) {
```
ChumakAn, line 42: let's rename it to unlink 

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,145 @@
...
+    private Node<T> tail;
+    private int size;
+
+    private Node<T> getNodeByIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Index " + index
+                    + " out of bound, size is " + size);
+        }
+        if (index < size / 2) {
+            Node<T> currentNode = head;
```
ChumakAn, line 16: you can declare this variable out of if-condition and depending on the condition assign it a value to avoid creating redundant variable

ChumakAn (changes requested):
Good! Only minor comments

PR #739, opened 2022-10-17
linked list solution

B1b1z9na:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,146 @@
...
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private static final int STANDART_LIST_SIZE = 0;
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    public MyLinkedList() {
+        size = STANDART_LIST_SIZE;
+    }
```
denys-domashevskyi, line 13: I think it's unnecessary, what is the default value of int?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,146 @@
...
+        isIndexExist(index);
+        if (head == null) {
+            return null;
+        }
+        Node<T> currentNode = head;
+        int i = 0;
+        while (i != index) {
+            currentNode = currentNode.next;
+            i++;
+        }
```
denys-domashevskyi, line 31: What about adding index condition and iterate to the half of the linked list depending on the index instead of going through the whole LL?

denys-domashevskyi (changes requested):
Nice soltution!
Let's improve it!
And our methods should be in the end of class, due to code style

denys-domashevskyi (approved):
Nice job!

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,148 @@
...
+
+    private static class Node<T> {
+
+        private T item;
+        private Node<T> next;
+        private Node<T> prev;
+
+        Node(T value) {
+            this.item = value;
+        }
```
ChumakAn, line 148: what about setting the next and prev value in the constructor too?  You will be able to pass it in constructor whether it's known, and avoid writing it after object creation

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,148 @@
...
+            T value = element.item;
+            if (value == object || value != null && value.equals(object)) {
+                return element;
+            }
+            element = element.next;
+        } while (element.next != null);
+        return null;
+    }
+
+    private Node<T> checkNode(Node<T> node) {
```
ChumakAn, line 122: shouldn't it be the unlink method?

ChumakAn (changes requested):
Good, small fixes

ChumakAn (approved):

PR #745, opened 2022-10-21
Solution was added.

KatyaHumeniuk:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,174 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> first;
+    private Node<T> last;
+    private int size;
+
+    private static class Node<T> {
```
aaananas, line 10: private methods, classes should be in the end, after public

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,174 @@
...
+        checkIndex(index, size);
+        T removed = (T) findNode(index).item;
+        unlink(findNode(index), index);
+        return removed;
     }
 
     @Override
     public boolean remove(T object) {
+        int index = 0;
+        Node getNode = first;
```
aaananas, line 154:
```suggestion
        Node node = first;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,174 @@
...
+        while (getNode != null) {
+            if ((object == null && getNode.item == null)
+                    || object != null && (object.equals(getNode.item))) {
+                remove(index);
```
aaananas, line 158: just call unlink here - you already have the node
calling remove(I) you will iterate through the whole list again

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,174 @@
...
+        findNode(index).item = value;
+        return oldValue;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index, size);
+        T removed = (T) findNode(index).item;
+        unlink(findNode(index), index);
```
aaananas, line 147: call findNode() just once and save the value in variable
because you are iterating over the whole list 2 times here

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,174 @@
...
+        checkIndex(index, size);
+        return (T) findNode(index).item;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        checkIndex(index, size);
+        T oldValue = (T) findNode(index).item;
+        findNode(index).item = value;
```
aaananas, line 139: the same with iteration

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,174 @@
...
+        for (int i = size - 1; i >= size / 2; i--) {
+            if (i == index) {
+                return getNode;
+            }
+            getNode = getNode.prev;
+        }
+        return null;
+    }
+
+    private void unlink(Node node, int index) {
```
aaananas, line 80:
```suggestion
    private void unlink(Node<T> node) {
```
you can compare references 
if node == head it means that index = 0
if node == tail it means that it is the last node

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,174 @@
...
+        if (index >= size || index < 0) {
+            throw new IndexOutOfBoundsException("Index is not valid. Index: "
+                    + index + ". Size of list: " + size);
+        }
+    }
+
+    private void addMiddle(T value, int index) {
+        Node middleNode = new Node(findNode(index - 1), value, findNode(index));
+        findNode(index - 1).next = middleNode;
+        findNode(index).prev = middleNode;
```
aaananas, line 32: 4 iterations over the list
if you have some node - you can access the previous node with node.prev

aaananas (changes requested):

aaananas (approved):

PR #752, opened 2022-11-03
Made LinkedList to work as java.util.LinkedList

maks-untilov:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,136 @@
...
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException(
+                    "Can't element index " + index);
+        }
```
zachar-chorny, line 27: You can delete this check. This is because before the findByIndex check, you get to the if block if index==size

zachar-chorny (changes requested):
Well done

Rommelua (approved):

PR #753, opened 2022-11-03
Linked list solved

o-shyshkan:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,177 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> currentNode;
```
aaananas, line 6: do you need this field?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,177 @@
...
+    private int size = 0;
```
```suggestion
    private int size;
```
int value is 0 by default

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,177 @@
...
+    private Node<T> head;
+    private Node<T> tail;
+
+    private static class Node<T> {
```
aaananas, line 11: private classes should be in the end

aaananas (approved):
great job

PR #754, opened 2022-11-05
implementation MyLinkedList

BusyhinD:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,156 @@
...
+    private void checkPositionIndex(int index) {
+        if (index < 0 || index > size) {
+            outOfBoundMsg(index);
+        }
+    }
+
+    private void outOfBoundMsg(int index) {
+        throw new IndexOutOfBoundsException("Index: "
+                + index + " not found, current size: " + size);
+    }
```
aaananas, line 98:
```suggestion
    private void checkElementIndex(int index) {
        if (index < 0 || index >= size) {
            throw new IndexOutOfBoundsException("Index: "
                + index + " not found, current size: " + size);
        }
    }

    private void checkPositionIndex(int index) {
        if (index < 0 || index > size) {
            throw new IndexOutOfBoundsException("Index: "
                + index + " not found, current size: " + size);
        }
    }
```
jut throw exception, no need to use method

aaananas (approved):

PR #760, opened 2022-11-11
realized linked list methods

Nikostrik:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,135 @@
...
+        if (head == null) {
+            head = tail = newNode;
+        } else if (index == 0) {
+            newNode.next = head;
+            head = newNode;
+        } else if (index == size) {
+            tail.next = newNode;
+            newNode.prev = tail;
+            tail = newNode;
+
```
aaananas, line 37:
```suggestion
        } else if (index == size) {
            add(value);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,135 @@
...
+            tail.next = newNode;
+            newNode.prev = tail;
+            tail = newNode;
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        size++;
```
aaananas, line 25: you should not increment size at the beginning
what if the index is wrong and you would throw an exception

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,135 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    static class Node<T> {
```
aaananas, line 100: what about access modifier?

aaananas (changes requested):

aaananas (approved):

PR #769, opened 2022-11-17
1

dimonkoval:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,159 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size = 0;
```
aaananas, line 6:
```suggestion
    private int size;
```
int value is 0 by default

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,159 @@
...
+        if (!isElementIndex(index)) {
+            throw new IndexOutOfBoundsException(outOfBoundsMsg(index));
+        }
+    }
+
+    private boolean isElementIndex(int index) {
+        return index >= 0 && index < size;
+    }
+
+    T unlink(Node<T> x) {
```
aaananas, line 141:
```suggestion
    private T unlink(Node<T> x) {
```

src/main/java/core/basesyntax/Node.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public class Node<T> {
```
aaananas, line 3: make this LinkedList inner class - because it should be used only in LinkedList

aaananas (approved):
great job

PR #782, opened 2022-12-05
the custom methods were overridden

dima-holovko:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,171 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head = null;
+    private Node<T> tail = null;
+    private int size;
```
aaananas, line 8:
```suggestion
    private Node<T> head;
    private Node<T> tail;
    private int size;
```
class fields have default values

aaananas (changes requested):

PR #789, opened 2022-12-12
Linkedlistsolution

Uka0001:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,145 @@
...
+        unlink(node);
+        size--;
+        return node.value;
     }
 
     @Override
     public boolean remove(T object) {
+        for (Node<T> node = head; node != null; node = node.next) {
+            if (node.value == null && object == null
+                    || node.value != null && node.value.equals(object)) {
```
KostyaNikiforov, line 76:
```suggestion
            if (node.value == object || node.value != null && node.value.equals(object)) {
```

KostyaNikiforov (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
Uka0001: Fixed. IDEA cast expression to Objects.equals(node.value, object) , so I change your variant to IDEA variant.

KostyaNikiforov (approved):
🔥

PR #792, opened 2022-12-13
LinkedList implementation was added

vgladkyi:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,139 @@
...
+            for (int i = size - 1; i > index; i--) {
+                node = node.prev;
+            }
+            return node;
+        }
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Wrong index");
```
aleksandr-bevza, line 126: Include index and size values to this exception message

aleksandr-bevza (changes requested):
Well done! Just one small fix needed.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,139 @@
...
+        unlink(node);
+        size--;
+        return node.value;
     }
 
     @Override
     public boolean remove(T object) {
+        for (Node<T> node = first; node != null; node = node.next) {
+            if (node.value == null && object == null
+                    || node.value != null && node.value.equals(object)) {
```
zachar-chorny, line 76:
```suggestion
            if (node.value == object || node.value != null && node.value.equals(object)) {
```

zachar-chorny (approved):
Good job

PR #796, opened 2022-12-19
hw-linked-list-solution

Bogdan-Pokhilko:
Created constant value: size, head and tail;
Created class Node with value, prev and next;
Created a method for Node with inputs and inputs executions; Created addHead method, which adds 1st element to the list and checks if array is empty, or already consists of elements; Created addTail method which adds the last element to the list and links with previous element; Created addMidNode method, which adds new element between existing elements and links between them; Created getNodeIndex method, which searches for the correct element by given index. Depending on the number of the element and with help of for loop, moving from head, or tail of the list to find the right element; Created removeNode method which removing the element and unlinks it from the list; Created indexExists method which checks if the given index is not out of boundaries of an array size; In add method, checked if an element is first in the array therefore assign it as head, otherwise assign it as a tail; plus increase size by one; In add(value, index) method, checked if index is not negative and not bigger than size of the array. Checking the same requirement as add method and assigning head or tail, if the index is somewhere between the elements call addMidNode method, increase size by one; In addAll method, with help of for loop, add all the elements to the array; In get method check if the index exists, than call getNodeIndex method; In set method check if index exists, than adding new element to the array and link new element with neighbours; In remove method, check if index exists and call removeNode method; In remove(object) method, with help of while we go through the whole array and search for the asked object; In size method returning size value;
In isEmpty method returning true if size equals 0;

### DIFF
--- a/src/main/java/core/basesyntax/MyLinkedList.java
+++ b/src/main/java/core/basesyntax/MyLinkedList.java
@@ -3,45 +3,171 @@ package core.basesyntax;
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size;
+    private Node<T> head;
+    private Node<T> tail;
+
+    class Node<T> {
+        private T value;
+        private Node<T> prev;
+        private Node<T> next;
+
+        public Node(T value, Node<T> prev, Node<T> next) {
+            this.value = value;
+            this.prev = prev;
+            this.next = next;
+        }
+    }
+
     @Override
     public void add(T value) {
+        if (size == 0) {
+            addHead(value);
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,166 @@
...
+    private Node<T> tail;
+
     @Override
     public void add(T value) {
+        if (size == 0) {
+            addHead(value);
+        } else {
+            addTail(value);
+        }
+        //size++;
```
aaananas, line 17:
```suggestion
```

aaananas (approved):

PR #802, opened 2022-12-22
All tests successfully passed

taras-zhyravetskyi:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,156 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private static final int INITIAL_LIST_SIZE = 10;
+    private static final int FIRST_INDEX = 1;
```
Gleb-Shevchenko: Do you need this constants?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,156 @@
...
+    private int size;
+    private T[] elementData;
```
Gleb-Shevchenko: LinkedList is not an array inside - just a group of Node objects.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,156 @@
...
+            this.prev = prev;
+            this.item = item;
+            this.next = next;
+        }
+    }
+
+    MyLinkedList() {
+        first = new Node<T>(null, null, null);
+        last = new Node<T>(null, null,null);
+    }
```
Gleb-Shevchenko, line 25: Fields first and last should be fields of class without any declaration. You don't need this constructor.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,156 @@
...
+
     @Override
     public void add(T value) {
+        Node<T> newNode = new Node<T>(last, value, null);
```
Gleb-Shevchenko, line 29: Change last to null. Then you can compare references of newNode to null

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,156 @@
...
+        if (size == 0) {
+            first = newNode;
+            first.prev = null;
+        }
+        last.next = newNode;
+        last = newNode;
+        size++;
```
Gleb-Shevchenko, line 36: Your logic in methode above is generally right. What if you compare not size but first to null?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,156 @@
...
     @Override
     public void add(T value, int index) {
+        if (size == index) {
+            add(value);
+            return;
+        }
+        checkIndex(index);
+        Node<T> currentNode = findNode(index);
+        Node<T> newNode = new Node<T>(currentNode.prev, value, currentNode);
+        if (index == 0) {
```
Gleb-Shevchenko, line 48: You can check not index == 0, but  if previous node == null; Then try to use in this method if-else construction.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,156 @@
...
+            first = newNode;
+            newNode.next.prev = newNode;
+            size++;
```
Gleb-Shevchenko, line 51: size++ should be only at the and of the method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,156 @@
...
+        }
+        newNode.prev.next = newNode;
+        newNode.next.prev = newNode;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
+        for (T value : list) {
+            Node<T> newNode = new Node<T>(last, value, null);
```
Gleb-Shevchenko, line 62: You can call here add() method instead of code inside a for loop

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,156 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void unlink(Node node) {
+        if (size == 1) {
```
Gleb-Shevchenko, line 117: I guess here you can only compare if last and first == null. And check your type parameter in input Node

src/main/java/core/basesyntax/MyLinkedList.java
taras-zhyravetskyi: Deleted

src/main/java/core/basesyntax/MyLinkedList.java
taras-zhyravetskyi: +

src/main/java/core/basesyntax/MyLinkedList.java
taras-zhyravetskyi: Change to null

src/main/java/core/basesyntax/MyLinkedList.java
taras-zhyravetskyi: Done

src/main/java/core/basesyntax/MyLinkedList.java
taras-zhyravetskyi: Leaved one size++ at the end of method

src/main/java/core/basesyntax/MyLinkedList.java
taras-zhyravetskyi: done

Gleb-Shevchenko (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,140 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private static final int LIST_DIVIDER = 2;
```
KostyaNikiforov, line 6: This is not a list, but an int value

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,140 @@
...
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Incorect index: " + index);
+        }
+    }
+
+    private Node<T> findNode(int index) {
+        if (index > size / LIST_DIVIDER) {
```
KostyaNikiforov, line 129: And this is a redundant constant because it is quite obvious without any additional names that we need to divide a size into 2.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,140 @@
...
+        private T item;
+        private Node<T> next;
+        private Node<T> prev;
+
+        Node(Node<T> prev, T item, Node<T> next) {
+            this.prev = prev;
+            this.item = item;
+            this.next = next;
+        }
+    }
```
KostyaNikiforov, line 21: Usually, inner classes are placed at the end of the main one

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):
🔥

PR #816, opened 2022-12-26
LL solution,build success,mvn clean package ok

Mykola-Osolinskyi:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,155 @@
...
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size;
+    private Node<T> first;
+    private Node<T> last;
+
     @Override
     public void add(T value) {
+        final Node<T> l = last;
+        final Node<T> newNode = new Node<>(l, value, null);
```
Gleb-Shevchenko, line 13: Why this fields are final?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,155 @@
...
+                }
+            }
+        } else {
+            for (Node<T> x = first; x != null; x = x.next) {
+                if (object.equals(x.value)) {
+                    unlink(x);
+                    return true;
+                }
+            }
+        }
```
Gleb-Shevchenko, line 84: You wrote here 2 loops, that could be turned into one. So make one loop and in its body write general condition as for Objects.equals. 
x.value== object || x.value!= null && x.value.equals(object)
And then unlink

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,155 @@
...
+            this.value = element;
+            this.next = next;
+            this.prev = prev;
+        }
+    }
+
+    T unlink(Node<T> x) {
+        final T element = x.value;
+        final Node<T> next = x.next;
+        final Node<T> prev = x.prev;
```
Gleb-Shevchenko, line 120: I am not sure fields should be final

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,155 @@
...
+            last = prev;
+        } else {
+            next.prev = prev;
+            x.next = null;
+        }
+
+        x.value = null;
+        size--;
+        return element;
+    }
```
Gleb-Shevchenko, line 139: Guess this method should not return any value. And it misses access modificators

src/main/java/core/basesyntax/MyLinkedList.java
Mykola-Osolinskyi: this fields can be final, or not final
I made them final, because the same fields are final in LL documentation

src/main/java/core/basesyntax/MyLinkedList.java
Mykola-Osolinskyi: this fields can be final, or not final
I made them final, because the same fields are final in LL documentation

src/main/java/core/basesyntax/MyLinkedList.java
Mykola-Osolinskyi: in LL documentation this method return value of unlinked node.

Gleb-Shevchenko (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,149 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size;
+    private Node<T> first;
+    private Node<T> last;
+
     @Override
     public void add(T value) {
+        final Node<T> l = last;
```
KostyaNikiforov, line 12: Can you just use `last` field instead of this local var?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,149 @@
...
+
+        node.value = null;
+        size--;
+        return element;
+    }
+
+    private Node<T> getNodeByIndex(int index) {
+        checkIndex(index);
+        Node<T> node;
+        if (index < (size >> 1)) {
```
KostyaNikiforov, line 138:
```suggestion
        if (index < size >> 1) {
```

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):

PR #831, opened 2023-01-02
created inner class Node<T> and created additional methods:

nkharyshyn:
-findNode to find needed node by index;
-linkLast and linkBefore to add new elements;
-checkIndex to check if index is correct;
-unlink to remove node

### DIFF
--- a/src/main/java/core/basesyntax/MyLinkedList.java
+++ b/src/main/java/core/basesyntax/MyLinkedList.java
@@ -3,45 +3,164 @@ package core.basesyntax;
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    private static class Node<T> {
+        private T element;
+        private Node<T> prev;
+        private Node<T> next;
+
+        public Node(Node<T> prev, T element, Node<T> next) {
+            this.element = element;
+            this.prev = prev;
+            this.next = next;
+        }
+    }
+
     @Override
     public void add(T value) {
+        linkLast(value);
     }
 
     @Override
     public void add(T value, int index) {
+        if (index == size) {
+            linkLast(value);
+        } else {
+            Node<T> node = findNode(index);
+            linkBefore(value, node);
+        }
     }
 
     @Override
     public void addAll(List<T> list) {
+        for (T listElement : list) {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
...
+        } else {
+            while (node != null) {
+                if (node.element != null
+                        && node.element.equals(object)) {
+                    unlink(node);
+                    return true;
+                }
+                node = node.next;
+            }
+        }
```
zachar-chorny, line 81:
```suggestion
            while (node != null) {
                if (node.element == object  || node.element != null
                        && node.element.equals(object)) {
                    unlink(node);
                    return true;
                }
                node = node.next;
            }
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    private static class Node<T> {
```
zachar-chorny, line 10: private class should be after all methods

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
...
+        public Node(Node<T> prev, T element, Node<T> next) {
+            this.element = element;
+            this.prev = prev;
+            this.next = next;
+        }
+    }
+
     @Override
     public void add(T value) {
+        linkLast(value);
```
zachar-chorny, line 24: move logic from linkLast to add

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
...
-        return null;
+        return findNode(index).element;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        T setElement = findNode(index).element;
+        findNode(index).element = value;
+        return setElement;
```
zachar-chorny, line 53: you can call findNode only once here

zachar-chorny (changes requested):

aleksandr-bevza (approved):

PR #833, opened 2023-01-03
Added variables:

Iskamele:
- int size = 0 - to display the actual size of the LinkedList;
- Node<T> head - to identify the first element in the LinkedList;
- private Node<T> tail - for the last element, respectively.

Added public MyLinkedList() constructor to initialize an empty LinkedList.

Added class Node<T>, which stores the passed value, as well as links to the previous and next node. Inside the class, the Node constructor is implemented.

Added the Node<T> getNode(int index) method, which should return the node searched by index. It is implemented in accordance with the recommendation from Common Mistakes. First of all, we declare the variable node.
If index <= size / 2, then we are looking for a node from the head. Otherwise, we are looking from the tail.
In both cases, we loop from start to index, iterating over node.next or node.prev. We return node.

Overridden method void add(T value) from interface MyLinkedListInterface. We declare the node variable and initialize it with the tail. It will serve as temporary storage. We declare the newNode variable, which we initialize using the Node constructor. We pass a new node to the tail.
If there were no nodes in the leaf, then node == null, then we pass newNode to head. Otherwise, we pass newNode to node.next.
We make an increment of size.

Overridden method void add(T value, int index) from interface MyLinkedListInterface. First of all, we check the validity of the index using checkPositionIndex(index). In case we add an element to the head of the List, then index == size, call add(value) and break the method. In all other cases, we make a temporary node for the shifted element and newNode, which we initialize with the constructor. If the previous node of the temporary node != null, then pass newNode to the previous node of the temporary node. Otherwise, we pass newNode to head.
After that, we pass newNode to node.prev.
We make an increment of size.

Overridden method void addAll(List<T> list) from interface MyLinkedListInterface. We loop through the list and pass each element to the add() method.

Overridden method T get(int index) from interface MyLinkedListInterface. We check the validity of the index using checkElementIndex(index). Then we return the element we are looking for using the getNode(index).item method.

Overridden method T set(T value, int index) from interface MyLinkedListInterface. We check the validity of the index using checkElementIndex(index). We declare the currentNode variable and pass the value of the node at the getNode(index) index to it. We declare the variable oldValue, which stores the value of currentNode. After that, the value of currentNode is replaced by the incoming T value. We return oldValue.

Overridden method T remove(int index) from interface MyLinkedListInterface. We check the validity of the index using checkElementIndex(index). We declare the node variable and initialize it with getNode(index). Then we call the unlink() method and pass node.
We return the value of node.item.

Overridden boolean remove(T object) method from MyLinkedListInterface interface. Since we can't determine from which side we can get to the desired element faster, we are forced to loop from the head to the tail of the sheet until we hit null. If node.item != null;
And node.item is equal to the object you are looking for; Or node.item == object, then we pass the node value to the unlink() method. And return true.
If, as a result of the loop, the element is not found, then we return false.

The int size() method from the MyLinkedListInterface interface is overridden, which returns the value of the size variable.

Overridden the boolean isEmpty() method from the MyLinkedListInterface interface, which returns true if size == 0.

Added void checkElementIndex(int index) method.
The method handles cases where negative values are entered, or when the index is equal to or greater than size because the index of elements in a sheet starts at 0. If index < 0 or index >= size, then we throw an IndexOutOfBoundsException error.

Added void checkPositionIndex(int index) method.
The method provides for cases where negative values are entered or if the index exceeds size. They can be equal because methods that use this check can use an index equal to size, such as adding a new element to the end of a sheet. If index < 0 or index > size, then we throw an IndexOutOfBoundsException error.

Added String indexOutOfBoundsMessage(int index) method, which returns a message for IndexOutOfBoundsException, which is thrown by two methods to check the index. This method is implemented with an eye to the original implementation of LinkedList.

Added void unlink(Node<T> node) method.
Its task is to untie the nodes' neighbors, thereby "removing" them from the list. In the first if-else block, if the current node is not head, then we associate its previous neighbor with the next one, otherwise, we pass the value of the next neighbor in the head (because in this case, we remove the element zero by index). In the second if-else block, if the current node is not tail, then we associate the next element with the previous one, otherwise, we pass the value of the previous element to the tail (in case we delete the last element by index). We decrement size.

### DIFF
--- a/src/main/java/core/basesyntax/MyLinkedList.java
+++ b/src/main/java/core/basesyntax/MyLinkedList.java
@@ -3,7 +3,7 @@ package core.basesyntax;
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
-    private int size = 0;
+    private int size;
     private Node<T> head;
     private Node<T> tail;
 
--- a/src/main/java/core/basesyntax/MyLinkedList.java
+++ b/src/main/java/core/basesyntax/MyLinkedList.java
@@ -10,38 +10,10 @@ public class MyLinkedList<T> implements MyLinkedListInterface<T> {
     public MyLinkedList() {
     }
 
-    private static class Node<T> {
-        private T item;
-        private Node<T> next;
-        private Node<T> prev;
-
-        Node(Node<T> prev, T element, Node<T> next) {
-            this.item = element;
-            this.next = next;
-            this.prev = prev;
-        }
-    }
-
-    private Node<T> getNode(int index) {
-        Node<T> node;
-        if (index <= size / 2) {
-            node = head;
-            for (int i = 0; i < index; i++) {
-                node = node.next;
-            }
-        } else {
-            node = tail;
-            for (int i = size - 1; i > index; i--) {
-                node = node.prev;
-            }
-        }
-        return node;
-    }
-
     @Override
     public void add(T value) {
         Node<T> node = tail;
-        Node<T> newNode = new Node<>(node, value, null);
+        Node<T> newNode = new Node<>(node, value);
         tail = newNode;
         if (node == null) {
             head = newNode;
@@ -122,6 +94,23 @@ public class MyLinkedList<T> implements MyLinkedListInterface<T> {
         return size == 0;
     }
 
+    private static class Node<T> {
+        private T item;
+        private Node<T> next;
+        private Node<T> prev;
+
+        Node(Node<T> prev, T element, Node<T> next) {
+            this.item = element;
+            this.next = next;
+            this.prev = prev;
+        }
+
+        Node(Node<T> prev, T element) {
+            this.item = element;
+            this.prev = prev;
+        }
+    }
+
     private void checkElementIndex(int index) {
         if (index < 0 || index >= size) {
             throw new IndexOutOfBoundsException(indexOutOfBoundsMessage(index));
@@ -150,5 +139,21 @@ public class MyLinkedList<T> implements MyLinkedListInterface<T> {
             tail = node.prev;
         }
         size--;
+    }
+
+    private Node<T> getNode(int index) {
+        Node<T> node;
+        if (index <= size / 2) {
+            node = head;
+            for (int i = 0; i < index; i++) {
+                node = node.next;
+            }
+        } else {
+            node = tail;
+            for (int i = size - 1; i > index; i--) {
+                node = node.prev;
+            }
+        }
+        return node;
     }
 }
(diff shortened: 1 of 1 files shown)


karma-o (commented):
Это круто что ты расписал весь функционал и логику, но мне кажется что ты потратил это СЛИШКОМ много времени)
Особенно на описание буквально каждой строчки кода, это уже лишнее.
Если хочешь, так можно делать но я думаю пара предложений на 1 метод достаточно.
А вообще лучший код это тот который не нуждается в комментариях :wink:

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,152 @@
...
+                node = node.prev;
+            }
+        }
+        return node;
+    }
+
     @Override
     public void add(T value) {
+        Node<T> node = tail;
+        Node<T> newNode = new Node<>(node, value, null);
```
karma-o, line 44:
```suggestion
        Node<T> newNode = new Node<>(node, value);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,152 @@
...
+        private Node<T> prev;
+
+        Node(Node<T> prev, T element, Node<T> next) {
+            this.item = element;
+            this.next = next;
+            this.prev = prev;
+        }
+    }
+
+    private Node<T> getNode(int index) {
```
karma-o, line 25: move private methods and nested classes to the bottom

karma-o (changes requested):

Iskamele:
> Это круто что ты расписал весь функционал и логику, но мне кажется что ты потратил это СЛИШКОМ много времени) Особенно на описание буквально каждой строчки кода, это уже лишнее. Если хочешь, так можно делать но я думаю пара предложений на 1 метод достаточно. А вообще лучший код это тот который не нуждается в комментариях 😉

Что просто – написал сразу на англ, что посложнее – написал на ру и перевёл, проверил.
С одной стороны много времени, согласен, с другой же стороны лучше понимаешь код, когда описываешь его работу step by step. Это особенно заметно, когда копируешь части кода из стандартной библиотеки, вроде линкед листа)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,157 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size = 0;
```
aaananas, line 6:
```suggestion
    private int size;
```
class variables have default values

aaananas (approved):

PR #835, opened 2023-01-04
jv-linked-list first solution

kcasPilot:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,198 @@
...
+            return element;
+        }
+
+        public Node<E> getPrev() {
+            return prev;
+        }
+
+        public Node<E> getNext() {
+            return next;
+        }
```
aaananas, line 196:
```suggestion
    private static class Node<E> {
        private E element;
        private Node<E> prev;
        private Node<E> next;

        public Node(E element, Node<E> prev, Node<E> next) {
            this.element = element;
            this.prev = prev;
            this.next = next;
        }
```
no need for getters and setters

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,198 @@
...
+import java.util.NoSuchElementException;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> firstNode;
+    private Node<T> lastNode;
+    private int size;
+
     @Override
     public void add(T value) {
+        putNode(value, size == 0 ? "first" : "top");
```
aaananas, line 13:
```suggestion
        putNode(value);
```
it is not ok to use string value here
just add an `if (size == 0)` inside the method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,198 @@
...
+        }
+        if (node.equals(lastNode) && size > 1) {
+            node.getPrev().setNext(null);
+            lastNode = node.getPrev();
+        }
+        if (firstNode.equals(lastNode)) {
+            firstNode = null;
+            lastNode = null;
+        }
+        size--;
```
aaananas, line 160: use == to check if node **is the same** (reference is the same) object or is equal

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,198 @@
...
+        node = firstNode;
+        nodeIndex = 0;
+        while (nodeIndex < index) {
+            node = node.getNext();
+            nodeIndex++;
+        }
+        return node;
+    }
+
+    private void putNodeMiddle(T value, Node<T> operationNode) {
```
aaananas, line 115:
```suggestion
    private void putNodeBefore(T value, Node<T> node) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,198 @@
...
+        Node<T> node = getNodeByIndex(index);
+        T element = node.getElement();
+        node.setElement(value);
+        return element;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index, size - 1);
```
aaananas, line 53:
```suggestion
        checkIndex(index);
```
why -1? 
it is hard to read, in fact you dont need to pass another value in checkIndex() method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,198 @@
...
     }
 
     @Override
     public void add(T value, int index) {
+        checkIndex(index, size);
+        if (index > 0 && index < size) {
+            putNodeMiddle(value, getNodeByIndex(index));
+            return;
+        }
+        putNode(value, index == 0 ? size == 0 ? "first" : "bottom" : "top");
```
aaananas, line 23:
```suggestion
        if (index == size) {
            add(value);
            return;
        }
        checkIndex(index);
        putNodeBefore(value, getNodeByIndex(index));
```

aaananas (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
kcasPilot: Ok. Thanks )

nacenik (approved):
Good job!

PR #841, opened 2023-01-14
added MyLinkedList implementation

JRuiy:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
+            for (Node<T> i = head; i != null; i = i.next) {
+                if (i.item == null) {
+                    deleteLink(i);
+                    return true;
+                }
+            }
+        } else {
+            for (Node<T> i = head; i != null; i = i.next) {
+                if (i.item != null && i.item.equals(object)) {
+                    deleteLink(i);
```
karma-o, line 100: usually `i` stands for `index`, but in this case, it is not an index, so lets name it with something else

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void checkIndex(int index) {
+        if (!(index >= 0 && index < size)) {
+            throw new IndexOutOfBoundsException("Incorrect index");
```
karma-o, line 120:
```suggestion
        if (index < 0 && index >= size) {
            throw new IndexOutOfBoundsException("index: " + index + " is invalid");
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
     public boolean remove(T object) {
+        if (object == null) {
+            for (Node<T> i = head; i != null; i = i.next) {
+                if (i.item == null) {
+                    deleteLink(i);
+                    return true;
+                }
+            }
+        } else {
+            for (Node<T> i = head; i != null; i = i.next) {
```
karma-o, line 98: make it an `if` without `else` with a `object == i.item || object != null && object.equals(i.item)` condition

karma-o (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,146 @@
...
+        return size == 0;
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("index: " + index + " is invalid");
+        }
+    }
+
+    private Node<T> searchNode(int index) {
```
aaananas, line 100:
```suggestion
    private Node<T> getNode(int index) {
```

aaananas (approved):

PR #842, opened 2023-01-14
feat: jv-linked-list #comment linked list implementation

sania517:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,153 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private static class Node<T> {
```
aleksandr-bevza, line 6: Move this private class to the bottom of your MyLinkedList class

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,153 @@
...
+    }
+
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
     @Override
     public void add(T value) {
+        if (tail == null) {
+            this.head = this.tail = new Node(value, null, null);
```
aleksandr-bevza, line 25: No need to write 'this' here. They are 'this' by default
```suggestion
            head = tail = new Node(value, null, null);
```

aleksandr-bevza, line 25: You can avoid warnings from Idea (Raw use of parameterized class 'Node' ) if you add '<>' here. Fix it in other places where you have yellow warnings from Idea
```suggestion
            head = tail = new Node<>(value, null, null);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,153 @@
...
+        if (index > size || index < 0) {
+            throw new IndexOutOfBoundsException(
+              "index: " + index + " is more than list size: " + size + "."
+            );
+        }
+
+        if (index == size || size == 0) {
+            add(value);
+            return;
+        }
```
aleksandr-bevza, line 46:
```suggestion
        if (index == size) {
            add(value);
            return;
        }
        checkIndex(index);     
```

aleksandr-bevza (changes requested):

aleksandr-bevza (approved):

PR #849, opened 2023-01-16
My linked list solution

mrGvvD:
error was fixed


Ivan95kos (changes requested):
Знімок екрана 2023-01-16 о 21 29 34

You don't need to change pom.xml in this task

PR #850, opened 2023-01-16
jv-linked-list solution

mrGvvD:
Here were overrided next methods: add(),addAll(), get(), set(), remove(), size(), isEmpty() & created a few new methods.

### DIFF
--- a/src/main/java/core/basesyntax/MyLinkedList.java
+++ b/src/main/java/core/basesyntax/MyLinkedList.java
@@ -3,45 +3,131 @@ package core.basesyntax;
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size;
+    private Node<T> head;
+    private Node<T> tail;
+
     @Override
     public void add(T value) {
+        Node<T> currentTailNode = tail;
+        Node<T> newNode = new Node<>(currentTailNode, value, null);
+        tail = newNode;
+        if (size == 0) {
+            head = newNode;
+        } else {
+            currentTailNode.next = newNode;
+        }
+        size++;
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,131 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size;
+    private Node<T> head;
+    private Node<T> tail;
+
     @Override
     public void add(T value) {
+        Node<T> currentTailNode = tail;
```
nacenik, line 12: why do you need this var? You can use tail.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,131 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void checkIndexCorrectness(int index) {
```
nacenik, line 92:
```suggestion
    private void checkIndex(int index) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,131 @@
...
+        Node<T> currentNode = getNode(index);
+        T result = currentNode.value;
+        currentNode.value = value;
+        return result;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndexCorrectness(index);
```
nacenik, line 63: this row is useless, cause you do the same check inside getNode(index) method

nacenik (approved):
Good Job!

PR #853, opened 2023-01-17
complete(all) : finish task

urlo0k:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,178 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+
+    private Node<T> head;
+    private Node<T> tail;
+    private int size = 0;
```
aleksandr-bevza, line 9: No need to set it 0. It is 0 by default
```suggestion
    private int size;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,178 @@
...
+
+    public MyLinkedList() {
+        head = null;
+        tail = null;
+    }
+
+    public static class Node<T> {
```
aleksandr-bevza, line 16: This inner class is only for your MyLinkedList and should not be accessible from outside. Make it private.
Move this inner private class to the bottom of your MyLinkedList class
```suggestion
    private static class Node<T> {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,178 @@
...
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+
+    private Node<T> head;
+    private Node<T> tail;
+    private int size = 0;
+
+    public MyLinkedList() {
+        head = null;
+        tail = null;
+    }
```
aleksandr-bevza, line 14: No need to set head and tail null. They are null by default.
No need to create this constructor. It will be created by default because it is the only constructor in this class.
```suggestion
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,178 @@
...
+            tail.next = null;
+        } else {
+            node.prev.next = node.next;
+            node.next.prev = node.prev;
+        }
+        size--;
+    }
+
+    private void checkIndexExeption(int index) {
+        if (index >= size && index != 0 || index < 0) {
```
aleksandr-bevza, line 175:
```suggestion
        if (index < 0 || index >= size) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,178 @@
...
+            tail = node.prev;
+            tail.next = null;
+        } else {
+            node.prev.next = node.next;
+            node.next.prev = node.prev;
+        }
+        size--;
+    }
+
+    private void checkIndexExeption(int index) {
```
aleksandr-bevza, line 174:
```suggestion
    private void checkIndex(int index) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,178 @@
...
     @Override
     public void add(T value, int index) {
+        if (index < 0 && index > size) {
+            throw new IndexOutOfBoundsException("No such index exception: index "
+                    + index + " for size " + size);
+        }
+        if (index == size) {
+            add(value);
+            return;
+        }
```
aleksandr-bevza, line 54:
```suggestion
        if (index == size) {
            add(value);
            return;
        }
        checkIndexExeption(index);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,178 @@
...
+            if (object == null) {
+                if (getNode(i).value == null) {
+                    remove(i);
+                    return true;
+                }
+            } else if (object.equals(getNode(i).value)) {
+                remove(i);
+                return true;
+            }
+        }
```
aleksandr-bevza, line 117: Don't forget about performance. Each 'getNode' method calls loop inside. You can remove node using only 1 loop
```suggestion
        for (Node<T> node = head; node != null; node = node.next) {
            if (node.value == object || object != null && object.equals(node.value)) {
                unlink(node);
                return true;
            }
        }
```

aleksandr-bevza (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+
```
okuzan, line 6: extra line

okuzan (approved):
Good job!

PR #856, opened 2023-01-20
added to the class MyLinkedList:

nikolasdymchenko7:
- override the following methods: add(), addAll(), get(), set(), remove(), size(), isEmpty()
- static class Node
- additional methods: unlinkItem(), linkBeforeItem(), linkedLastItem(), findNodeByIndex(), checkElementIndex(), checkPositionIndex(), isElementIndex(), isPositionIndex()

### DIFF
--- a/src/main/java/core/basesyntax/MyLinkedList.java
+++ b/src/main/java/core/basesyntax/MyLinkedList.java
@@ -4,7 +4,7 @@ import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
     private int size;
-    private Node<T> last = null;
+    private Node<T> last;
     private Node<T> first;
 
     @Override
@@ -52,19 +52,10 @@ public class MyLinkedList<T> implements MyLinkedListInterface<T> {
 
     @Override
     public boolean remove(T object) {
-        if (object == null) {
-            for (Node<T> node = first; node != null; node = node.next) {
-                if (node.item == null) {
-                    unlinkItem(node);
-                    return true;
-                }
-            }
-        } else {
-            for (Node<T> node = first; node != null; node = node.next) {
-                if (object.equals(node.item)) {
-                    unlinkItem(node);
-                    return true;
-                }
+        for (Node<T> node = first; node != null; node = node.next) {
+            if (node.item == object || (object != null && object.equals(node.item))) {
+                unlinkItem(node);
+                return true;
             }
         }
         return false;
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,177 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size;
+    private Node<T> last = null;
```
```suggestion
    private Node<T> last;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,177 @@
...
+                    return true;
+                }
+            }
+        } else {
+            for (Node<T> node = first; node != null; node = node.next) {
+                if (object.equals(node.item)) {
+                    unlinkItem(node);
+                    return true;
+                }
+            }
```
aaananas, line 68: lets combine if conditions, because the body is identical

aaananas (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,168 @@
...
+            throw new IndexOutOfBoundsException("Index " + index + " is out of range!");
+        }
+    }
+
+    private boolean isElementIndex(int index) {
+        return index >= 0 && index < size;
+    }
+
+    private boolean isPositionIndex(int index) {
+        return index >= 0 && index <= size;
```
karma-o, line 168: We can also remove the boolean returning methods and inline them, since we are not using them anywhere else
and try not to negate the statements, and form them so that they would work without `!`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,168 @@
...
+        foundNode.prev = newNode;
+        if (prevNode == null) {
+            first = newNode;
+        } else {
+            prevNode.next = newNode;
+        }
+        size++;
+    }
+
+    private void linkedLastItem(T value) {
```
karma-o, line 122: lets remove this method and use just `add(T value)` instead of this one

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,168 @@
...
+            next.prev = prev;
+            foundNode.next = null;
+        }
+
+        foundNode.item = null;
+        size--;
+        return removeElement;
+    }
+
+    private void linkBeforeItem(T item, Node<T> foundNode) {
```
karma-o, line 110: Same as the `linkedLastItem()` method, it is used only in 1 place and it is `add(T value, int index)`, so I don't see any reason to move almost all of its functionality to a private method

karma-o (changes requested):

karma-o (approved):
:fire:

PR #858, opened 2023-01-24
Good afternoon, dear mentor. There are bugs in jv-linked-list code. I…

911Mik:
… do not understand the reason for 01/24/23 Question: how to fix it?

### DIFF
--- a/src/main/java/core/basesyntax/MyLinkedList.java
+++ b/src/main/java/core/basesyntax/MyLinkedList.java
@@ -3,45 +3,220 @@ package core.basesyntax;
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node tail;
+    private Node head;
+    private int size;
+
+    public Node getHead() {
+        return head;
+    }
+
+    public Node getTail() {
+        return tail;
+    }
+
+    public void setHead(Node head) {
+        this.head = head;
+    }
+
+    public void setTail(Node tail) {
+        this.tail = tail;
+    }
+
+    private class Node<T> {
+        private Node left;
+        private Node right;
+        private T value;
+
+        public Node(Node left, Node right, T value) {
+            this.left = left;
+            this.right = right;
+            this.value = value;
+        }
+
+        public Node getLeft() {
+            return left;
+        }
+
+        public Node getRight() {
+            return right;
+        }
+
+        public T getValue() {
+            return value;
+        }
+
+        public void setLeft(Node left) {
+            this.left = left;
+        }
+
+        public void setRight(Node right) {
+            this.right = right;
+        }
+
+        public void setValue(T value) {
+            this.value = value;
+        }
+    }
+
+    public void setSize(int size) {
+        this.size = size;
+    }
+
+    public MyLinkedList() {
+        head = null;
+        tail = null;
+        size = 0;
+    }
+
+    private void unlink(Node node) {
+        // 1 - node - 3
+        if (node.left == null) {
+            head = node.right;
+        } else {
+            node.left.right = node.right;
+        }
+        if (node.right == null) {
+            tail = node.left;
+        } else {
+            node.right.left = node.left;
+        }
+    }
+
+    private Node getNode(int index) {
+        int count = 0;
+        Node pointer = head;
+        while (pointer != null) {
+            if (count == index) {
+                return pointer;
+            }
+            pointer = pointer.right;
+            count++;
+        }
+        return null;
+    }
+
+    private void testIndex(int index) {
+        if (index >= size() || index < 0) {
+            //  if(index >= size || index < 0) {
+            throw new java.lang.IndexOutOfBoundsException();
+        }
+    }
+
     @Override
     public void add(T value) {
+        Node newNoge = new Node(tail, null, value);
+        if (head == null) {
+            head = newNoge;
+        }
+        if (tail != null) {
+            tail.right = newNoge;
+        }
+        tail = newNoge;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index == size()) {
+            add(value);
+            return;
+        }
+        testIndex(index);
+        // int count = 0;
+        Node pointer = getNode(index); // 1 - newNode - pointer
+        Node leftNode = null;
+        if (pointer != null) {
+            leftNode = pointer.left;
+        }
+        Node newNode = new Node(leftNode, pointer, value);
+        if (leftNode != null) {
+            leftNode.right = newNode;
+        }
+        if (pointer != null) {
+            pointer.left = newNode;
+        }
     }
 
     @Override
     public void addAll(List<T> list) {
+        for (T element : list) {
+            add(element);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        testIndex(index);
+        Node pointer = getNode(index);
+        if (pointer == null) {
+            return null;
+        } else {
+            return (T) pointer.getValue();
+        }
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        testIndex(index);
+        int count = 0;
+        Node<T> pointer = head;
+        T oldValue = null;
+        while (pointer != null) {
+            if (index == count) {
+                oldValue = pointer.getValue();
+                pointer.setValue(value);
+                break;
+            } else {
+                pointer = pointer.right;
+                count++;
+            }
+        }
+        return oldValue;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        testIndex(index);
+        T toReturn = null;
+        Node<T> pointer = getNode(index);
+        if (pointer != null) {
+            T result = pointer.getValue();
+            unlink(pointer);
+            size--;
+            return result;
+        }
+        return toReturn;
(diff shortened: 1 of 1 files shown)


okuzan (changes requested):
Delete the comment block in the size() method, indentation is wrong there: you can use Ctrl Alt L in Intellij to auto format. Also put constructor just after fields, methods should go after constructor

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,175 @@
...
     @Override
     public int size() {
-        return 0;
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return head == null;
```
okuzan, line 175:
```suggestion
        return size == 0;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,175 @@
...
     @Override
     public T remove(int index) {
-        return null;
+        testIndex(index);
+        T toReturn = null;
+        Node<T> pointer = getNode(index);
+        if (pointer != null) {
+            T result = pointer.getValue();
+            unlink(pointer);
+            size--;
```
okuzan, line 147: better include size-- to unlink method instead of repeatedly calling them together

okuzan (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,172 @@
...
+    private int size;
+
+    public MyLinkedList() {
+        head = null;
+        tail = null;
+    }
+
     @Override
     public void add(T value) {
+        Node<T> newNoge = new Node<>(tail, null, value);
```
andriy-shulepa, line 17:
```suggestion
        Node<T> newNode = new Node<>(tail, null, value);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,172 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private static class Node<T> {
```
andriy-shulepa, line 122: place nested class at the bottom of the file

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,172 @@
...
+            }
+        }
+        return oldValue;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        testIndex(index);
+        T toReturn = null;
```
andriy-shulepa, line 88: this variable is always null, looks like useless

andriy-shulepa (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,172 @@
...
+                count++;
+            }
+        }
+        return oldValue;
     }
 
     @Override
     public T remove(int index) {
+        testIndex(index);
+        //   T toReturn = null;
```
nacenik, line 88: don't commit comments in source code
```suggestion
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,172 @@
...
+        } else {
+            return pointer.getValue();
+        }
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        testIndex(index);
+        int count = 0;
```
nacenik, line 69: this is a redundant variable. Delete it 
```suggestion
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,172 @@
...
+        while (pointer != null) {
+            if (index == count) {
+                oldValue = pointer.getValue();
+                pointer.setValue(value);
+                break;
+            } else {
+                pointer = pointer.next;
+                count++;
+            }
+        }
```
nacenik, line 81: you have the same logic inside getNode(int index) method
```suggestion
        Node<T> pointer = getNode(index);
        T oldValue = null;
        if (pointer != null) {
           oldValue = pointer.value;
           pointer.value = value;
        }
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,172 @@
...
+            if (count == index) {
+                return pointer;
+            }
+            pointer = pointer.next;
+            count++;
+        }
+        return null;
+    }
+
+    private void testIndex(int index) {
```
nacenik, line 149: test prefix uses in test methods often, use check in regular methods
```suggestion
    private void checkIndex(int index) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,172 @@
...
+            this.value = value;
+        }
+
+        public T getValue() {
+            return value;
+        }
+
+        public void setValue(T value) {
+            this.value = value;
+        }
```
nacenik, line 172: you have acsess to nasted class fields directly, you don't need this methods.

nacenik (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
okuzan: not fixed

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,157 @@
...
+        } else {
+            node.next.prev = node.prev;
+        }
+        size--;
+    }
+
+    private Node<T> getNode(int index) {
+        int count = 0;
+        Node<T> pointer = head;
+        while (pointer != null) {
```
okuzan, line 132: use for loop instead of while with the counter. Also, check which path is shorter - from the head or from the tail, and iterate from there, this will boost your performance

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,157 @@
...
+        } else {
+            return pointer.value;
+        }
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        checkIndex(index);
+        Node<T> pointer = getNode(index);
```
okuzan, line 69: check your index in getNode() instead of repeatedly calling them together

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,157 @@
...
     @Override
     public T get(int index) {
-        return null;
+        checkIndex(index);
+        Node<T> pointer = getNode(index);
+        if (pointer == null) {
+            return null;
+        } else {
+            return pointer.value;
+        }
```
okuzan, line 63: checkIndex will be in getNode, and it will throw exception if index is invalid, so this code will be as simple as return getNode(index).value;

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,157 @@
...
+            Node<T> newNode = new Node<>(null, head, value);
+            head.prev = newNode;
+            head = newNode;
+        } else {
+            Node<T> currentNode = getNode(index);
+            Node<T> newNode = new Node<>(currentNode.prev, currentNode, value);
+            currentNode.prev.next = newNode;
+            currentNode.prev = newNode;
+        }
+        size++;
```
okuzan, line 45: this can be rewritten simpler if your getNode always either returns the node or throws the exception

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,157 @@
...
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> tail;
+    private Node<T> head;
+    private int size;
+
+    public MyLinkedList() {
+        head = null;
+        tail = null;
+    }
```
okuzan, line 13: these are their default values, so this constructor is not needed, the default one is always generated automatically

okuzan (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
okuzan: not fixed

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,153 @@
...
+        for (T element : list) {
+            add(element);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        Node<T> pointer = getNode(index);
+        if (pointer == null) {
```
okuzan, line 53: pointer can't be null. If getNode can't find the node (index is out of bounds, in all other cases node should be found) exception will be thrown, just return getNode(index).value;

okuzan (changes requested):
Good job! One amend needed

okuzan (approved):
You nailed it :) Great job

PR #869, opened 2023-02-07
MyLinkedList is realized

YaroslavKostenko89:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,139 @@
 package core.basesyntax;
 
 import java.util.List;
+import java.util.Objects;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size = 0;
```
nacenik, line 7: 0 is a value by default
```suggestion
    private int size;
```

nacenik (changes requested):

kozhukhovsky (approved):

PR #882, opened 2023-02-16
Class MyLinkedList was implements

GalubaGSV:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,177 @@
...
+        return current.item;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        T temp;
+        checkIndex(index);
+        Node<T> current = getNode(index);
+        temp = current.item;
```
outref, line 96:
```suggestion
        Node<T> current = getNode(index);
        T temp = current.item;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,177 @@
...
+        private Node<T> prev;
+
+        Node(MyLinkedList.Node<T> prev, T element, Node<T> next) {
+            this.item = element;
+            this.next = next;
+            this.prev = prev;
+        }
+    }
+
+    public void removeFirst() {
```
outref, line 22: you can remove nodes without using additional removeFirst or removeLast methods.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,177 @@
...
+            tail = temp;
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Incorrect index " + index);
+        }
```
outref, line 56: you have checkIndex() method for that.

outref (changes requested):
great solution) but I suggest looking at original java.util.LinkedList and taking some ideas from there)

src/main/java/core/basesyntax/MyLinkedList.java
GalubaGSV: here is a slightly different condition

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
...
+        if ((current.next == null && object != null)
+                || (object == null && current.equals(tail) && tail.item != null)) {
+            return false;
+        }
+        if (current.equals(tail)) {
+            unlink(tail);
+            return true;
+        }
+        unlink(current);
+        return true;
```
outref, line 111:
```suggestion
        Node<T> current = head;
        while (current.next != null) {
            //check if  current is the right node, if yes -> unlink and return true
            current = current.next;
        }
        return false;
```
I think you should simplify it like this

src/main/java/core/basesyntax/MyLinkedList.java
outref: yes, but you can put checkIndex() after if(index == size){...}

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
...
+            this.item = element;
+            this.next = next;
+            this.prev = prev;
+        }
+    }
+
     @Override
     public void add(T value) {
+        final Node<T> l = tail;
+        final Node<T> temp = new Node<>(l, value, null);
```
outref, line 25:
```suggestion
        final Node<T> temp = new Node<>(tail, value, null);
```

outref (changes requested):

outref (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,151 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    private static class Node<T> {
```
karma-o, line 10: private classes should be at the bottom

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,151 @@
...
     }
 
     @Override
     public void add(T value, int index) {
+        if (index == size) {
+            add(value);
+            return;
+        } else {
+            checkIndex(index);
+        }
```
karma-o, line 43:
```suggestion
        checkIndex(index);
        if (index == size) {
            add(value);
            return;
        }
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,151 @@
...
+        Node<T> current = getNode(index);
+        unlink(current);
+        return current.item;
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> current = head;
+        while (current != null) {
+            if ((object == current.item) || (object != null && object.equals(current.item))) {
```
karma-o, line 92: redundant braces

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,151 @@
...
+        } else {
+            tail.next = temp;
+            temp.prev = tail;
+            tail = temp;
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
```
karma-o, line 37: would be nice to have a bit less `else`s and maybe more `return`s or `break`s?

karma-o (changes requested):
LGTM, just some minor changes

karma-o (approved):

PR #892, opened 2023-02-19
Custom LinkedList done

IlkoGood:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,150 @@
...
+        for (T value : list) {
+            add(value);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        Node<T> current = getNode(index);
+        return current.value;
```
imhatep, line 60: we can simplify it return getNode(index).value;

imhatep (approved):
good job!

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,150 @@
...
+            currentNode.next.prev = currentNode.prev;
+        }
+        if (currentNode.prev == null) {
+            head = currentNode.next;
+        } else {
+            currentNode.prev.next = currentNode.next;
+        }
+    }
+
+    private boolean customEquals(Object a, Object b) {
```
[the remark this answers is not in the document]
IlkoGood: Good point, thanks

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,149 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    private static class Node<T> {
```
karma-o, line 10: private classes should be at the bottom along with methods

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,149 @@
...
     @Override
     public void add(T value, int index) {
+        if (index == size) {
+            add(value);
+        } else if (index == 0) {
+            head.prev = new Node<>(null, value, head);
+            head = head.prev;
+            size++;
+        } else {
+            Node<T> current = new Node<>(getNode(index).prev, value, getNode(index));
```
karma-o, line 43: do not call getNode twice, for the neighboring nodes

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,149 @@
...
+        size--;
+        return removedValue;
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> current = head;
+        for (int i = 0; i < size; i++) {
+            if (isEqual(current.value, object)) {
+                remove(i);
```
karma-o, line 87: you will iterate twice if you will use `remove(i)` here, lets change it to this instead
```suggestion
                unlink(current);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,149 @@
...
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void makeEmpty() {
+        head = null;
+        tail = null;
+        size = 0;
+    }
```
karma-o, line 109: maybe we should have it called in the `unlink()` rather than `remove()` with some size check

karma-o (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
IlkoGood: oh. ok

src/main/java/core/basesyntax/MyLinkedList.java
IlkoGood: So It's better to create new reference and work with it? Got it!

src/main/java/core/basesyntax/MyLinkedList.java
IlkoGood: hmm, that changes logic a little bit, I'll think how to improve

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,150 @@
...
-        return null;
+        return getNode(index).value;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        T replacedValue = getNode(index).value;
+        getNode(index).value = value;
+        return replacedValue;
```
Andriy-dz, line 55: you are searching twice for a node with the same index
```suggestion
        Node<T> node = getNode(index);
        T replacedValue = node.value;
        node.value = value;
        return replacedValue;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,150 @@
...
+            tail = currentNode.prev;
+            currentNode.prev.next = null;
+        } else {
+            currentNode.next.prev = currentNode.prev;
+        }
+        if (currentNode.prev == null) {
+            head = currentNode.next;
+        } else {
+            currentNode.prev.next = currentNode.next;
+        }
```
Andriy-dz, line 139:
```suggestion
        if (currentNode.next == null) {
            tail = currentNode.prev;
            currentNode.prev.next = null;
        } else if (currentNode.prev == null) {
            head = currentNode.next;
        } else {
            currentNode.next.prev = currentNode.prev;
            currentNode.prev.next = currentNode.next;
        }
```

Andriy-dz (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
IlkoGood: Damn... Why is it so obvious and so hard to see at the same time))

Andriy-dz (approved):
Good Job!!

PR #894, opened 2023-02-20
my-LinkedList solution

imhatep:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,145 @@
...
     public void addAll(List<T> list) {
+        for (T t : list) {
+            add(t);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index != size) {
```
IlkoGood, line 76: If you simplify add by index, divide index searching in half (head to tail and tail to head)
You will be able to rewrite checkIndex with index >= size.
And that would eliminate this check.
I hope my explanation is not confusing, if it is - we'll better pass it to mentors after you complete changes in add() and getNodeByIndex().

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,145 @@
...
+        }
+        throw new IndexOutOfBoundsException("You can't Get value from this index: " + index);
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        if (index != size) {
+            T element = getNodeByIndex(index).item;
+            Node<T> node = getNodeByIndex(index);
```
IlkoGood, line 86: We can call getNodeByIndex(index) once, as it will create only reference to the object and not a deep copy.
So we can manipulate with it's fields as with Node instance we want.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,145 @@
...
+        if (index != size && index >= 0) {
+            T removedElement;
+            if (index == 0) {
+                removedElement = first.item;
+                first = first.next;
+                if (first == null) {
+                    last = null;
+                }
+            } else {
+                Node<T> prev = getNodeByIndex(index - 1);
```
IlkoGood, line 104: According to CM we should create unlink() method for this.

IlkoGood (changes requested):
Many improvements are needed.

IlkoGood (approved):
Nice!

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,143 @@
...
+        Node<T> currentNode = getNodeByIndex(index);
+        T removedValue = currentNode.item;
+        if (size == 1) {
+            head = null;
+            tail = null;
+            size = 0;
+            return removedValue;
+        }
+        unlink(currentNode);
+        size--;
```
karma-o, line 97: if you remove the last element you will have a `size = -1` because of a the 0 assignment

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,143 @@
...
     public T remove(int index) {
-        return null;
+        Node<T> currentNode = getNodeByIndex(index);
+        T removedValue = currentNode.item;
+        if (size == 1) {
+            head = null;
+            tail = null;
+            size = 0;
+            return removedValue;
+        }
```
karma-o, line 95: this part looks more like an `unlink()` method logic

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,143 @@
...
     @Override
     public int size() {
-        return 0;
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size() == 0;
```
karma-o, line 136:
```suggestion
        return size == 0;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,143 @@
...
     }
+
+    private void checkIndex(int index) {
+        if (index >= size || index < 0) {
+            throw new IndexOutOfBoundsException("There is no element on this index: " + index);
```
karma-o, line 141: out of bounds and no element by index are not the same thing
```suggestion
            throw new IndexOutOfBoundsException("Invalid index: " + index);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,143 @@
...
+        }
+    }
+
```
karma-o, line 144: redundant empty line

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,143 @@
...
+                    || current.item != null && current.item.equals(object)) {
+                remove(i);
+                return true;
+            }
+            current = current.next;
+        }
         return false;
     }
 
+    private void unlink(Node<T> x) {
```
karma-o, line 115: private methods should be after public ones

karma-o (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,145 @@
...
+        if (index == size) {
+            add(value);
+        } else if (index == 0) {
+            head.prev = new Node<>(null, value, head);
+            head = head.prev;
+            size++;
+        } else {
+            Node<T> prevNode = getNodeByIndex(index).prev;
+            Node<T> nextNode = getNodeByIndex(index);
+            Node<T> current = new Node<>(prevNode, value, nextNode);
```
Andriy-dz, line 34:
```suggestion
            Node<T> node = getNodeByIndex(index);
            Node<T> current = new Node<>(node.prev, value, node );
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,145 @@
...
     public T get(int index) {
-        return null;
+        return getNodeByIndex(index).item;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        T element = getNodeByIndex(index).item;
+        getNodeByIndex(index).item = value;
```
Andriy-dz, line 56:
```suggestion 
        Node<T> node = getNodeByIndex(index)
        T element = node.item;
        node.item = value;
```

Andriy-dz: you are searching twice for a node with the same index

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,145 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void unlink(Node<T> x) {
```
Andriy-dz, line 95:
```suggestion
    private void unlink(Node<T> node) {
```

Andriy-dz (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,145 @@
...
+            node.next.prev = node.prev;
+        }
+        if (node.prev == null) {
+            head = node.next;
+            if (node.next != null) {
+                node.next.prev = null;
+            }
+        } else {
+            node.prev.next = node.next;
+        }
```
petro-ushchuk, line 111: Quite complicated, you can unlink by using only 2 `if-else` blocks without inner `if`s. Also, `size` adjustment is belongs to `unlink` method, since when you `unlink` your `size` changes immediately.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,145 @@
...
 
     @Override
     public T remove(int index) {
-        return null;
+        Node<T> currentNode = getNodeByIndex(index);
+        T removedValue = currentNode.item;
+        unlink(currentNode);
+        if (size > 0) {
+            size--;
+        }
```
petro-ushchuk, line 67: `size` will never be less than `0`, since when it `0 ` method `checkIndex` will throw an exception. Remove this condition.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,145 @@
...
+        return removedValue;
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> current = head;
+        for (int i = 0; i < size; i++) {
+            if (current.item == null && object == null
+                    || current.item != null && current.item.equals(object)) {
```
petro-ushchuk, line 76: You can simplify it by checking by reference instead of checking for null or non-null.

petro-ushchuk (changes requested):

petro-ushchuk (approved):
Good work 🥇

PR #896, opened 2023-02-21
implemented LinkedList realization with separated logic methods

Skrudj:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,163 @@
...
+                sought = sought.prev;
+            }
+            return sought;
+        }
+    }
+
+    private Node<T> getNode(T value) {
+        Node<T> node = firstNode;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(node.item, value)) {
```
Nikitos787, line 132: we shouldn't use class Object.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,163 @@
...
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private Node<T> getNode(int index) {
+        if (index < 0) {
+            throw new IndexOutOfBoundsException("Can't get element by negative index");
+        } else if (index >= size) {
+            throw new IndexOutOfBoundsException("No such element exists");
```
Nikitos787, line 113: You can use your method "checkIndex"

Nikitos787 (changes requested):
I think my advices is fair

src/main/java/core/basesyntax/MyLinkedList.java
Skrudj: Oh. Forgot this moment

src/main/java/core/basesyntax/MyLinkedList.java
Skrudj: but it's class Objects, not Object

Nikitos787: Objects* . It's written in Common mistakes )

Nikitos787 (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,160 @@
...
+        checkIndex(index);
+        if (index == size) {
+            add(value);
+        } else if (index == 0) {
+            Node<T> newNode = new Node<>(null, value, firstNode);
+            firstNode.prev = newNode;
+            firstNode = newNode;
+            size++;
+        } else {
+            Node<T> afterNode = getNode(index);
```
maxline, line 45: This piece of code is redundant

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,160 @@
...
     public T get(int index) {
-        return null;
+        return getNode(index).item;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        checkIndex(index);
+        T oldValue = getNode(index).item;
```
maxline, line 73: Let's introduce a variable for `getNode(index)`

src/main/java/core/basesyntax/MyLinkedList.java
maxline: I think we can put 
` if (index >= size) {
            throw new IndexOutOfBoundsException("No such element exists");` 
 inside of  `checkIndex(index)`.
 If we need to differentiate `>` and `>=` cases, we can call e.g. `checkIndex(index+1)`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,160 @@
...
-        return false;
+        return size == 0;
+    }
+
+    private Node<T> getNode(int index) {
+        checkIndex(index);
+        if (index >= size) {
+            throw new IndexOutOfBoundsException("No such element exists");
+        }
+        Node<T> sought;
```
maxline, line 113: please rename `sought` to the more accurate name

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,160 @@
...
+        if (index < size / 2) {
+            sought = firstNode;
+            for (int i = 1; i <= index; i++) {
```
maxline, line 116:
```suggestion
            for (int i = 1; i < index; i++) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,160 @@
...
+                sought = sought.next;
+            }
+        } else {
+            sought = lastNode;
+            for (int i = size - 2; i >= index; i--) {
```
maxline, line 121:
```suggestion
            for (int i = size - 1; i > index; i--) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,160 @@
...
+            }
+            node = node.next;
+        }
+        return null;
+    }
+
+    private void unlink(Node<T> node) {
+        if (node.next == null) {
+            if (node.prev != null) {
+                node.prev.next = null;
```
maxline, line 142: let's introduce variables for `node.next`, and `node.prev` in the begin of the method to increase readability.

maxline (changes requested):
Good job! let's improve the solution

.gitignore
```diff
@@ -1,3 +1,4 @@
 .idea/*
 *.iml
 target/*
+Notes.txt
```
Andriy-dz, line 4: don`t refactor this code

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,162 @@
...
+        }
+        size--;
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0) {
+            throw new IndexOutOfBoundsException("Can't operate negative index");
+        } else if (index >= size) {
+            throw new IndexOutOfBoundsException("Index " + index + " ot of bounds of size " + size);
+        }
```
Andriy-dz, line 150: it is better to combine these checks and raise an error with the inscription "Index " + index + " must be in the range from 0 to " + size; . Or something like that

Andriy-dz (approved):

PR #899, opened 2023-02-21
Solved task

KyslyiViacheslav:
implemented linked list

### DIFF
--- a/src/main/java/core/basesyntax/MyLinkedList.java
+++ b/src/main/java/core/basesyntax/MyLinkedList.java
@@ -1,6 +1,7 @@
 package core.basesyntax;
 
 import java.util.List;
+import java.util.Objects;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
     private int currentSize;
@@ -55,26 +56,51 @@ public class MyLinkedList<T> implements MyLinkedListInterface<T> {
     @Override
     public T remove(int index) {
         checkIndexBounds(index);
-        if (index == 0 && currentSize == 1) {
-            T tempToReturn = head.value;
-            unlinkSingleNode();
-            return tempToReturn;
-        } else {
-            Node<T> currentNode = getNodeByIndex(index);
-            remove(currentNode.value);
-            return currentNode.value;
-        }
+        Node<T> currentNode = getNodeByIndex(index);
+        T tempToReturn = currentNode.value;
+        unlink(currentNode);
+        return tempToReturn;
     }
 
     @Override
     public boolean remove(T value) {
-        if (isValuesEquals(head.value, value) && currentSize == 1) {
-            return unlinkSingleNode();
-        } else if (isValuesEquals(head.value, value)) {
-            return unlinkHeadNode();
-        }else{
-            return unlinkNodeInDifferentPosition(value);
+        boolean isRemoteSuccessful = false;
+        if (isValuesEquals(head.value, value)) {
+            isRemoteSuccessful = unlink(head);
+        } else {
+            isRemoteSuccessful = unlink(new Node<>(null, value, null));
         }
+        return isRemoteSuccessful;
+    }
+
+    private boolean unlink(Node<T> node) {
+        boolean isUnlinkSuccessful = false;
+        if (isValuesEquals(head.value, node.value) && currentSize == 1) {
+            tail.next = null;
+            tail.prev = null;
+            tail.value = null;
+            head = tail;
+            isUnlinkSuccessful = true;
+            currentSize--;
+        } else if (isValuesEquals(head.value, node.value)) {
+            head.next.prev = null;
+            head = head.next;
+            isUnlinkSuccessful = true;
+            currentSize--;
+        } else {
+            Node<T> currentNode = head;
+            while (currentNode != null) {
+                if (isValuesEquals(currentNode.value, node.value)) {
+                    currentNode.prev.next = currentNode.next;
+                    currentNode.next.prev = currentNode.prev;
+                    isUnlinkSuccessful = true;
+                    currentSize--;
+                    break;
+                }
+                currentNode = currentNode.next;
+            }
+        }
+        return isUnlinkSuccessful;
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,171 @@
...
+        if (index < 0 || index >= currentSize) {
+            throw new IndexOutOfBoundsException();
+        }
+    }
+
+    private void addFirstNode(T value) {
+        Node<T> node = new Node<>(null, value, null);
+        head = node;
+        tail = node;
+        currentSize++;
```
yurihhhT, line 118: I think better to use
 head = new Node<>(null, value, null);
 tail = head;

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,171 @@
...
+        Node<T> node = new Node<>(tail, value, null);
+        tail.next = node;
+        tail = node;
+        currentSize++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > currentSize) {
+            throw new IndexOutOfBoundsException();
```
yurihhhT, line 26: When you throw exeption write description as:
throw new IndexOutOfBoundsException("Index " + index + " out of bounds");

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,171 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return currentSize == 0;
+    }
+
+    private void checkIndexBounds(int index) {
+        if (index < 0 || index >= currentSize) {
+            throw new IndexOutOfBoundsException();
```
yurihhhT, line 110: same as on 25 and 26 line

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,171 @@
...
+        currentSize++;
+    }
+
+    private void unlink() {
+        tail.next = null;
+        tail.prev = null;
+        tail.value = null;
+        head = tail;
+        currentSize--;
+    }
```
yurihhhT, line 134: You can use universal method for all objects in LinkedList

yurihhhT (changes requested):
leaved commend

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,171 @@
 package core.basesyntax;
 
 import java.util.List;
+import java.util.Objects;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int currentSize;
+    private Node<T> head;
+    private Node<T> tail;
```
yurihhhT, line 9: test1

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,171 @@
...
     public void add(T value, int index) {
+        if (index < 0 || index > currentSize) {
+            throw new IndexOutOfBoundsException();
+        } else if (currentSize == index) {
+            add(value);
+            return;
+        } else if (index == 0) {
+            addNodeToZeroPosition(value);
+            return;
+        }
```
yurihhhT, line 33: test2

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,171 @@
...
+        return getNodeByIndex(index).value;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        checkIndexBounds(index);
+        Node<T> currentNode = getNodeByIndex(index);
+        T temp = currentNode.value;
+        currentNode.value = value;
```
yurihhhT, line 59: test3

yurihhhT (approved):
good job, mate!

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,182 @@
...
+            return;
+        }
+        addNodeToTailPosition(value);
     }
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > currentSize) {
+            throw new IndexOutOfBoundsException("Input index : " + index
+                    + " out of bonds: [" + 0 + ", " + currentSize + "]");
```
karma-o, line 23: redundant string concat
```suggestion
                    + " out of bonds: [0 - "  + currentSize + "]");
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,182 @@
...
 
     @Override
-    public boolean remove(T object) {
-        return false;
+    public boolean remove(T value) {
+        if (isValuesEquals(head.value, value) && currentSize == 1) {
+            return unlinkSingleNode();
+        } else if (isValuesEquals(head.value, value)) {
+            return unlinkHeadNode();
+        }
```
karma-o, line 75: I don't think we need an exclusive check of the head, I'd rather move the part of handling the case of removing the last element to the `unlink` method, to have a single one of these, and passed to it not a value but a node. 
Be careful of redundant iterations to find the node you are trying to delete

karma-o (changes requested):
An overall comment, it is good that you split your functionality, but IMHO I'd maybe not split the logic into such a tiny pieces, and you are rarely actually reusing it and having an actual impact

src/main/java/core/basesyntax/MyLinkedList.java
KyslyiViacheslav: I don't create one method unlink, where I can check Node position and unlink node from List, because I think that the single-responsibility method for unlink or add Node is more readable and simple for understanding, to the other dev which read my code first time and for me after some time. If this approach to writing code incorrect i can rewrite this logic.

Andriy-dz: I think it is better to listen to Vadim's advice, because all three methods perform the same function depending on the position of the element.

Andriy-dz (approved):

PR #900, opened 2023-02-21
Implemented my own LinkedList.

VitaliiPotseluiko:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,144 @@
...
+        return size == 0;
+    }
+
+    private Node<T> getCurrentNode(int index) {
+        Node<T> currentNode = head;
+        for (int i = 0; i < index; i++) {
+            currentNode = currentNode.next;
+        }
+        return currentNode;
+    }
```
OlehSarapuk, line 126: from common mistakes:
Don’t iterate your nodes from the head to tail or vice versa.
Pay attention to the index you receive as an input, you can iterate only first or second half of the list, depending on the index value. That will boost your lists’ performance.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,144 @@
...
+
```
OlehSarapuk, line 127: from common mistakes:
Create a separate method unlink(Node node).

OlehSarapuk (changes requested):

OlehSarapuk (approved):
well done!

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,156 @@
...
+        Node<T> currentNode = getCurrentNode(index);
+        return unlink(currentNode);
     }
 
     @Override
     public boolean remove(T object) {
+        if (size == 1) {
+            remove(size - 1);
+            return true;
+        }
```
sarakhmen, line 84: This could lead to a potential bug. Think of what will happen, when your list contains only 1 element, which does not equal a passed argument.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,156 @@
...
+
+    private T unlink(Node<T> node) {
+        if (node.prev != null) {
+            node.prev.next = node.next;
+        }
+        if (node.next != null) {
+            node.next.prev = node.prev;
+        }
+        if (node == head) {
+            head = node.next;
```
sarakhmen, line 115: don't forget to clear references from both sides:
```suggestion
        if (node == head) {
            head = node.next;
            head.prev = null;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,156 @@
...
+        }
+        if (node == tail) {
+            tail = node.prev;
```
sarakhmen, line 118: same issue here

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,156 @@
...
+        Node<T> anotherNode = new Node<>(currentNode.prev, value, currentNode);
+        currentNode.prev.next = anotherNode;
+        currentNode.prev = anotherNode;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
+        if (list == null) {
+            throw new RuntimeException("Current list is null!");
```
sarakhmen, line 49: unclear message

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,156 @@
...
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Index " + index + " is out of bound for length "
+                    + size);
+        }
+        if (index == size) {
+            add(value);
+            return;
+        }
```
sarakhmen, line 31: I'd recommend making this check first, and then calling checkIndex() method instead of above 'if' block

sarakhmen, line 31:
```suggestion
        if (index == size) {
            add(value);
            return;
        }
        checkIndex(index);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,156 @@
...
+            head.prev = newNode;
+            head = newNode;
+            size++;
+            return;
+        }
+        Node<T> currentNode = getCurrentNode(index);
+        Node<T> anotherNode = new Node<>(currentNode.prev, value, currentNode);
+        currentNode.prev.next = anotherNode;
+        currentNode.prev = anotherNode;
+        size++;
```
sarakhmen, line 43: I'd recommend simplifying this code. Think of how you can avoid duplications

sarakhmen (changes requested):
Let's improve your solution :)

Andriy-dz (changes requested):
not fixed

nacenik (approved):
Good job! 👍

PR #901, opened 2023-02-21
implementationed interface MyLinkedListInterface

IvanZuravlev228:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,184 @@
...
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Non-existent position");
+        }
+    }
+
+    private void checkElementIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("index out of bounds");
+        }
+    }
```
ArtemGrunin, line 173: in my opinion, these two methods can be combined into one (yours is checkElementIndex) with the message "Index Out Of Bounds!", by adding size to the signature and in the add(T value, int index) method, call checkIndex(index, size + 1);

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,184 @@
...
+    private void addFirst(T value) {
+        Node<T> h = head;
+        Node<T> newNode = new Node<>(null, value, h);
+        head = newNode;
+        if (h == null) {
+            tail = newNode;
+        } else {
+            h.prev = newNode;
+        }
+        size++;
```
ArtemGrunin, line 96: You don't need h variable hear. You can use head instead h if you replace head = newNode; on the line before size++

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,184 @@
...
+        Node<T> t = tail;
+        Node<T> newNode = new Node<>(t, value, null);
+        tail = newNode;
+        if (t == null) {
+            head = newNode;
+        } else {
+            t.next = newNode;
+        }
+        size++;
+    }
```
ArtemGrunin, line 109: you dont need t variable here. 

` private void addLast(T value) {       
        if (tail == null) {
            head = newNode;
        } else {
            tail.next = newNode;
        }
        tail = newNode;
        size++;
    }`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,184 @@
...
+                }
+            }
+        } else {
+            for (Node<T> x = head; x != null; x = x.next) {
+                if (object.equals(x.value)) {
+                    unlink(x);
+                    return true;
+                }
+            }
+        }
```
ArtemGrunin, line 73: could you refactor it with one loop? something like for (Node<T> x = head; x != null; x = x.next) {
                 if (x.value == null || object.equals(x.value)) {
                     unlink(x);
                     return true;
                 }
             }

ArtemGrunin (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
IvanZuravlev228: I can't do that because on line 90 "head" = "newNode" and on line 91 I will always get "false"

src/main/java/core/basesyntax/MyLinkedList.java
IvanZuravlev228: same problem like in first

src/main/java/core/basesyntax/MyLinkedList.java
IvanZuravlev228: if input will be null, you code will get mistake, becouse null.equals(x.value) - mistake

ArtemGrunin: of course u 'r' right.  
what about this issue:
if (x.value == null || object !=null && object.equals(x.value))

src/main/java/core/basesyntax/MyLinkedList.java
ArtemGrunin: ....... if you replace head = newNode; on the line 96 before size++

src/main/java/core/basesyntax/MyLinkedList.java
ArtemGrunin: same answer)....... if you replace tail = newNode; on the line 107 before size++

ArtemGrunin (changes requested):

ArtemGrunin (approved):
Nice )

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,176 @@
...
+
+    private void checkPositionIndex(int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Non-existent position");
+        }
+    }
+
+    private void checkElementIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("index out of bounds");
```
nacenik, line 161: Not enough info inside the exception message. Which index?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,176 @@
...
     }
 
     @Override
     public void add(T value, int index) {
+        checkPositionIndex(index);
+        if (index == size) {
+            addLast(value);
+        } else {
+            addBefore(value, findNode(index));
+        }
```
nacenik, line 27: Here you can optimize your code. 
firstly, check `index == size `and if not then call method `checkElementIndex`. That helps you to rid of almost the same check `checkPositionIndex`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,176 @@
...
+        element.value = null;
+        size--;
+        return unlinkValue;
+    }
+
+    private void checkPositionIndex(int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Non-existent position");
+        }
+    }
```
nacenik, line 157: you can rid of this method.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,176 @@
...
     public void addAll(List<T> list) {
+        for (T element : list) {
+            add(element);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkElementIndex(index);
```
nacenik, line 39: call `checkElementIndex` inside `findNode` to optimize your code

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,176 @@
...
     public T remove(int index) {
-        return null;
+        checkElementIndex(index);
+        return unlink(findNode(index));
     }
 
     @Override
     public boolean remove(T object) {
+        for (Node<T> x = head; x != null; x = x.next) {
+            if (Objects.equals(x.value, object)) {
```
nacenik, line 61: [Don't use class Objects.](https://mate-academy.github.io/jv-program-common-mistakes/java-core/collections/linked-list.html#dont-use-class-objects)

nacenik (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -19,10 +18,10 @@ public void add(T value) {
 
     @Override
     public void add(T value, int index) {
-        checkPositionIndex(index);
         if (index == size) {
             addLast(value);
         } else {
+            checkElementIndex(index);
```
maxline, line 24: this redundant, because you call checkElementIndex() inside of findeNode()

maxline (approved):
Good job! Approved with a small comment, take a notice for the future, please.

PR #904, opened 2023-02-21
implement MyLinkedList class

denys-mg:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,200 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private static class Node<E> {
+        private E item;
```
ArtemPonomar, line 103: В методах ця змінна називається value, краще її переіменувати щоб назви співпадали

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,200 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private static class Node<E> {
```
ArtemPonomar, line 102: Wrong placement, move inner class after all other methods of class MyLinkedList

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,200 @@
...
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Incorrect index " + index);
+        }
+        if (index == size) {
+            add(value);
+            return;
+        }
```
ArtemPonomar, line 32:
```suggestion
        if (index < 0 || index > size) {
            throw new IndexOutOfBoundsException("Incorrect index " + index);
        }
        if (index == size) {
            add(value);
            return;
        }
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,200 @@
...
     @Override
     public T remove(int index) {
-        return null;
+        checkIndexBounds(index);
+        if (index == size - 1) {
+            return removeLast();
+        }
+        if (index == 0) {
+            return removeFirst();
+        }
```
ArtemPonomar, line 72:
```suggestion
```
Can`t see the necessity to use these methods. 
method unlink should already cover this cases

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,200 @@
...
+        Node<T> nextFirst = head.next;
+        head = nextFirst;
+        if (nextFirst == null) {
+            tail = null;
+        } else {
+            nextFirst.prev = null;
+        }
+        size--;
+        return element;
+    }
```
ArtemPonomar, line 158: Delete these methods if you decide to not use them here: 
https://github.com/mate-academy/jv-linked-list/pull/904/files#r1114608108

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,200 @@
...
+
+    private static class Node<E> {
+        private E item;
+        private Node<E> prev;
+        private Node<E> next;
+
+        public Node(Node<E> prev, E item, Node<E> next) {
+            this.prev = prev;
+            this.item = item;
+            this.next = next;
```
ArtemPonomar, line 110: Generate and use getters and setters to access Node parameters.

src/main/java/core/basesyntax/MyLinkedList.java
ArtemPonomar, line 158: Delete these methods if you decide to not use them here: 
https://github.com/mate-academy/jv-linked-list/pull/904/files#r1114608108

ArtemPonomar (changes requested):
Student review

src/main/java/core/basesyntax/MyLinkedList.java
denys-mg: just let it be 'item' like in an original LinkedList

src/main/java/core/basesyntax/MyLinkedList.java
denys-mg: I don't agree with you, as we should first check if we get the correct index and then call add method. And checkIndexBounds() it's quite another condition, as we should let in this case (index == size).

src/main/java/core/basesyntax/MyLinkedList.java
denys-mg: I split this logic to get some optimization in it, aтв I think these methods should be here

src/main/java/core/basesyntax/MyLinkedList.java
denys-mg: Don't see any reason for this, as we use this class with its fields only inside MyLinkedList class

ArtemPonomar (approved):
approved

src/main/java/core/basesyntax/MyLinkedList.java
nacenik: Artem said a good improvement. That decreases your code. You have almost the same now, but with additional almost duplicated code. Look, if `index == size` - you just call the method that adds a new node to the end of the list, otherwise - you check an index using the method `checkIndexBounds(index);` which contains almost the same block of code: `index < 0 || index >= size`.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,200 @@
...
+        Node<T> node = findNodeByIndex(index);
+        addBeforeNode(value, node);
     }
 
     @Override
     public void addAll(List<T> list) {
+        Iterator<T> iterator = list.iterator();
+        while (iterator.hasNext()) {
+            add(iterator.next());
+        }
```
nacenik, line 46: you can use forEach loop here. It uses an iterator for collection under the hood

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,200 @@
...
+
+    private void addAtTheBeginning(T value) {
+        Node<T> first = head;
+        Node<T> newNode = new Node<>(null, value, first);
+        head = newNode;
+        if (first == null) {
+            tail = newNode;
+        } else {
+            first.prev = newNode;
+        }
```
nacenik, line 110: you can add logic from this block to the method `addBeforeNode(T value, Node<T> node)` and use only that method. That will optimize your code an program logic

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,200 @@
...
+            throw new IndexOutOfBoundsException("Incorrect index " + index);
+        }
+        if (index == size) {
+            add(value);
+            return;
+        }
+        if (index == 0) {
+            addAtTheBeginning(value);
+            return;
+        }
```
nacenik, line 36: this part could be deleted if improve method `addBeforeNode`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,200 @@
...
+        Iterator<T> iterator = list.iterator();
+        while (iterator.hasNext()) {
+            add(iterator.next());
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkIndexBounds(index);
```
nacenik, line 51: call this method inside `findNodeByIndex` to optimize your code, cause you have to check the index every time you need to get node by index.

nacenik (changes requested):

Andriy-dz (approved):

PR #906, opened 2023-02-22
Jv linked list solution

MelnychukMyhailo:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,171 @@
...
+                    + " No such index or index is less than 0");
+        }
+    }
+
+    private Node<T> getNodeByIndex(int index) {
+        indexValidation(index);
+        if (index >= size / 2) {
+            return countFromTail(index);
+        }
+        return countFromHead(index);
```
okuzan, line 119: good use case for ternary operator

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,171 @@
...
+    }
+
+    private void indexValidation(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Wrong index value."
+                    + " No such index or index is less than 0");
+        }
+    }
+
+    private void indexValidationForAdding(int index) {
```
okuzan, line 107: it's ok, not really necessary to extract logic used only once

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,171 @@
...
+        removedNode.prev.next = removedNode.next;
+        removedNode.next.prev = removedNode.prev;
+    }
+
+    private static class Node<T> {
+        private T value;
+        private Node<T> prev;
+        private Node<T> next;
+
+        Node(Node<T> prev,T value, Node<T> next) {
```
okuzan, line 165: `private`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,171 @@
...
     @Override
     public void addAll(List<T> list) {
+        if (list != null) {
+            for (T listElement : list) {
+                add(listElement);
+            }
+        } else {
+            throw new RuntimeException("Can`t add this list, because list is null."
+                    + " Try to use correct list");
+        }
```
okuzan, line 54: we don't need if else in such scenarios, you just call if with throwing first, then goes all the rest code. but here we don't need that - java.util.ArrayList would just throw NPE. Leave only for-loop and you'll have the same behavior (for-each uses iterator internally, and we can't get it from `null`)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,171 @@
...
+            newNode.prev = tail;
+            tail.next = newNode;
+        }
+        tail = newNode;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        indexValidationForAdding(index);
```
okuzan, line 26: we could use the same indexValidation method, if we placed it after if(index == size){}

okuzan (changes requested):

kozhukhovsky (approved):

PR #907, opened 2023-02-22
My own LinkedList Implemented

copypasterr:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,151 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> prev;
+    private Node<T> next;
```
Skrudj, line 7: it's better to call tail and head, or firstNode and lastNode. in code it's seems that you are operating not the head or tail, but the current prev or next

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,151 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private boolean areEqual(T a, T b) {
+        return (a == null && b == null) || (a != null && a.equals(b));
+    }
```
Skrudj, line 120: for this exists Objects.equals(a, b)

Skrudj (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
copypasterr: we shouldn't use class Object.

src/main/java/core/basesyntax/MyLinkedList.java
copypasterr: we shouldn't use class Object.

src/main/java/core/basesyntax/MyLinkedList.java
copypasterr: fixed

src/main/java/core/basesyntax/MyLinkedList.java
Skrudj: I also already met with that. agree with you

Skrudj (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,151 @@
...
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> tail;
+    private Node<T> head;
+    private int size;
+
+    public MyLinkedList() {
+        tail = null;
+        head = null;
+        size = 0;
```
nacenik, line 13: You really don't need to do it, cause class variables contain this values by default

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,151 @@
...
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private boolean areEqual(T a, T b) {
+        return (a == null && b == null) || (a != null && a.equals(b));
+    }
+
+    private void isIndexOutOfBoundsException(int index) {
```
nacenik, line 122: better to rename to a more appropriate name like `checkIndex` or `checkIndexBounds`, cause the word `exception` in check methods is not suitable

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,151 @@
...
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        isIndexOutOfBoundsException(index);
+        Node<T> current = tail;
+        for (int i = 0; i < index; i++) {
+            current = current.next;
+        }
```
nacenik, line 68: You should create the method `getNodeByIndex` or like this, and reuse it when you need to get node by index. [Don't forget about optimization](https://mate-academy.github.io/jv-program-common-mistakes/java-core/collections/linked-list.html#dont-iterate-your-nodes-from-the-head-to-tail-or-vice-versa). Also, it's good idea to call method `isIndexOutOfBoundsException` inside it to get rid of line-by-line calling this 2 methods.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,151 @@
...
+
+    private void isIndexOutOfBoundsException(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Index: " + index + ", Size: " + size);
+        }
+    }
+
+    private void unlink(Node<T> node) {
+        if (node.prev == null) {
+            tail = node.next;
```
nacenik, line 130: if `node.prev == null` is true, it means that you need to set `head = node.next`, cause head - first node in the list, tali - last node.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,151 @@
...
+        } else {
+            node.prev.next = node.next;
+        }
+        if (node.next == null) {
+            head = node.prev;
```
nacenik, line 135: if `node.next == null` is true, it means that you need to set `tail = node.prev`, cause head - first node in the list, tali - last node.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,151 @@
...
+        if (index == 0) {
+            current = tail;
+        } else if (index == size - 1) {
+            current = head;
+        } else {
+            current = tail;
+            for (int i = 0; i < index; i++) {
+                current = current.next;
+            }
+        }
```
nacenik, line 90: here you need to use method `getNodeByIndex` that you should create

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,151 @@
...
     }
 
     @Override
     public T get(int index) {
-        return null;
+        isIndexOutOfBoundsException(index);
+        Node<T> current = tail;
+        for (int i = 0; i < index; i++) {
+            current = current.next;
+        }
```
nacenik, line 58: here you need to use method `getNodeByIndex` that you should create

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,151 @@
...
+            head = head.next;
+        } else {
+            Node<T> current = tail;
+            for (int i = 0; i < index; i++) {
+                current = current.next;
+            }
+            current.prev.next = new Node<>(current.prev, value, current);
+            current.prev = current.prev.next;
+        }
+        size++;
```
nacenik, line 42: you need to divide logic from here into different methods and optimize some code. Small hint: if `index == size`, just call method add(T value).

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,151 @@
...
+
+    public MyLinkedList() {
+        tail = null;
+        head = null;
+        size = 0;
+    }
+
     @Override
     public void add(T value) {
+        add(value, size);
```
nacenik, line 18: add logic here to add a new node to the end of the list

nacenik (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,149 @@
...
+        if (tail == null) {
+            addFirst(value);
+        } else if (index == 0) {
+            tail.prev = new Node<>(null, value, tail);
+            tail = tail.prev;
+        } else if (index == size) {
+            addLast(value);
+        } else {
+            addMiddle(value, index);
+        }
```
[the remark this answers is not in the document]
copypasterr: I tried if wrong, advise how best to do

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,149 @@
...
+        if (tail == null) {
+            tail = head = new Node<>(null, value, null);
+        } else if (index == 0) {
+            addFirst(value);
+        } else if (index == size) {
+            addLast(value);
+        } else {
+            addMiddle(value, index);
+        }
```
maxline, line 28:
```suggestion
         if (index == size) {
            add(value);
            return:
        if (index == 0) {
            addFirst(value);
        } else {
            addMiddle(value, index);
        }
```

maxline (changes requested):
Please try this

Andriy-dz (approved):

PR #910, opened 2023-02-22
Maksym Yashyn HomeWork

Nimakel:


NickolayVoloshchuk (approved):
Good job, In my humble opinion)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,162 @@
...
+                }
+            }
+        } else {
+            for (Node<T> i = head; i != null; i = i.next) {
+                if (object.equals(i.item)) {
+                    unlink(i);
+                    return true;
+                }
+            }
+        }
```
KseniiaMakarova, line 145: actions are identical
let's use only 1 for-loop and check for item equality in 1 condition (use ||)

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,155 @@
...
+            add(value);
+            return;
+        }
+        checkIndex(index);
+        linkBefore(value, findByIndex(index));
     }
 
     @Override
     public void addAll(List<T> list) {
+        for (T node : list) {
```
maxline, line 27: better to rename as `value`, `node` is like inner implementation, not what the client pass as method parameter.

maxline (changes requested):
God job! I left just one small remark .

maxline (approved):
Fine! Approved

PR #912, opened 2023-02-22
Homework llist

Nikitos787:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,148 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void checkIndex(int index, int size) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Wrong index");
```
EliSoll, line 100: Just for clarifying the message we can add the current index inside the body of our exception's message.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,148 @@
...
+        }
+    }
+
+    private Node<T> getNodeByIndex(int index) {
```
EliSoll, line 104: As for me "ByIndex" + "index" as argument is redundant, getNode is enough)

EliSoll (approved):
good job overall! wait for mentors and good luck)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,148 @@
...
+        return oldItem;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index, size);
+        Node<T> getNode = getNodeByIndex(index);
+        T removeElement = getNode.item;
+        unLink(getNode);
```
maxline, line 71: you can change method param as `unLink(getNode)`   and call `getNode.item` inside the method body.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,148 @@
...
+        return removeElement;
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> current = head;
+        for (int i = 0; i < size; i++) {
+            if (current.item == object || (current.item != null) && current.item.equals(object)) {
+                unLink(getNodeByIndex(i));
```
maxline, line 80: It is not necessary to find node by index, you already have needed value - `current`. Calling `getNodeByIndex(i)` increase method complexity because it iterates over all values.
Also you can change `for` with index by `for each` loop.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,148 @@
...
+            for (int i = size - 1; i > index; i--) {
+                current = current.prev;
+            }
+        }
+        return current;
+    }
+
+    private void unLink(Node<T> removedElement) {
+        if (isEmpty()) {
+            throw new RuntimeException("Can't remove it");
```
maxline, line 122: `throw` section work as `return`,  so  `else` clause redundant and can by simplified.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,148 @@
...
+        } else {
+            current = tail;
+            for (int i = size - 1; i > index; i--) {
+                current = current.prev;
+            }
+        }
+        return current;
+    }
+
+    private void unLink(Node<T> removedElement) {
```
maxline, line 120: let's refactor the method that it returns not void but `T`, where `T` is removed element.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,148 @@
...
+        } else if (removedElement == head) {
+            head = removedElement.next;
+            head.prev = null;
+        } else if (removedElement == tail) {
+            tail = removedElement.prev;
+            tail.next = null;
+        } else {
+            removedElement.prev.next = removedElement.next;
+            removedElement.next.prev = removedElement.prev;
+        }
```
maxline, line 135: Let's refactor this code to encrease readability

```
        Node<T> prev = removedElement.prev;
        Node<T> next = removedElement.next;

        if (prev == null) {
            head = next;
        } else {
            prev.next = next;
            removedElement.prev = null;
        }
```
The similar check for `(next == null)`.
And set  `inputNode.item = null` in the end.
Where `Node<T> inputNode` - is method input parameter.

maxline (changes requested):
Good job! Let's improve the solution

nacenik (approved):
Perfect! 
Keep it up! 🔥

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,152 @@
...
     public void addAll(List<T> list) {
+        for (T element : list) {
+            add(element);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkIndex(index, size);
```
Andriy-dz, line 34: you don't need to check for index validation, because you do it in the getNode() method. The same correction in the following cases

Andriy-dz (changes requested):

Andriy-dz (approved):

PR #913, opened 2023-02-22
HomeWork JavaCore/LinkedList

NickolayVoloshchuk:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,185 @@
...
+        if (!isElementIndex(index)) {
+            throw new IndexOutOfBoundsException(outOfBoundsMsg(index));
+        }
+    }
+
+    private void checkPositionIndex(int index) {
+        if (!isPositionIndex(index)) {
+            throw new IndexOutOfBoundsException(outOfBoundsMsg(index));
+        }
+    }
```
Nimakel, line 110: In my opinion, we can simplify this methods to just one method and use that method. 
I don't know if this is considered a mistake, but i accept it

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,185 @@
...
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
     @Override
     public void add(T value) {
+        if (head == null) {
+            linkFirst(value);
+        } else {
+            linkLast(value);
```
maxline, line 15: linkFirst() and linkLast() methods called only one time, so no necessity introduce a methods for its. you can put the logic right here and take care about avoiding code duplication.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,185 @@
...
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        checkPositionIndex(index);
+        if (index == size) {
+            add(value);
```
maxline, line 24: we can add `return` after `add()`, then `else` clause can be simplified

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,185 @@
...
+        } else {
+            next.prev = prev;
+            node.next = null;
+        }
+        node.value = null;
+        size--;
+        return element;
+    }
+
+    private Node<T> node(int index) {
```
maxline, line 134: let's rename method as verb, e.g. `getNode()` or `getByIndex()`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,185 @@
...
     @Override
     public T remove(int index) {
-        return null;
+        checkElementIndex(index);
+        return unlink(node(index));
     }
 
     @Override
     public boolean remove(T object) {
+        for (Node<T> i = head; i != null; i = i.next) {
```
maxline, line 69: Let's rename `i`, cause it is not an index. 
And I think will be better to use `while` for iterating, because `for with index` is commonly used for iterating structures that have index, which is absent here.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,185 @@
...
+        return size == 0;
+    }
+
+    private boolean isElementIndex(int index) {
+        return index >= 0 && index < size;
+    }
+
+    private boolean isPositionIndex(int index) {
+        return index >= 0 && index <= size;
+    }
```
maxline, line 94: This two method can be merged to one. we can differentiate them by put `index` or `index-1` when calling. And in result the code can be put to the `checkIndex(index)`, so we will have one method for checking instead of four.

src/main/java/core/basesyntax/MyLinkedList.java
maxline: agree, commented above

src/main/java/core/basesyntax/MyLinkedList.java
NickolayVoloshchuk: if 0 will be checked?

NickolayVoloshchuk: If I should check 0?

src/main/java/core/basesyntax/MyLinkedList.java
NickolayVoloshchuk: I cannot agree with that. If you show me how to implement it better - only then will I change my mind.

src/main/java/core/basesyntax/MyLinkedList.java
NickolayVoloshchuk: I found another solution to this problem, tnx.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,153 @@
...
     @Override
     public void add(T value) {
+        Node<T> newNode;
+        if (head == null) {
+            newNode = new Node<>(null, value, null);
+            head = newNode;
+            tail = newNode;
+        } else {
+            Node<T> node = tail;
+            newNode = new Node<>(node, value, null);
```
nacenik, line 19: You don't need an additional variable here. You can use tail 
```suggestion
            newNode = new Node<>(tail, value, null);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,153 @@
...
+        for (T value : list) {
+            add(value);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkIndex(index);
+        return getByIndex(index).value;
```
nacenik, line 55: to optimize your code, you can call the method `checkIndex` inside `getByIndex`

nacenik (changes requested):
Almost done, just fix some minor mistakes

maxline (approved):
Well done! approved!

PR #917, opened 2023-02-22
Implemented MyLinkedList

De1eF:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,185 @@
...
+    private Node<T> tail = null;
+    private int size;
+
+    public Node<T> getHead() {
+        return head;
+    }
+
+    public Node<T> getTail() {
+        return tail;
+    }
```
Hlib13, line 17: You don't need this getters.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,185 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private Node<T> getNodeByIndex(int index) {
+        checkForValidIndex(index);
+        if (index >= size) {
```
Hlib13, line 103: add here index < 0 check

src/main/java/core/basesyntax/MyLinkedList.java
karma-o: that's true, remove them and reference variables directly

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,173 @@
...
+            tail = current;
+        }
+    }
+
+    static class Node<T> {
+        private Node<T> next;
+        private Node<T> prev;
+        private T element;
+
+        Node(T element, Node<T> next, Node<T> prev, int index) {
```
karma-o, line 167: redundant index parameter

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,173 @@
...
+    private void connectToNext(Node<T> current, Node<T> next) {
+        if (next == null) {
+            current.next = null;
+            return;
+        }
+        current.next = next;
+        next.prev = current;
+    }
+
+    private void setHeadNTail(Node<T> current) {
```
karma-o, line 153:
```suggestion
    private void setHeadAndTail(Node<T> current) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,173 @@
...
+
     @Override
     public void add(T value) {
+        add(value, size);
     }
 
     @Override
     public void add(T value, int index) {
+        checkForValidIndex(index);
+        Node<T> newNode = new Node<>(value,null, null, size);
```
karma-o, line 27: missing index check for size, we should think where to put it, perfect decision would be to add it to the index check method, or have 2 index check methods for each size check variant

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,173 @@
...
+        if (index < size) {
+            connectToNext(newNode, getNodeByIndex(index));
+        }
+        if (index > 0) {
+            connectToPrev(newNode, getNodeByIndex(index - 1));
```
karma-o, line 32: I think I understand what you did there, with assigning the nodes based on the index instead of strict cases like if the list is empty, or `index == size`, or index is the middle section. It is quite original, and I do not see any problems with that so far. so I think if it works we can stick with that. 
The only thing I wanted to mention is to not call `getNodeByIndex()` twice, have a var, and get a second one from the first one

karma-o (changes requested):

sarakhmen (changes requested):
Please fix failed CI

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,177 @@
...
-        return null;
+        T prevElement = getNodeByIndex(index).element;
+        getNodeByIndex(index).element = value;
+        return prevElement;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkForValidIndex(index);
```
jaroslavcyk, line 51: you can remove this line cause you checkValidIndex in getNodeByIndex

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,177 @@
...
+    private int size;
+
     @Override
     public void add(T value) {
+        add(value, size);
     }
 
     @Override
     public void add(T value, int index) {
+        checkForValidIndex(index);
```
jaroslavcyk, line 18: you can remove this line cause you checkValidIndex in getNodeByIndex

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,177 @@
...
+        private Node<T> prev;
+        private T element;
+
+        Node(T element, Node<T> next, Node<T> prev) {
+            this.element = element;
+            this.next = next;
+            this.prev = prev;
+        }
     }
 }
```
jaroslavcyk, line 177: add empty line in the end

jaroslavcyk (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,177 @@
...
+            setHeadAndTail(prev);
+        }
+        if (next != null) {
+            connectToPrev(next, prev);
+            setHeadAndTail(next);
+        }
+        size--;
+    }
+
+    private void checkForValidIndex(int index) {
```
karma-o, line 133: `validateIndex` or `checkIndex` sounds better

karma-o (approved):

PR #920, opened 2023-02-22
Implemented solution for hw.

RandomEastEcho:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,139 @@
...
+        indexLengthCheck(index, length);
+        Node<T> temp;
+        temp = head;
+        for (int i = 0; i < index; i++) {
+            temp = temp.next;
+        }
+        return temp;
+    }
+
+    private void removeLink(Node<T> node) {
```
s-hrynus, line 129: I think u should rename to "unlink" like in cm list.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,139 @@
...
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
+        if (list == null) {
+            throw new RuntimeException("This list is null!");
+        }
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
```
s-hrynus, line 66: mb better for each, will be simply 

```suggestion

                for (T t : list) {
                      add(t);
                }
```

s-hrynus (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,139 @@
...
+        private Node<T> prev;
+        private T value;
+        private Node<T> next;
+
+        public Node(Node<T> prev, T value, Node<T> next) {
+            this.prev = prev;
+            this.value = value;
+            this.next = next;
+        }
+    }
```
petro-ushchuk, line 20: It would be good to keep inner classes in the bottom of class.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,139 @@
...
     @Override
     public void add(T value) {
+        Node<T> newNode;
+        if (head == null) {
+            newNode = new Node<>(null, value, null);
+            head = newNode;
+        } else {
+            newNode = new Node<>(tail, value, null);
+            tail.next = newNode;
+        }
```
petro-ushchuk, line 31: if your head is null, then tail also null. you can initilzie node always `new Node<>(tail, value, null);` on the line 24 and remove lines 26 and 29

petro-ushchuk (approved):
Good work! Left in comments some ideas with improvments 🔥

PR #924, opened 2023-02-22
jv - LL

s-hrynus:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size() == 0;
+    }
+
+    private void checkIndex(int index) {
+        if (!(index >= 0 && index < size)) {
+            throw new IndexOutOfBoundsException("Index of bound. Input index is wrong!");
```
RandomEastEcho, line 99: Add index output in your exeption.  For example ("some error message" + index);

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
+        }
+    }
+
+    private void checkAddIndex(int index) {
+        if (!(index >= 0 && index <= size)) {
+            throw new IndexOutOfBoundsException("Index of bound. Input index is wrong!");
```
RandomEastEcho, line 105: Same here - add index output.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
+            } else {
+                prevNode.next = newNode;
+            }
+        }
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
+        for (T element : list) {
```
RandomEastEcho, line 48: Add null check for list.

RandomEastEcho (approved):
Approved but added some suggestions.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    private static class Node<T> {
```
karma-o, line 10: move to the bottom

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
+    private void link(T value) {
+        Node<T> prevNode = tail;
+        Node<T> newNode = new Node<>(tail, value, null);
+        tail = newNode;
+        if (prevNode == null) {
+            head = newNode;
+        } else {
+            prevNode.next = newNode;
+        }
+    }
```
karma-o, line 118: this method is called in `add(value, index)`, which is called in `add(value)`, so all of that is basically a redundant method calls, just move the entire logic of that method to `add(value)`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
     public void add(T value) {
+        add(value, size);
     }
 
     @Override
     public void add(T value, int index) {
+        checkAddIndex(index);
+        if (index == size) {
+            link(value);
+        } else {
```
karma-o, line 32: in your case it is always false
```suggestion
        if (index == size) {
            add(value);
            return;
        }
        checkAddIndex(index);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
+        }
+    }
+
     @Override
     public void add(T value) {
+        add(value, size);
     }
 
     @Override
     public void add(T value, int index) {
```
karma-o, line 28: more `return`s and less `else`s with overcomplicated `if`s

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
+            node = head;
+            for (int i = 0; i < index; i++) {
+                node = node.next;
+            }
+        }
+        return node;
+    }
+
+    private boolean compareEquality(T element1, T element2) {
+        return ((element2 == element1) || (element1 != null && element1.equals(element2)));
```
karma-o, line 137: redundant braces

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
+        if (prev == null) {
+            head = next;
+        } else {
+            prev.next = next;
+            findNode.prev = null;
+        }
+        findNode.element = null;
+        size--;
+        return element;
     }
```
karma-o, line 159: can be a bit shorter

src/main/java/core/basesyntax/MyLinkedList.java
karma-o: agree

src/main/java/core/basesyntax/MyLinkedList.java
karma-o: agree

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
+    }
+
+    private void checkIndex(int index) {
+        if (!(index >= 0 && index < size)) {
+            throw new IndexOutOfBoundsException("Index of bound. Input index is wrong!");
+        }
+    }
+
+    private void checkAddIndex(int index) {
+        if (!(index >= 0 && index <= size)) {
```
karma-o, line 104: better to avoid negating and overcomplicating conditions if we can, check other method as well
```suggestion
        if (index < 0 && index > size) {
```

karma-o (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -96,13 +96,13 @@ public boolean isEmpty() {
 
     private void checkIndex(int index) {
         if (!(index >= 0 && index < size)) {
-            throw new IndexOutOfBoundsException("Index of bound. Input index is wrong!");
+            throw new IndexOutOfBoundsException("Index of bound. Input index is wrong! " + index);
```
karma-o, line 99: can we make it more pretty?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,151 @@
...
     }
 
     @Override
     public void add(T value, int index) {
+        if (index == size) {
+            add(value);
+            return;
+        }
+        checkAddIndex(index);
+        Node<T> node = findNode(index);
```
petro-ushchuk, line 30: Better move `checkAddIndex` inside `findNode` instead of calling it every time before this method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,151 @@
...
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Index of bound. Input index is wrong! " + index);
+        }
+    }
+
+    private void checkAddIndex(int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Index of bound. Input index is wrong! " + index);
+        }
+    }
```
petro-ushchuk, line 115: You can create 1 method instead of these 2 by passing `size` or `size + 1` as a parameter of the method.

petro-ushchuk (changes requested):
Very good solution, try to fix these comments and we are done here 🔥

src/main/java/core/basesyntax/MyLinkedList.java
karma-o: It seems like you misunderstood Petro here, he wanted to sort of combine these 2 methods in 1 by adding an upperBound to the parameters of this method instead of a hardcoded `size`. 
I will approve since it was an already working and nice solution before that comment but let's pay more attention next tim

karma-o (approved):
See the comment above ⬆️

PR #926, opened 2023-02-23
wrote an implementation of MyLinkedList

pechivo:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,166 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void validateIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException(index + " index too big for the size " + size);
```
en4ik99, line 120: I think that here it is necessary to change the messages,  if the index is less than zero?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,166 @@
...
+        Node<T> node = getNode(index);
+        return unlink(node);
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> node = head;
+        int index = 0;
+        while (node != null) {
+            if (Objects.equals(node.item, object)) {
```
en4ik99, line 98: As describe in typical mistake. we should not use class object.

src/main/java/core/basesyntax/MyLinkedList.java
pechivo: thanks for review)

en4ik99 (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,162 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    public MyLinkedList() {
+    }
```
petro-ushchuk, line 11: Redundant constructor, u already have this one by default. Google `default constructor java`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,162 @@
...
+
+        public Node(T item) {
+            this.item = item;
+        }
+    }
+
     @Override
     public void add(T value) {
+        Node<T> newNode = new Node<>(value);
+        if (size == 0) {
```
petro-ushchuk, line 26: use your `isEmpty` instead

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,162 @@
...
 
     @Override
     public void add(T value, int index) {
+        if (index == size) {
+            add(value);
+            return;
+        }
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Invalid index " + index + " for size " + size);
+        }
```
petro-ushchuk, line 43: You already have `validateIndex`, try to pass `size` (in other cases) or `size + 1` (in this case) as a parameter

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,162 @@
...
+        Node<T> nodeByIndex = getNode(index);
+        T oldValue = nodeByIndex.item;
+        nodeByIndex.item = value;
+        return oldValue;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        validateIndex(index);
```
petro-ushchuk, line 86: Move `validateIndex` inside `getNode` and remove it before you call `getNode` here and other places

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,162 @@
...
+                tail = null;
+            }
+        } else if (node == tail) {
+            tail.prev.next = null;
+            tail = tail.prev;
+        } else {
+            node.prev.next = node.next;
+            node.next.prev = node.prev;
+        }
+        size--;
```
petro-ushchuk, line 161: move `size--` in the start of method, and call `!isEmpty()` instead of `size != 1`

petro-ushchuk (changes requested):
Good work! Please, fix these small issues and we are done here 🥇

src/main/java/core/basesyntax/MyLinkedList.java
pechivo: thanks for idea!

pechivo:
> Good work! Please, fix these small issues and we are done here 🥇

thanks for the review, I have made changes👌

karma-o (approved):
🔥 🚒

PR #931, opened 2023-02-23
added hw solution

KyrylH:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,184 @@
...
+            linkFirst(value);
+        } else {
+            linkLast(value);
+        }
     }
 
     @Override
     public void add(T value, int index) {
+        if (index > size || index < 0) {
+            throw new IndexOutOfBoundsException("");
```
pechivo, line 26: write a message to exception with the value of variables Index, 
that invalid with such size, for example

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,184 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void linkFirst(T value) {
+        Node<T> h = head;
```
pechivo, line 93: better to change name "h" for more understandable

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,184 @@
...
+            unlinkElement.next = null;
+        }
+        unlinkElement.element = null;
+        size--;
+        return element;
+    }
+
+    private void checkPositionIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Element doesn't exist");
```
pechivo, line 171: add variable of index in exception message

pechivo (changes requested):

pechivo (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,184 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    public MyLinkedList() {
+        size = 0;
+    }
```
petro-ushchuk, line 12: Useless constructor, size is 0 by default. Remove it.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,184 @@
...
     public void addAll(List<T> list) {
+        for (T element : list) {
+            add(element);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkPositionIndex(index);
```
petro-ushchuk, line 44: Move `checkPositionIndex` inside `getNode` and remove usages before `getNode`. You make a checks only when call `getNode`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,184 @@
...
+        }
+        size++;
+    }
+
+    private Node<T> getNode(int index) {
+        Node<T> node;
+        int i;
+        if (size == 0) {
+            node = null;
+        }
```
petro-ushchuk, line 133: No need to check it, when your size `0` you throw an exception in `checkPositionIndex`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,184 @@
...
+        Node<T> oldNode = getNode(index);
+        T oldValue = oldNode.element;
+        oldNode.element = value;
+        return oldValue;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkPositionIndex(index);
```
petro-ushchuk, line 59: Previous comment

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,184 @@
...
     public T get(int index) {
-        return null;
+        checkPositionIndex(index);
+        return getNode(index).element;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        checkPositionIndex(index);
```
petro-ushchuk, line 50: Previous comment

petro-ushchuk (changes requested):
Good work! Please, fix these small issues

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,175 @@
...
+        } else {
+            linkLast(value);
+        }
     }
 
     @Override
     public void add(T value, int index) {
+        if (index > size || index < 0) {
+            throw new IndexOutOfBoundsException("Element doesn't exist at " + index);
+        }
```
petro-ushchuk, line 23: U already have method `checkPositionIndex` reuse it

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,175 @@
...
+            }
+        }
+        if (object != null) {
+            for (Node<T> x = head; x != null; x = x.next) {
+                if (object.equals(x.element)) {
+                    unlink(x);
+                    return true;
+                }
+            }
+        }
```
petro-ushchuk, line 71: Please DRY, u already have the same loop, just move `object.equals(x.element)` condition in the line `59`

petro-ushchuk (changes requested):
Very good solution, but we should fix these small issues.

src/main/java/core/basesyntax/MyLinkedList.java
KyrylH: These two checks are different

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,175 @@
...
+        for (Node<T> x = head; x != null; x = x.next) {
+            if (x.element == object) {
+                unlink(x);
+                return true;
+            }
+            if (object != null && object.equals(x.element)) {
+                unlink(x);
+                return true;
+            }
+        }
```
karma-o, line 65: Petro meant this but you should have asked him that, if you think you may have misunderstood something
```suggestion
            if (x.element == object || object != null && object.equals(x.element)) {
                unlink(x);
                return true;
            }
        }
```

karma-o (approved):
Approved because of a bit of a misleading comment on a minor thing, but let's be more proactive next time! 😉

PR #934, opened 2023-02-23
solved task

slaybrute:


olekskov (changes requested):
Good job, need to implement some code enhancements

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,144 @@
...
+            head = newNode;
+        } else {
+            tail.prev.next = newNode;
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index <= size && index >= 0) {
```
uliana-andrushchuk, line 25: You can use `checkIfIndexIsValid` method here instead of `if else`

uliana-andrushchuk (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
slaybrute: I can't use here my checkIfIndexIsValid method, because in the metho body such condition if (!(index < size && index >= 0)). But create new method for only one condition it's no sense.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,144 @@
...
+        }
+        size--;
+        return removedNode.value;
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> tmpNode = head;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(tmpNode.value, object)) {
```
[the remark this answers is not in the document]
slaybrute: Okay, I'll remove class Object, but It reduce lines of code. Why we couldn't use it ?

Ivan95kos (approved):
Great job

PR #935, opened 2023-02-23
double-linked list relize

kromch:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,141 @@
...
+                add(element);
+            }
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        Node<T> tempPoint = getByIndex(index);
+        return tempPoint != null ? tempPoint.data : null;
```
petro-ushchuk, line 56: better to throw an exception like `IndexOutOfBoundException`. `null` returns smells very bad, for these proposes better to throw an exception or use `Optional` class. In this case, please, throw an exception.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,141 @@
...
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        T result = null;
+        Node<T> tempPoint = getByIndex(index);
+        if (tempPoint != null) {
```
petro-ushchuk, line 63: Throw exception in `getByIndex`, check comment in the top

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,141 @@
...
+        }
+        return result;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        T result = null;
+        Node<T> tempPoint = getByIndex(index);
+        if (tempPoint != null) {
```
petro-ushchuk, line 74: Comment in the top

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,141 @@
...
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private Node<T> getByIndex(int index) {
+        if (index < 0 || index > size - 1) {
+            throw new IndexOutOfBoundsException("Invalid index : " + index + " Out bounds of List");
+        }
```
petro-ushchuk, line 122: U throw an exception in the case index out of bound, so you can remove these `if` in the top regarding to comments

petro-ushchuk (changes requested):
Very-very good work! But too much `if` cases, refactor your code regarding to comments.

src/main/java/core/basesyntax/MyLinkedList.java
kromch: done. all checkouts on null i took away.
null value added to throw exception condition

src/main/java/core/basesyntax/MyLinkedList.java
kromch: ok

src/main/java/core/basesyntax/MyLinkedList.java
kromch: ok

src/main/java/core/basesyntax/MyLinkedList.java
kromch: ok

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,132 @@
...
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private Node<T> getByIndex(int index) {
+        Node<T> tempPoint = first;
+        if (index < 0 || index > size - 1 || tempPoint == null) {
+            throw new IndexOutOfBoundsException("Invalid index : " + index + " Out bounds of List");
+        }
```
karma-o, line 116: * index check should be a separate method 
* remove null check we have size here so it will handle this
* why `size - 1`? shouldn't it be `index >= size` comparing some index to -1 seems strange to me

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,132 @@
...
+        for (int i = 0; i < index; i++) {
+            tempPoint = tempPoint.next;
+        }
```
karma-o, line 119: I can see that you have seen the [checklist](https://mate-academy.github.io/jv-program-common-mistakes/java-core/collections/linked-list.html) for this task, but I do not see any **main points** from it, one of them is finding the node by index with iterating from the beginning or from the end, depending on what is closer to it

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,132 @@
...
 
     @Override
     public boolean remove(T object) {
-        return false;
+        boolean result = false;
+        Node<T> tempPoint = first;
+        for (int i = 0; i < size; i++) {
+            if (tempPoint != null && (tempPoint.data == object || (tempPoint.data != null
+                    && tempPoint.data.equals(object)))) {
+                remove(i);
```
karma-o, line 93: another main point of the [checklist](https://mate-academy.github.io/jv-program-common-mistakes/java-core/collections/linked-list.html), we should have method `unlink(node)`, because you are iterating over your nodes 2 times in a row to find the right one: on line 90 and on line 117

karma-o (changes requested):
You need to pay much more attention, to the checklist first of all + all of the other details, they are really important and should not be ignored, just because our code works

karma-o (commented):
Also, do not use `#number` in commits, comments or anywhere basically if you do not want to mention some PR in it, `#` means PR number like #935

src/main/java/core/basesyntax/MyLinkedList.java
kromch: ok

src/main/java/core/basesyntax/MyLinkedList.java
kromch: ok

src/main/java/core/basesyntax/MyLinkedList.java
kromch: took away

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,160 @@
...
+    private Node<T> getByIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Invalid index : " + index + " Out bounds of List");
+        }
+        Node<T> tempPoint = index < (size >> 1) ? first : last;
+        boolean direct = tempPoint == first;
+        for (int i = direct ? 0 : size - 1; direct ? i < index : i > index;
+                i = direct ? i + 1 : i - 1) {
+            tempPoint = direct ? tempPoint.next : tempPoint.prev;
+        }
```
sarakhmen, line 135: Unreadable code, let's simplify it

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,160 @@
...
     }
 
     @Override
     public boolean remove(T object) {
-        return false;
+        boolean result = false;
+        Node<T> tempPoint = first;
+        for (int i = 0; i < size; i++) {
+            if (tempPoint != null && (tempPoint.data == object || (tempPoint.data != null
+                    && tempPoint.data.equals(object)))) {
```
sarakhmen, line 92: this condition could be simplified

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,160 @@
...
+            last = tempPoint.prev;
+        }
+        size--;
+        return tempPoint.data;
     }
 
     @Override
     public boolean remove(T object) {
-        return false;
+        boolean result = false;
```
sarakhmen, line 88: let's avoid using this variable

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,160 @@
...
+            tempPoint.prev.next = tempPoint.next;
+        }
+        if (tempPoint.prev == null) {
+            first = tempPoint.next;
+        }
+        if (tempPoint.next == null) {
+            last = tempPoint.prev;
+        }
+        size--;
+        return tempPoint.data;
```
sarakhmen, line 83: let's try to reuse your exclude() method here.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,160 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void exclude(Node<T> node) {
```
sarakhmen, line 112:
```suggestion
    private void unlink(Node<T> node) {
```

sarakhmen (changes requested):
Let's improve your solution

src/main/java/core/basesyntax/MyLinkedList.java
kromch: ok

src/main/java/core/basesyntax/MyLinkedList.java
kromch: used

src/main/java/core/basesyntax/MyLinkedList.java
kromch: took away

src/main/java/core/basesyntax/MyLinkedList.java
kromch: ok

sarakhmen (approved):
Good job

PR #936, opened 2023-02-23
implemented version of LinkedList

oleksiistepaniak:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,132 @@
...
+        } else {
+            currentNode.prev.next = newNode;
+        }
+        currentNode.prev = newNode;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
+        for (int i = 0; i < list.size(); i++) {
```
maxline, line 42: it is better use `for each` here `for (T value : list) {`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,132 @@
...
-        return null;
+        Node<T> currentNode = findByIndex(index);
+        unlink(currentNode);
+        return currentNode.value;
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> currentNode = head;
+        for (int i = 0; i < size; i++) {
```
maxline, line 70: it is better to use  ` while (node != null) {` instead of `for` with index

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,132 @@
...
+            if (currentNode.value == object || currentNode.value != null
+                    && currentNode.value.equals(object)) {
+                remove(i);
```
maxline, line 73: just call `unlink(currentNode)` instead of `remove(i)`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,132 @@
...
+        return size == 0;
+    }
+
+    private Node<T> findByIndex(int index) {
+        if (index >= size || index < 0) {
+            throw new IndexOutOfBoundsException("Invalid index: " + index + " Size: " + size);
+        }
+        int currentIndex = 0;
+        Node<T> currentNode = head;
+        while (currentNode != null) {
```
maxline, line 97: The searching can be improved. Let's divide the list by 2 parts, and make iteration from head or from tail, depend on index is closer to the head or tail.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,132 @@
...
+        return head;
+    }
+
+    private void unlink(Node<T> currentNode) {
+        if (currentNode == head) {
+            head = currentNode.next;
+            if (size > 1) {
+                head.prev = null;
+            }
+        } else if (currentNode == tail) {
```
maxline, line 113: Let's try to refactor the whole `if else` block in this method not by checking `size` and `== head`, but checking `currentNode.prev == null` `currentNode.next == null` etc.
You also can introduce a new variables for better readability e.g.:
```
   final T element = currentNode.value;
        final Node<T> next = currentNode.next;
        final Node<T> prev = currentNode.prev;
```

maxline (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
Andriy-dz: he probably meant to record something like this. it would also be good to make variables for currentNode.next and currentNode.prev. Maybe I got the names wrong
```suggestion
       if (currentNode.prev == null) {
            first = currentNode.next;
        } else {
            currentNode.prev.next = currentNode.next;
        }
        if (currentNode.next == null) {
            last = currentNode.prev;
        } else {
            currentNode.next.prev = currentNode.prev;
        }
```

src/main/java/core/basesyntax/MyLinkedList.java
Andriy-dz: and you don`t fix this

Andriy-dz (changes requested):

nacenik (changes requested):
Almost done, fix one mistake

nacenik (approved):
Perfect! 
Keep it up! 👍

PR #939, opened 2023-02-23
add decision for linkedList 1.0

vfedychkanych:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,178 @@
...
+    private void linkLast(T element) {
+        Node<T> newNode = new Node<>(tail, element, null);
+        tail.next = newNode;
+        tail = newNode;
+        size++;
+    }
+
+    private void unlink(Node<T> node) {
+        size--;
+        if (node == head && node == tail) {
```
ArtemGrunin, line 81: replace  with if (seze == 1)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,178 @@
...
+            head = null;
+            tail = null;
+            return;
```
ArtemGrunin, line 84: you do not need 'return' here

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,178 @@
...
+        } else if (node == head) {
+            head = node.next;
+            node.next.prev = null;
+            return;
```
ArtemGrunin, line 88: you do not need 'return' here

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,178 @@
...
+        } else if (node == tail) {
+            tail = node.prev;
+            node.prev.next = null;
+            return;
```
ArtemGrunin, line 92: you do not need 'return' here

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,178 @@
...
+            }
+            node = node.next;
+        }
+        throw new RuntimeException("Element " + value + " didn't find");
+    }
+
+    private Node<T> findNodeByIndex(int index) {
+        indexCheck(index);
+        int part = size % 2 == 0 ? size / 2 : size / 2 + 1;
+        if (index < part) {
```
ArtemGrunin, line 39: you can check just if ((size / 2) > index) without variable part )

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,178 @@
...
+        }
+        linkBefore(value, findNodeByIndex(index));
     }
 
     @Override
     public void addAll(List<T> list) {
+        for (T item : list) {
+            this.add(item);
+        }
     }
```
ArtemGrunin, line 138: There should be a check for null

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,178 @@
...
+        } else {
+            prev.next = newNode;
+        }
+        size++;
+    }
+
+    private void setFirstValue(T value) {
+        Node<T> newNode = new Node<>(null, value, null);
+        head = newNode;
+        tail = newNode;
```
ArtemGrunin, line 68: You can write 
head = tail = newNode;
if you want))

ArtemGrunin (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
vfedychkanych: I need return to exit the method at the moment I need

src/main/java/core/basesyntax/MyLinkedList.java
vfedychkanych: there the method add(item) is called in which everything is checked, therefore 2 the same check is not needed

src/main/java/core/basesyntax/MyLinkedList.java
vfedychkanych: There, everything is checked through the tail and head variables, so if I change 1 if(), then I need to change the logic of the method

src/main/java/core/basesyntax/MyLinkedList.java
ArtemGrunin: ок) let it be)

src/main/java/core/basesyntax/MyLinkedList.java
ArtemGrunin: what if  list == null  ?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,179 @@
...
+    }
+
+    private void indexCheck(int index) {
+        if (index >= size || index < 0) {
+            throw new IndexOutOfBoundsException("Your index was " + index
+                    + ", but size is " + size);
+        }
+    }
+
+    private Node<T> findNodeByValue(T value) {
```
karma-o, line 24: private go after public ones

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,179 @@
...
 
     @Override
     public void add(T value, int index) {
+        addCheck(index);
+        if (isEmpty()) {
+            setFirstValue(value);
+            return;
+        }
+        if (index == size) {
+            linkLast(value);
```
karma-o, line 125: I think we can remove the first `if` by replacing it
```suggestion
            add(value);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,179 @@
...
+            return;
+        }
+        linkBefore(value, findNodeByIndex(index));
     }
 
     @Override
     public void addAll(List<T> list) {
+        if (list == null) {
+            throw new IndexOutOfBoundsException("List is null");
+        }
```
karma-o, line 135: we do not have any index in this method to throw this exception
```suggestion
        if (list == null) {
            return;
        }
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,179 @@
...
+    private Node<T> findNodeByValue(T value) {
+        Node<T> node = head;
+        for (int i = 0; i < size; i++) {
+            if (value == node.item || value != null
+                    && value.equals(node.item)) {
+                return node;
+            }
+            node = node.next;
+        }
+        throw new RuntimeException("Element " + value + " didn't find");
```
karma-o, line 33: it should have been `NoSuchElementException` , but overall check the comment bellow

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,179 @@
...
 
     @Override
     public boolean remove(T object) {
-        return false;
+        try {
+            unlink(findNodeByValue(object));
+        } catch (RuntimeException e) {
+            return false;
+        }
+        return true;
```
karma-o, line 169: * first of all, should we catch runtime exceptions? It is a bad practice to do your logic with try\catch blocks instead of booleans and ifs, even though it is not the best either
* Second thing is that here having a separate method is causing more complexity, could have just get the logic from `findNodeByValue` and instead of returning the node, deleted it right away and returned true, and false otherwise, we can have it like this since we are not reusing it anywhere else at the moment

karma-o (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,174 @@
...
+            head = node.next;
+            node.next.prev = null;
+            return;
+        } else if (node == tail) {
+            tail = node.prev;
+            node.prev.next = null;
+            return;
+        }
+        node.next.prev = node.prev;
+        node.prev.next = node.next;
```
nacenik, line 162: I'd recommend you to put this inside `else` block, `size--;` put after `else` and delete all `return` from this method. That will optimize your code.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,174 @@
...
+    }
+
+    private static class Node<T> {
+        private T item;
+        private Node<T> next;
+        private Node<T> prev;
+
+        Node(Node<T> prev, T element, Node<T> next) {
```
nacenik, line 170: don't forget about access modifiers

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,174 @@
...
+            throw new IndexOutOfBoundsException("Your index was " + index
+                    + ", but size is " + size);
+        }
+    }
+
+    private Node<T> getNode(T value) {
+        Node<T> node = head;
+        for (int i = 0; i < size; i++) {
+            if ((node.item == value) || (node.item != null && node.item.equals(value))) {
+                return node;
```
nacenik, line 98: instead of return node and make some additional check, you can call method unlink and after this return true here, otherwise return false. It will optimize your code

nacenik (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,167 @@
...
+                    + ", but size is " + size);
+        }
+    }
+
+    private void indexCheck(int index) {
+        if (index >= size || index < 0) {
+            throw new IndexOutOfBoundsException("Your index was " + index
+                    + ", but size is " + size);
+        }
+    }
```
petro-ushchuk, line 98: You can do only 1 method, think how.

petro-ushchuk (approved):
Good work! Check comment below and think how you can improve your code according to DRY. 🔥

PR #941, opened 2023-02-23
LinkedList hw impl

jaroslavcyk:
Implement own LinkedList with all Overrided method like add remove etc. Create my methode indexOnRange, unlink and getNodeByIndex

### DIFF
--- a/src/main/java/core/basesyntax/MyLinkedList.java
+++ b/src/main/java/core/basesyntax/MyLinkedList.java
@@ -3,45 +3,181 @@ package core.basesyntax;
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private static final String EXCEPTION_MESSAGE = "Index of the bounds. Check your input index";
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    private static class Node<T> {
+        private T value;
+        private Node<T> next;
+        private Node<T> prev;
+
+        Node(T value) {
+            this.prev = null;
+            this.value = value;
+            this.next = null;
+        }
+
+        Node(Node<T> prev, T value, Node<T> next) {
+            this.prev = prev;
+            this.value = value;
+            this.next = next;
+        }
+    }
+
     @Override
     public void add(T value) {
+        Node<T> node = new Node<>(value);
+        if (head == null) {
+            head = node;
+            tail = node;
+        } else {
+            node.prev = tail;
+            tail.next = node;
+            tail = node;
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        indexOnRange(index);
+        Node<T> newNode = new Node<>(value);
+        if (head == null) {
+            head = newNode;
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,181 @@
...
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException(EXCEPTION_MESSAGE);
+        }
+    }
+
+    private void indexOnRangeEqualsSize(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException(EXCEPTION_MESSAGE);
+        }
     }
```
petro-ushchuk, line 182: We can have 1 method, they are the same. Just pass `size` or `size + 1` as a parameter to your `indexOnRange`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,181 @@
...
+            currentNode = newNode;
+            size++;
+        }
+        tail = currentNode;
     }
 
     @Override
     public T get(int index) {
+        indexOnRangeEqualsSize(index);
+        Node<T> currentNode = getNodeByIndex(index);
```
petro-ushchuk, line 92: U already called `indexOnRangeEqualsSize` inside `getNodeByIndex`, line 91 useless

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,181 @@
...
+            return currentNode.value;
+        }
         return null;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        indexOnRangeEqualsSize(index);
+        Node<T> currentNode = getNodeByIndex(index);
```
petro-ushchuk, line 102: The same story as a comment in top.

petro-ushchuk (changes requested):
Veeeery good, but we should fix these small issues 😢

src/main/java/core/basesyntax/MyLinkedList.java
jaroslavcyk: i just add one if into the add() method, thats ok?

karma-o: I think that is acceptable since it is actually a only single use case

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,175 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void indexOnRangeEqualsSize(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException(EXCEPTION_MESSAGE);
```
karma-o, line 174: we should also add an index parameter to the exception message

karma-o (approved):
Left some minor comment

PR #946, opened 2023-03-03
Implemented methods:

ValeriyM250:
add(T value)
add(T value, int index)
public void addAll(List<T> list)
public T get(int index)
public T set(T value, int index)
public T remove(int index)
public boolean remove(T object)
public int size()
public boolean isEmpty()

### DIFF
--- a/src/main/java/core/basesyntax/MyLinkedList.java
+++ b/src/main/java/core/basesyntax/MyLinkedList.java
@@ -3,45 +3,140 @@ package core.basesyntax;
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+
+    private class Node<T> {
+        private Node<T> prev;
+        private T value;
+        private Node<T> next;
+
+        public Node(Node<T> prev, T value, Node<T> next) {
+            this.prev = prev;
+            this.value = value;
+            this.next = next;
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,140 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+
+    private class Node<T> {
```
andriy-shulepa, line 7: inner class should be at the bottom of the file

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,140 @@
...
+
+        public Node(Node<T> prev, T value, Node<T> next) {
+            this.prev = prev;
+            this.value = value;
+            this.next = next;
+        }
+    }
+
+    private static final int START_ARRAY_LENGTH = 10;
+    private Node [] elements;
```
andriy-shulepa, line 20: in a linked list elements shouldn't be saved in an array. Instead, the list should only know about the first node and the last node. other nodes should be accessed by iterating from the end of the list

andriy-shulepa (changes requested):

PR #958, opened 2023-03-19
jv-linked-list

Nataliia-Ladzoryshyn:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,225 @@
...
+            }
+            return temp;
+        } else if (index == size - 1) {
+            if (size == 0) {
+                return null;
+            } else if (size == 1) {
+                T temp = head.value;
+                head = tail = null;
+                size = 0;
+                return temp;
```
olekskov, line 130: redundant block, if size is 1 then you checkIndex will allow only index == 0 and this logic is already mentioned above

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,225 @@
...
+            } else {
+                Node<T> current = head;
+                for (int i = 0; i < size - 2; i++) {
+                    current = current.next;
+                }
+                final T temp = tail.value;
+                tail = current;
+                tail.next = null;
+                size--;
+                return temp;
```
olekskov, line 140: overcomputing, we don't need to go to tail from head through whole list, we just need to unlink last node and change tail:
```suggestion
                final T temp = tail.value;
                tail = tail.prev;
                tail.next = null;
                size--;
                return temp;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,225 @@
...
+            for (int i = 0; i < index; i++) {
+                previous = previous.next;
+            }
+            Node<T> current = previous.next;
+            previous.next = current;
+            current.prev = previous.prev;
+            previous.prev.next = current;
+            T temp = previous.value;
+            size--;
+            return temp;
```
olekskov, line 153: redundant variable:
```suggestion
            size--;
            return previous.value;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,225 @@
...
+            }
+        } else {
+            for (Node<T> x = head; x != null; x = x.next) {
+                if (object.equals(x.value)) {
+                    unlink(x);
+                    return true;
+                }
+            }
+        }
         return false;
```
olekskov, line 174: 1. please don't use 1-letter variable naming except obvious counter/index indicator
2. Two cycles may be combined:
```suggestion
        for (Node<T> currentNode = head; currentNode != null; currentNode = currentNode.next) {
            if (currentNode.value == object || 
                    currentNode.value != null && currentNode.value.equals(object)) {
                unlink(currentNode);
                return true;
            }
        }
        return false;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,225 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    static class Node<T> {
```
olekskov, line 187:
```suggestion
    private class Node<T> {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,225 @@
...
+        private Node<T> prev;
+
+        public Node(Node<T> prev, T value, Node<T> next) {
+            this.next = next;
+            this.value = value;
+            this.prev = prev;
+        }
+    }
+
+    public void unlink(Node<T> current) {
```
olekskov, line 199: This and also for all methods below
```suggestion
    private void unlink(Node<T> current) {
```

olekskov (changes requested):
Nice to see working solution, good job! But require some improvements

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,186 @@
...
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
     @Override
     public void add(T value) {
+        final Node<T> newTail = tail;
+        Node<T> newNode = new Node<>(tail, value, null);
+        tail = newNode;
```
uliana-andrushchuk, line 15: this variable is redundant, move `tail = newNode;` command after is/else construction and use tail itself instead of it's copy to new variable

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,186 @@
...
+        Object[] listArray = list.toArray();
+        final int numNew = listArray.length;
+        Node<T> pred;
+        Node<T> succ;
+        succ = null;
+        pred = tail;
+        for (Object object : listArray) {
+            T element = (T) object;
+            add(element);
+        }
```
uliana-andrushchuk, line 56:
```suggestion
        for (T element : list) {
            add(element);
        }
```
You can simplify this method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,186 @@
...
     public T set(T value, int index) {
-        return null;
+        checkIndex(index);
+        Node<T> current = getNodeByIndex(index);
+        T finalValue = current.value;
+        current.value = value;
+        return finalValue;
+    }
+
+    public Node<T> getNodeByIndex(int index) {
```
uliana-andrushchuk, line 74:
```suggestion
    private Node<T> getNodeByIndex(int index) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,186 @@
...
+
+    private void checkPositionIndex(int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException();
+        }
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException();
```
uliana-andrushchuk, line 183: Please, add some informative message to this exception

uliana-andrushchuk (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,178 @@
...
     public T set(T value, int index) {
-        return null;
+        checkIndex(index);
+        Node<T> current = getNodeByIndex(index);
+        T finalValue = current.value;
+        current.value = value;
+        return finalValue;
+    }
+
+    private Node<T> getNodeByIndex(int index) {
```
olekskov, line 66: private methods should be placed after public ones

olekskov (approved):
Approved, please pay attention to comment below for further tasks

PR #978, opened 2023-04-07
implemented some methods in my own LinkedList

pgolibroda:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,155 @@
...
+        } else {
+            currentNode = tail;
+            for (int i = size - 1; i > index; i--) {
+                currentNode = currentNode.prev;
+            }
+        }
+        return currentNode;
+    }
+
+    private void unLink(Node<T> node) {
```
okuzan, line 143:
```suggestion
    private void unlink(Node<T> node) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,155 @@
...
+            this.next = next;
+        }
+    }
+
+    private void checkIndexForAddIndex(int index) {
+        if (index >= 0 && index <= size) {
+            return;
+        }
+        throw new IndexOutOfBoundsException("Cannot add element , because index "
+                + index + " is invalid");
```
okuzan, line 106: invert the condition and place throwing inside an if block

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,155 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private static class Node<T> {
```
okuzan, line 89: place inner class at the very end of the file

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,155 @@
...
+        return removeValue;
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> node = head;
+        while (node != null) {
+            if (node.item == object || node.item != null && node.item.equals(object)) {
+                unLink(node);
+                size--;
```
okuzan, line 71: include `size--;` to unlink itself

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,155 @@
...
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        Node<T> newNode = findByIndex(index);
+        T removeValue = newNode.item;
+        unLink(newNode);
+        size--;
+        return removeValue;
```
okuzan, line 62:
```suggestion
        return unlink(findByIndex(index));
```
make unlink return T node instead of it being void and this method will be as simple as:

okuzan (changes requested):
Good job so far!

Ivan95kos (approved):
Great job

PR #979, opened 2023-04-12
all test passed

sshestaka:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,195 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size;
+    private Node<T> first;
+    private Node<T> last;
+
+    private static class Node<T> {
```
Ivan95kos, line 10: class should not be a static
```suggestion
    private class Node<T> {
```

Ivan95kos, line 10: All inner classes should be at the end of class

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,195 @@
...
+            Node<T> newNode = new Node<>(last, value, null);
+            last.next = newNode;
+            last = newNode;
+            size++;
+        }
     }
 
     @Override
     public void add(T value, int index) {
+        //add first
```
Ivan95kos, line 36: please, remove all comments 
code should be readable withoud comments
```suggestion
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,195 @@
...
+            oldNodePrev.next = newNode;
+            size++;
+        }
     }
 
     @Override
     public void addAll(List<T> list) {
+        Object[] arrayList = list.toArray();
+        int length = arrayList.length;
+        if (length == 0) {
```
Ivan95kos, line 56: do not create redundant variable
```suggestion
        if (arrayList.length == 0) {
```

Ivan95kos (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
sshestaka: OK

sshestaka: But in LinkedList.java this class is static...

sshestaka: OK

src/main/java/core/basesyntax/MyLinkedList.java
sshestaka: OK. But why? With comments it's easier for me and for my colleagues to understand what each methods does ...

src/main/java/core/basesyntax/MyLinkedList.java
sshestaka: Yes, I understand it. But, I have done as in the LinkedList.java class (please take a look on photo below). I think they created a redundant variable for better clarity and to make the "if condition" more shorter ...

sshestaka: image

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,191 @@
...
     }
 
     @Override
     public void add(T value, int index) {
+        if (index == 0) {
+            addFirst(value);
+        } else if (index == size) {
+            addLast(value);
+        } else {
+            checkAddIndex(index);
```
uliana-andrushchuk, line 29: Please, use method `checkIndexIsValid` here
and then you can remove `checkAddIndex` method

src/main/java/core/basesyntax/MyLinkedList.java
uliana-andrushchuk: In this method variable `numNew` used 2 times in LinkedList.java class

uliana-andrushchuk (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
sshestaka: Yes, you are right! I did not see this after all the manipulations with "add"

kozhukhovsky (approved):

PR #980, opened 2023-04-14
MyLinkedList implemented

TymurProkhorov:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,188 @@
...
+    @Override
+    public String toString() {
+        StringBuilder a = new StringBuilder();
+        Node<T> currentNode = head;
+        while (currentNode != null) {
+            a.append(currentNode.value).append(", ");
+            currentNode = currentNode.next;
+        }
+        return a.substring(0, a.length() - 2);
+    }
```
kozhukhovsky, line 105: dont need this method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,188 @@
...
+            }
+            return node;
+        } else {
+            Node<T> node = tail;
+            for (int i = size - 1; i > index; i--) {
+                node = node.prev;
+            }
+            return node;
+        }
+    }
```
kozhukhovsky, line 177: I don’t understand why we need the second method, if we can combine it into one. And the very uninformative name "node"

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,188 @@
...
 
     @Override
     public boolean remove(T object) {
+        if (object == null) {
+            for (Node<T> currentNode = head; currentNode != null; currentNode = currentNode.next) {
+                if (currentNode.value == null) {
+                    unlink(currentNode);
+                    return true;
+                }
+            }
```
kozhukhovsky, line 74: We don't need this block, just add a condition in the second part

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,188 @@
...
+        } else {
+            for (Node<T> currentNode = head; currentNode != null; currentNode = currentNode.next) {
+                if (currentNode.value != null && currentNode.value.equals(object)) {
```
kozhukhovsky, line 77:
```suggestion
                if ((currentNode.value != null && currentNode.value.equals(object)) || currentNode.value == object) {
```

kozhukhovsky (changes requested):

Ivan95kos (approved):
Great job

PR #981, opened 2023-04-15
task jv_linked_list solution

ArtyomK75:


src/main/java/core/basesyntax/ListNode.java
```diff
@@ -0,0 +1,35 @@
+package core.basesyntax;
+
+class ListNode<T> {
```
Ivan95kos, line 3: Знімок екрана 2023-04-15 о 13 08 30

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,179 @@
...
+
+    public MyLinkedList() {
+
+    }
+
+    public MyLinkedList(T value) {
+        head = new ListNode<>(value);
+        tail = head;
+        size = 1;
+    }
```
Ivan95kos, line 19: useless constructors
```suggestion
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,179 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private static final String INDEX_OUT_OF_BOUNDS = "Index is out of bounds of list";
+    private ListNode<T> head;
+    private ListNode<T> tail;
+    private int size = 0;
```
Ivan95kos, line 9: int has default value 0
```suggestion
    private int size;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,179 @@
...
     }
 
     @Override
     public void addAll(List<T> list) {
+        for (T value : list) {
+            add(value);
+        }
+    }
+
+    private ListNode<T> getListNode(int index) {
```
Ivan95kos, line 75: Private methods and classes should be after public ones in your class.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,179 @@
...
+            nextNode = nextNode.getNext();
+            if (nextNode == null) {
+                break;
+            }
+        }
+        if (nextNode == null) {
+            return false;
+        }
+        removeNode(nextNode);
+        return true;
```
Ivan95kos, line 169: Create a separate method unlink(Node node).

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,179 @@
...
     public void addAll(List<T> list) {
+        for (T value : list) {
+            add(value);
+        }
+    }
+
+    private ListNode<T> getListNode(int index) {
+        if (index > size || index < 0) {
+            throw new IndexOutOfBoundsException(INDEX_OUT_OF_BOUNDS);
+        }
```
Ivan95kos, line 78: move to private method `checkIndex(int index) {}`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,179 @@
...
+
+        if (size == 0 || index == size) {
+            return null;
+        }
```
Ivan95kos, line 82: you should not return null, it is bad practice
```suggestion
```

Ivan95kos (changes requested):
Please, pay attention to common mistakes
so hard to read try to do methods more simply

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,190 @@
...
+            tail = tail.next;
+            size++;
+        }
     }
 
     @Override
     public void add(T value, int index) {
+        ListNode<T> node = getListNode(index);
+        ListNode<T> newNode = new ListNode<>(value);
+        boolean isAfterTailInsert = false;
```
Ivan95kos, line 39: Try to refactor this method `add(T value, int index)` without this variable `isAfterTailInsert`

You can use this example:
```
    @Override
        checkIndex(index);
        if(index == size) {
            add(value);
        } else {
            Node<T> currentNode = getCurrentNodeBy(index);
            Node<T> newNode = new Node<>(currentNode.previous, value, currentNode);

            if (head == currentNode) {
                head = newNode;
            } else {
                currentNode.previous.next = newNode;
            }
            currentNode.previous = newNode;
            size++;
        }
    }
```

Ivan95kos (changes requested):

Ivan95kos (approved):
Nice job

PR #985, opened 2023-04-23
solution hw12

vrnvv:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,145 @@
...
     }
-
+    
     @Override
     public boolean remove(T object) {
-        return false;
+        Node<T> node = headNode;
+        if (currentSize == 0) {
+            return false;
+        } else {
```
olekskov, line 83: Consider removing the `else` statement, as it is unnecessary. The code inside the else block can be executed directly after checking `if (currentSize == 0)`.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,145 @@
...
+        checkIndex(index);
+        return findNodeByIndex(index).item;
     }
-
+    
     @Override
     public T set(T value, int index) {
+        checkIndex(index);
+        Node<T> newNode = findNodeByIndex(index);
+        if (newNode != null) {
```
olekskov, line 58: Consider removing the check for `newNode != null`, since the method `findNodeByIndex(index)` always returns a valid node for the given index. This makes the if condition redundant.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,145 @@
...
+        }
         return null;
     }
-
+    
     @Override
     public T remove(int index) {
+        checkIndex(index);
+        Node<T> node = findNodeByIndex(index);
+        if (node != null) {
```
olekskov, line 70: Consider removing the check for `node != null`, since the method `findNodeByIndex(index)` always returns a valid node for the given index. This makes the if condition redundant.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,145 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> headNode;
+    private Node<T> tailNode;
+    private int currentSize;
```
olekskov, line 8: I can't see reason to make more complicated name. We don't have other sizes here so can't really mismatch them
```suggestion
    private int size;
```

olekskov (changes requested):

olekskov (approved):
Good job!

PR #987, opened 2023-04-25
completed task

sZlishchev:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,182 @@
...
     }
 
     @Override
     public void add(T value, int index) {
+        checkRangeForAdd(index);
+        Node<T> newNode;
+        if (head == null) {
+            addFirst(value);
+        } else if (index == 0) {
+            newNode = new Node<>(null, value, head);
```
rovein, line 40: Let's create another second overloaded constructor, which will accept the (T element, Node<T> next). Use it here and in all other occurrences.

rovein (approved):
Good job! But see one minor comment.

PR #991, opened 2023-05-02
Performed my own implementation the LinkedList

V-Solianyk:


Gamemod13 (approved):
Good work! I haven`t any ideas, what's can be wrong in that code, maybe a little bit decompositions, but as for me - it's not nessessory)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
...
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    private static class Node<T> {
+        private Node<T> next;
+        private T value;
+        private Node<T> previous;
```
maxline, line 13:
```suggestion
        private T value;
        private Node<T> next;
        private Node<T> previous;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    private static class Node<T> {
```
maxline, line 10: please move `private` class to the end of the file, according to ordering recommendations

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
...
+        }
+    }
+
+    private void checkIndex(int index) {
+        if (index > size || index < 0) {
+            trowException();
+        }
+    }
+
+    private void trowException() {
```
maxline, line 138: this is a bad practice to create a separate method that only throwing an exception

maxline (changes requested):
good job, let's improve the solution

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
...
+        checkByIndexEqualSize(index);
+        Node<T> deletedNode = nodeSearch(index);
+        if (index == 0) {
+            head = head.next;
+        } else if (index == size - 1) {
+            tail = tail.previous;
+        } else {
+            deletedNode.previous.next = deletedNode.next;
+            deletedNode.next.previous = deletedNode.previous;
+        }
```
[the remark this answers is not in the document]
V-Solianyk: Thanks for not paying attention to the obvious

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,146 @@
...
+            addTail(newNode);
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index > size || index < 0) {
+            throw new IndexOutOfBoundsException("An incorrect index was entered - " + index);
+        }
```
maxline, line 25: this check is redundant, we have check `index == size` at line 29, and `if (index >= size || index < 0)`  in method `checkIndex()`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,146 @@
...
+        } else if (index == size) {
+            addTail(newNode);
+        } else if (index == 0) {
+            Node<T> newHead = new Node<>(head, value, null);
+            head.previous = newHead;
+            head = newHead;
+        } else {
+            Node<T> nodeByIndex = searchNode(index);
+            Node<T> previousNode = nodeByIndex.previous;
+            Node<T> addedMiddleNode = new Node<>(nodeByIndex, value, previousNode);
```
maxline, line 38:
```suggestion
            Node<T> newNode = new Node<>(nodeByIndex, value, previousNode);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,146 @@
...
+        oldNode.value = value;
+        return oldValue;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        Node<T> deletedNode = searchNode(index);
+        unlink(deletedNode);
+        size--;
```
maxline, line 69: move `size--` to `unlink()` method to avoid code duplication, check other places in the code please

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,146 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private static class Node<T> {
```
maxline, line 98: this inner class declaration should be moved to the end of the class, after the private  methods

maxline (changes requested):
Well done! We are close to the final version of the solution, I left a several minor comments, please take a look.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,148 @@
...
+    }
+
+    private boolean addFirst(Node<T> newNode) {
+        if (size == 0) {
+            head = newNode;
+            tail = newNode;
+            return false;
+        }
+        return true;
+    }
```
maxline, line 125: please refactor in this way:
```suggestion
    private void addFirst(Node<T> newNode) {
            head = newNode;
            tail = newNode;
    }
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,148 @@
...
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        Node<T> newNode = new Node<>(null, value, null);
+        if (index == 0) {
+            addFirstElement(value, newNode);
+        } else if (index == size) {
+            addTail(newNode);
```
maxline, line 25: move this `if (index == size)` block on the first place (before `if (index ==0)`. in this way:
```
if (index == size) {
      add(value);
      return;
}
```
this help to simplify `if else `constructions and help to avoid additional check that list is not empty when we put new element to `index = 0`

maxline (changes requested):
The solution was close to finish, but you added a piece of code that is not good enough, so I left a comments to refactor the code in more proper way. Take a look please

PR #992, opened 2023-05-02
Fix common mistakes. Implements LinkedList

Gamemod13:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,170 @@
...
+        for (T item : list) {
+            add(item);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException();
```
V-Solianyk, line 49: can be brought into a single method to check the index and throw an error

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,170 @@
...
+        }
+        return getNode(index).value;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException();
+        }
```
V-Solianyk, line 58: can be brought into a single method to check the index and throw an error
code duplication

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,170 @@
...
+        T oldValue = current.value;
+        current.value = value;
+        return oldValue;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException();
```
V-Solianyk, line 68: code duplication the same situation as above

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,170 @@
...
+    }
+
     @Override
     public void add(T value) {
+        addLast(value);
     }
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
```
V-Solianyk, line 23: you can add informativeness to the text of the error in the method

V-Solianyk (changes requested):

V-Solianyk (approved):
great job, but I realized that I gave you advice to throw an error in a separate method, although this is a bad practice from my own experience, I think the mentor will correct this point

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,181 @@
...
+        addLast(value);
     }
 
     @Override
     public void add(T value, int index) {
+        checkIndex(index);
+        if (index == size) {
+            addLast(value);
+        } else if (index == 0) {
+            addFirst(value);
```
maxline, line 27: I think this method is redundant because it called only one time, move the body from the method `addFirst()` to here.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,181 @@
...
+    }
+
     @Override
     public void add(T value) {
+        addLast(value);
     }
 
     @Override
     public void add(T value, int index) {
+        checkIndex(index);
```
maxline, line 23: this method is redundant, just add `checkIndexOutOfBoundsInclusiveSize()` inside of method  `getNode(index);`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,181 @@
...
+        }
+    }
+
+    private void checkIndexOutOfBoundsInclusiveSize(int index) {
+        if (index < 0 || index >= size) {
+            throwIndexOutOfBoundsEx(index);
+        }
+    }
+
+    private void throwIndexOutOfBoundsEx(int index) {
```
maxline, line 165: method that contains only throwing exception is a bad practice. please remove and move the code to the `checkIndexOutOfBoundsInclusiveSize()` method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,181 @@
...
+        if (next == null) {
+            tail = prev;
+        } else {
+            next.prev = prev;
+            node.next = null;
+        }
+        size--;
+    }
+
+    private void checkIndex(int index) {
```
maxline, line 153: this method is redundant

maxline (changes requested):
Very good job! I left just a several comments, please take a look.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
...
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    public MyLinkedList() {
+        head = null;
+        tail = null;
+        size = 0;
```
sarakhmen, line 13: these fields are actually initialized this way by default so you don't need to make it explicitly

sarakhmen (approved):
Good job :)

sarakhmen (approved):
Good job :)

PR #993, opened 2023-05-03
Solved

Poloutinho:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,202 @@
...
+            curNode = curNode.prev;
+            counter--;
+        }
+        return null;
+    }
+
+    private Node<T> findNodeByElement(T element) {
+        Node<T> curNode = head;
+        while (curNode != null) {
+            if ((curNode.value != null && curNode.value.equals(element))
```
sZlishchev, line 169: Let's create separate method for this logic

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,202 @@
...
+
     @Override
     public void add(T value) {
+        Node<T> node = new Node<>(value, tail, null);
+        if (head == null) {
+            head = node;
+            tail = head;
+        } else {
+            tail.next = node;
+            tail = node;
```
sZlishchev, line 18: Here use the same logic like in addToTail method. You can use that method here, but don't forget that you increase size inside addToTail method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,202 @@
...
+        }
         return false;
     }
+
+    private static class Node<T> {
+        private T value;
+        private Node<T> next;
+        private Node<T> prev;
+
+        private Node(T value, Node<T> prev, Node<T> next) {
```
sZlishchev, line 198: Change the sequence of params for example (prev, value, next) it's more informative

sZlishchev (changes requested):
Good job! But let's improve your solution))

sZlishchev (approved):
Great job! Keep pushing))

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,205 @@
...
+
+    private Node<T> findNodeByIndex(int index) {
+        isValidIndexForAdd(index);
+        if (index == 0) {
+            return head;
+        } else if (index == size - 1) {
+            return tail;
+        }
+        if (index < size / 2) {
+            return startFromHead(index);
```
maxline, line 134: method naming, it is not clear from the method name that it is find and return the node.
the same remark for the method `startFromTail()`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,205 @@
...
 
     @Override
     public T remove(int index) {
-        return null;
+        Node<T> deletedElement = findNodeByIndex(index);
+        if (deletedElement == null) {
+            throw new IndexOutOfBoundsException("wrond index: " + index + ", for empty list");
+        }
+        unlink(deletedElement);
+        size--;
```
maxline, line 69: size-- can be moved inside unlink() method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,205 @@
...
+        return oldValue;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        Node<T> deletedElement = findNodeByIndex(index);
+        if (deletedElement == null) {
+            throw new IndexOutOfBoundsException("wrond index: " + index + ", for empty list");
+        }
```
maxline, line 67: redundant `if` clause, we already check index ant throws exception inside `isValid..()` method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,205 @@
...
+        return deletedElement.value;
     }
 
     @Override
     public boolean remove(T object) {
-        return false;
+        Node<T> deletedElement = findNodeByElement(object);
+        boolean isDeleted = unlink(deletedElement);
+        if (isDeleted) {
+            size--;
```
maxline, line 78: size-- can be moved inside unlink() method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,204 @@
...
+        if (index > size || index < 0) {
+            throw new IndexOutOfBoundsException("invalid index: " + index);
+        }
+    }
+
+    private void isValidIndexForSetAndGet(int index) {
+        if (index >= size || index < 0) {
+            throw new IndexOutOfBoundsException("invalid index: " + index);
+        }
+    }
```
sarakhmen, line 120: one of these methods (for add) is redundant and could be removed. Another one should be renamed. For example, you could use 'checkIndex' or 'validateIndex'

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,205 @@
...
     public void add(T value, int index) {
+        isValidIndexForAdd(index);
+        if (isEmpty()) {
+            add(value);
+        } else if (index == 0) {
+            addToHead(value);
+        } else {
+            Node<T> previous = findNodeByIndex(index);
+            if (previous == null) {
+                addToTail(value);
```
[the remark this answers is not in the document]
sarakhmen: not fixed, please pay attention to the @maxline comment

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,204 @@
...
+        } else {
+            tail.next = node;
+            tail = node;
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        isValidIndexForAdd(index);
```
sarakhmen, line 25:
```suggestion
```
this check is redundant here as all necessary validations will be performed later in this method or withing your findNodeByIndex

sarakhmen (changes requested):
Let's improve your solution. Please pay attention to the previous mentor's comments

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,204 @@
...
+        }
+    }
+
+    private Node<T> findNodeByIndex(int index) {
+        isValidIndexForAdd(index);
+        if (index == 0) {
+            return head;
+        } else if (index == size - 1) {
+            return tail;
+        }
```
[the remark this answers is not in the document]
Poloutinho: The if-else statements in the findNodeByIndex() method are necessary to handle this cases:

If the index is 0, return the head node.
If the index is size - 1, return the tail node.

PR #994, opened 2023-05-03
task solution implemented

Tupadzaky:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,172 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> last;
+    private Node<T> first;
+    private int size;
+
+    public MyLinkedList() {
+    }
```
V-Solianyk, line 11: Why do you need a default constructor?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,172 @@
...
+
+    private void linkLast(T value) {
+        Node<T> l = last;
+        Node<T> newNode = new Node<>(l, value, null);
+        last = newNode;
+        if (l == null) {
+            first = newNode;
+        } else {
+            l.next = newNode;
```
V-Solianyk, line 105: variables with only one letter are bad practice

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,172 @@
...
+        }
+        if (next == null) {
+            last = prev;
+        } else {
+            next.prev = prev;
+            target.next = null;
+        }
+        target.item = null;
+        size--;
+        return element;
```
V-Solianyk, line 140: definitely, this code works correctly, but I would simplify it because it is difficult to read, but I think in this case the mentor will give better advice

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,172 @@
...
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private static class Node<T> {
+        private Node<T> prev;
+        private T item;
+        private Node<T> next;
```
V-Solianyk, line 89: from my experience had a similar error, this class should go after private methods

V-Solianyk (changes requested):
It's a great job, I would correct some things, and you can leave some for the metronome to consider.

src/main/java/core/basesyntax/MyLinkedList.java
V-Solianyk: private T unlink(Node<T> target) {
        final T element = target.item;
        final Node<T> prev = target.prev;
        final Node<T> next = target.next;
        if (prev == null) {
            first = next;
        } else {
            prev.next = next;
            target.prev = null;
        }
        if (next == null) {
            last = prev;
        } else {
            next.prev = prev;
            target.next = null;
        }
        target.item = null;
        size--;
        return element;
    }

V-Solianyk: I meant this code, which I wrote in the content, because I highlighted a little bit of the wrong area during the review

V-Solianyk (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
+        } else {
+            next.prev = prev;
+            target.next = null;
+        }
+        target.item = null;
+        size--;
+        return element;
+    }
+
+    private Node<T> node(int index) {
```
sarakhmen, line 128: this method purpose isn't clear enough, let's rename it to something like 'getNode' or 'findNode'

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void linkLast(T value) {
+        Node<T> oldLast = last;
+        Node<T> newNode = new Node<>(oldLast, value, null);
```
sarakhmen, line 85: instead of passing null values to the constructor, I'd recommend overriding it

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
     public void addAll(List<T> list) {
+        for (T element : list) {
+            linkLast(element);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkElementIndex(index);
```
sarakhmen, line 34: your checkElementIndex method usually is called before node() method, so maybe it's better to move this check to the node() method

sarakhmen (changes requested):
Good job but let's slightly improve your solution :)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,169 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> first;
+    private Node<T> last;
+    private int size;
+
     @Override
     public void add(T value) {
+        linkLast(value);
```
maxline, line 12: ` linkLast(value)` method is redundant. just move code here

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,169 @@
...
     }
 
     @Override
     public void add(T value, int index) {
+        checkPositionIndex(index);
+        if (index == size) {
+            linkLast(value);
```
maxline, line 19:
```suggestion
          add(value);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,169 @@
...
+        } else {
+            linkBefore(value, getNode(index));
+        }
     }
 
     @Override
     public void addAll(List<T> list) {
+        for (T element : list) {
+            linkLast(element);
```
maxline, line 28:
```suggestion
            add(value);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,169 @@
...
 
     @Override
     public T remove(int index) {
-        return null;
+        return unlink(getNode(index));
     }
 
     @Override
     public boolean remove(T object) {
+        if (object == null) {
```
maxline, line 52: instead of this `if` let's implement the method:
```
    }
```    
and use it for comparing the node inside the loop

maxline (changes requested):
Good job! It is close to finish, left a few remarks that can be easily fixed.

sarakhmen (approved):
Good job :)

PR #1003, opened 2023-05-07
LinkedList has been implemented

m-pnsk:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,177 @@
...
+
+        private Node(Node<T> prev, T value) {
+            this.prev = prev;
+            this.value = value;
+        }
+
+        private Node(T value, Node<T> next) {
+            this.value = value;
+            this.next = next;
+        }
```
chrisnyrka, line 177: As a common mistake, when creating a new Node, pass all the fields to Node constructor

chrisnyrka (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
m-pnsk: I disagree. It is a normal practice to overload the constructor. In order not to hardcode the null value directly in the code, the following constructors are used that set the null value by default.

chrisnyrka (approved):

PR #1007, opened 2023-05-07
implemented task

mpankrad:


OlehPryshchepa (changes requested):
good job, but let`s improve

OlehPryshchepa (approved):
good job

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,177 @@
...
+        Node<T> nodeAfter = toRemove.next;
+        if (nodeBefore == null) {
+            first = nodeAfter;
+        }
+        if (nodeAfter == null) {
+            last = nodeBefore;
+        } else {
+            nodeAfter.prev = nodeBefore;
+        }
+        size--;
```
maxline, line 75: please create  a separate method unlink(Node node) for this code. you already have the method removeNode(). https://mate-academy.github.io/jv-program-common-mistakes/java-core/collections/linked-list.html#create-a-separate-method-unlinknode-node

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,177 @@
...
+
+    private void addMiddle(T value, int index) {
+        Node<T> current = getNode(index);
+        Node<T> newNode = new Node<>(current.prev, value, current);
+        current.prev.next = newNode;
+        current.prev = newNode;
+        size++;
+    }
+
+    private boolean removeNode(Node<T> node) {
```
maxline, line 165:
```suggestion
    private boolean unlink(Node<T> node) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,177 @@
...
+        }
+        size--;
+        return toRemove.value;
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> currentNode = first;
+        while (currentNode != null) {
+            if (currentNode.value == null) {
```
maxline, line 83: instead of check for null here, let's create a new method and call it in the loop: 
```
    }
```

maxline (changes requested):
Good job! Let's improve the solution

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,168 @@
...
     @Override
     public int size() {
-        return 0;
+        return size;
     }
 
-    @Override
-    public boolean isEmpty() {
-        return false;
+    @Override public boolean isEmpty() {
```
sarakhmen, line 74:
```suggestion
    @Override 
    public boolean isEmpty() {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,168 @@
...
     public T get(int index) {
-        return null;
+        return getNode(index).value;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        checkRange(index);
+        Node<T> nodeAtIndex = getNode(index);
```
sarakhmen, line 42: chechRange() method is called twice: here and within getNode() method. So I'd recommend removing checkRange call from this method. Let's fix the same problem in all places

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,168 @@
...
+
+    private void checkRangeAdd(int index) {
+        if (index < 0 || index > size + 1) {
+            throw new IndexOutOfBoundsException("Index of elements: " + index
+                    + " Size of the list: " + size);
+        }
+    }
+
+    private boolean areEqual(T a, T b) {
+        return (a == null && b == null) || (a != null && a.equals(b));
```
sarakhmen, line 93:
```suggestion
        return a == b || a != null && a.equals(b);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,168 @@
...
+                    + "Size of the list: " + size);
+        }
+    }
+
+    private void checkRangeAdd(int index) {
+        if (index < 0 || index > size + 1) {
+            throw new IndexOutOfBoundsException("Index of elements: " + index
+                    + " Size of the list: " + size);
+        }
+    }
```
sarakhmen, line 90: I think you can remove this method as it's actually not necessary. Also remove its call from add() method

sarakhmen (changes requested):
Let's improve your solution a bit

sarakhmen (approved):
Good job ;)

PR #1009, opened 2023-05-08
Homework: JV Linked List

ViktorBykh:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,180 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void checkRangeForAdd(int index) {
```
DV7Art, line 133: I think, that  u need to rename this method cause u don't u this check only for "add"  methods.
```suggestion
    private void checkRange(int index) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,180 @@
...
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size = 0;
+    private Node<T> firstNode;
+    private Node<T> lastNode;
+
+    public MyLinkedList() {
+        firstNode = null;
+        lastNode = null;
+    }
```
DV7Art, line 13: these fields are actually initialized this way by default so you don't need to make it explicitly

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,175 @@
...
+    private int size = 0;
+    private Node<T> firstNode;
+    private Node<T> lastNode;
+
+    class Node<T> {
+        private T value;
+        private Node<T> prev;
+        private Node<T> next;
+
+        public Node(Node<T> prev, T value, Node<T> next) {
```
maxline, line 15: `private Node(...`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,175 @@
...
     public T get(int index) {
-        return null;
+        checkRange(index);
+        return getNode(index).value;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        checkRange(index);
```
maxline, line 70: move `checkRange(index);` inside of the `getNode()` method to avoid code duplication.
check all code occurrences please

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,175 @@
...
+        } else {
+            currentNode.prev.next = currentNode.next;
+        }
+        if (currentNode.next == null) {
+            lastNode = currentNode.prev;
+        } else {
+            currentNode.next.prev = currentNode.prev;
+        }
+        size--;
+        return currentNode.value;
```
maxline, line 92: Create a separate method `unlink(Node node)` according the common mistakes. Move this code to the `unlink() ` method. Do not duplicate the code in `remove(int index)`  and  `remove(T object)`  methods.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,175 @@
...
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> currentNode = firstNode;
+        while (currentNode != null) {
+            if (elementsAreEqual(object, currentNode)) {
```
maxline, line 99:
```suggestion
            if (elementsAreEqual(object, currentNode.value)) {
```

maxline (changes requested):
good job! let's improve the solution

sarakhmen (changes requested):
Let's slightly improve your solution :)

sarakhmen (approved):
Good job ;)

PR #1014, opened 2023-05-08
Added task realization

pkobevko:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
...
+        private Node<T> next;
+        private Node<T> prev;
+
+        Node(Node<T> prev, T value, Node<T> next) {
+            this.value = value;
+            this.next = next;
+            this.prev = prev;
+        }
     }
 }
```
volodymyrruso, line 166: file doesn't end with new line

volodymyrruso (changes requested):
good job, just add one new line at the end

src/main/java/core/basesyntax/MyLinkedList.java
pkobevko: probably github automatically trims the last empty line, because it is visible in intellij idea, and if it was not there, the tests would not pass
image

volodymyrruso (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> first;
+    private Node<T> last;
+    private int size;
+
     @Override
     public void add(T value) {
+        Node<T> node = new Node<>(null, value, null);
```
sarakhmen, line 12: let's pass your 'last' field to the constructor here directly

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
...
+            first = node;
+            size++;
+            return;
+        }
+        Node<T> nextNode = getNode(index);
+        nextNode.prev.next = node;
+        node.prev = nextNode.prev;
+        nextNode.prev = node;
+        node.next = nextNode;
+        size++;
```
sarakhmen, line 43: Let's slightly refactor your code here. The second 'if' condition could be placed right after first one. verifyIndexIsValid could be removed as it's called within getNode() method. And let's pass all the required arguments to the Node constructor

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
...
     public T get(int index) {
-        return null;
+        verifyIndexIsValid(index);
+        return getNode(index).value;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        verifyIndexIsValid(index);
```
sarakhmen, line 61: code duplication as verifyIndexIsValid is called inside getNode(). Please fix in all places

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
...
+    private boolean verifyIndexIsValid(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException(
+                    String.format("Invalid index passed: %d for size: %d", index, size));
+        }
+        return true;
+    }
+
+    private boolean elementsAreEqual(T first, T last) {
+        return (first == last) || ((first != null) && (first.equals(last)));
```
sarakhmen, line 152:
```suggestion
        return first == last || first != null && first.equals(last);
```

sarakhmen (changes requested):
Good job but let's improve your solution :)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,156 @@
...
+        node.value = value;
+        return oldValue;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        Node<T> nodeToRemove = getNode(index);
+        unlink(nodeToRemove);
+        size--;
```
maxline, line 65: move size-- inside the method unlink()

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,156 @@
...
+    private boolean elementsAreEqual(T first, T last) {
+        return first == last || first != null && first.equals(last);
+    }
+
+    private static class Node<T> {
+        private T value;
+        private Node<T> next;
+        private Node<T> prev;
+
+        Node(Node<T> prev, T value, Node<T> next) {
```
maxline, line 152: make it private

maxline (approved):
Well done! Approved with a few small remarks, please take a look for the future.

PR #1016, opened 2023-05-08
create own implementation of linked list

ArinaPolna:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,152 @@
...
+
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
     @Override
     public void add(T value) {
+        Node node = new Node<>(null, value, null);
+        if (size == 0) {
+            head = tail = node;
```
m-pnsk, line 15: You have the isEmpty() method. It is better to use it here

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,152 @@
...
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        checkIndex(index);
+        Node node = new Node<>(null, value, null);
+        if (index == size) {
+            add(value);
+            return;
```
m-pnsk, line 30: "return" redundant
```suggestion
            add(value);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,152 @@
...
+            current.prev = node;
+        }
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
```
m-pnsk, line 49: Good solution, but you can do better with a for-loop if you wish
```suggestion
        for (T listElement : list) {
            add(listElement);
        }
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,152 @@
...
+        if (current.equals(head)) {
+            head = current.next;
+        } else if (current.equals(tail)) {
+            tail = current.prev;
+        } else {
+            current.prev.next = current.next;
+            current.next.prev = current.prev;
+        }
+        size--;
+    }
```
m-pnsk, line 135: After deletion, your new head (or tail) continues to reference the old head (or tail). `head.prev` must always be `null`. (`tail.next` must always be `null`)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,152 @@
...
+        while (index != size) {
+            if ((currentNode.item == object)
+                    || (currentNode.item != null && currentNode.item.equals(object))) {
+                return currentNode;
+            }
+            currentNode = currentNode.next;
+            index++;
+        }
+        return null;
+    }
```
m-pnsk, line 123: How do you think, this solution is better or not?
```suggestion
    private Node<T> findNodeByObject(T object) {
          for (Node<T> currentNode = head; currentNode != null; currentNode = currentNode.next) {
            if ((currentNode.item == object)
                    || (currentNode.item != null && currentNode.item.equals(object))) {
                return currentNode;
            }
        }
        return null;
    }
```

m-pnsk (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,152 @@
...
     }
 
     @Override
     public void add(T value, int index) {
+        checkIndex(index);
+        Node node = new Node<>(null, value, null);
+        if (index == size) {
+            add(value);
+            return;
+        } else if (index == 0) {
```
maxline, line 31: because of `return` in line 30, we can get rid of `else` clause:
```suggestion
        if (index == 0) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,152 @@
...
+        for (Node<T> currentNode = head; currentNode != null; currentNode = currentNode.next) {
+            if ((currentNode.item == object)
+                    || (currentNode.item != null && currentNode.item.equals(object))) {
+                return currentNode;
+            }
+        }
+        return null;
+    }
+
+    private void remover(Node<T> current) {
```
maxline, line 121: rename this method as `unlink()` according to common mistakes suggestions

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,152 @@
...
+            for (int i = size - 1; i > index; i--) {
+                currentElement = currentElement.prev;
+            }
+        }
+        return currentElement;
+    }
+
+    private Node<T> findNodeByObject(T object) {
+        for (Node<T> currentNode = head; currentNode != null; currentNode = currentNode.next) {
+            if ((currentNode.item == object)
```
maxline, line 113: extact a comparing nodes to the new method, e.g.:
```
    }
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,152 @@
...
+            node.prev = tail;
+            tail.next = node;
+            tail = node;
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        checkIndex(index);
```
maxline, line 26: this method is redundant and can be removed, you check index == size in line 28, the rest of the range is checked in `findNodeByIndex()`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,152 @@
...
+                return currentNode;
+            }
+        }
+        return null;
+    }
+
+    private void remover(Node<T> current) {
+        if (current.prev == null && current.next == null) {
+            head = null;
+        } else if (current.equals(head)) {
```
maxline, line 124: let's use `==` instead of `equals` here, we need to be sure that nodes have the same link. or simple check `current.prev == null`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,152 @@
...
+            head = current.next;
+            head.prev = null;
+        } else if (current.equals(tail)) {
```
maxline, line 127: see the previous comment

maxline (changes requested):
good job! let's improve the solution

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,149 @@
...
+        }
+        size--;
+    }
+
+    private static class Node<T> {
+        private T item;
+        private Node<T> prev;
+        private Node<T> next;
+
+        public Node(Node<T> prev, T item, Node<T> next) {
```
maxline, line 145: private

maxline (approved):
Well done! Approved with one minor remark take a look for the future.

PR #1022, opened 2023-05-09
Created own LinkedList class which behave same as java.util.LinkedList.

ArtEg0:
Also added own methods for inner logic of LinkedList.

### DIFF
--- a/src/main/java/core/basesyntax/MyLinkedList.java
+++ b/src/main/java/core/basesyntax/MyLinkedList.java
@@ -1,47 +1,157 @@
 package core.basesyntax;
 
 import java.util.List;
+import java.util.Objects;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    private static class Node<T> {
+        private T value;
+        private Node<T> prev;
+        private Node<T> next;
+
+        public Node(T value) {
+            this.value = value;
+            this.prev = null;
+            this.next = null;
+        }
+    }
+
+    public MyLinkedList() {
+    }
+
     @Override
     public void add(T value) {
+        Node<T> newNode = new Node<>(value);
+        if (head == null && tail == null) {
+            head = newNode;
+            tail = newNode;
+        } else {
+            tail.next = newNode;
+            newNode.prev = tail;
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,157 @@
...
+                newNode.next = currentNode;
+                newNode.prev = currentNode.prev;
+                currentNode.prev = newNode;
+            } else if (isHead(currentNode)) {
+                head = newNode;
+                head.next = currentNode;
+                currentNode.prev = head;
+            }
+            size++;
+        }
```
olekskov, line 59: 1. It's better to avoid additional tabulations where possible to make code more readable
2. Conditions seems over-designed. As we checking it's same object we may use == in such case
```suggestion
        if (index == size) {
            add(value);
            return;
        }
        checkIndex(index);
        Node<T> newNode = new Node<>(value);
        Node<T> currentNode = getNodeByIndex(index);
        if (currentNode == head) {
            head = newNode;
            head.next = currentNode;
            currentNode.prev = head;
        } else {
            currentNode.prev.next = newNode;
            newNode.next = currentNode;
            newNode.prev = currentNode.prev;
            currentNode.prev = newNode;
        }
        size++;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,157 @@
...
+            head = head.next;
+        } else if (isTail(removedElement)) {
+            tail = tail.prev;
+        } else if (size == 1) {
+            head = null;
+            tail = null;
+        }
+        removedElement.next = null;
+        removedElement.prev = null;
+        size--;
```
olekskov, line 146: 1. You forgot to set next/prev as null for new tail/head after removal
2. Conditions may be simplified
```suggestion
        if (size == 1) {
            head = null;
            tail = null;
        } else if (removedElement == head) {
            head = head.next;
            head.prev = null;
        } else if(removedElement == tail) {
            tail = tail.prev;
            tail.next = null
        } else {
            removedElement.next.prev = removedElement.prev;
            removedElement.prev.next = removedElement.next;
        }
        size--;
```

kozhukhovsky (approved):

PR #1023, opened 2023-05-09
Hw -linkedList

RostyslavOnysh:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,176 @@
...
+            current.prev.next = newNode;
+            current.prev = newNode;
+        }
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
+        if (list == null) {
+            throw new NullPointerException("The list parameter cannot be null.");
```
olekskov, line 54: Instead of throwing a 'NullPointerException', it is more appropriate to throw an 'IllegalArgumentException' with the same message.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,176 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private boolean isEqual(T value, T obj) {
+        return Objects.equals(obj, value);
```
olekskov, line 126: Please write your implementation of null-safe comparison as usage of Objects class is not allowed for this task

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,176 @@
...
+
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
     @Override
     public void add(T value) {
+        Node<T> newNode = new Node<>(tail, value, null);
+        if (tail == null) {
+            head = newNode;
```
olekskov, line 16: with such implementation your head will have tail as prev, which may cause unpredicted behaviour in some cases

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,180 @@
...
+        } else {
+            Node<T> current = findNode(index);
+            newNode = new Node<>(current.prev, value, current);
+            current.prev.next = newNode;
+            current.prev = newNode;
+
+        }
+        size++;
+        if (tail == newNode) {
+            head.prev = newNode;
```
kozhukhovsky, line 47: This line is unnecessary as the tail is already being updated in other parts of the code properly.

kozhukhovsky (approved):

PR #1028, opened 2023-05-16
add my solution

aTiukhtii:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,141 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size;
+    private Node<T> head;
+    private Node<T> tail;
+
+    class Node<T> {
```
kozhukhovsky, line 10: Needs to be made private and moved down the class

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,141 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void indexVerify(int index) {
```
kozhukhovsky, line 104:
```suggestion
    private void checkIndex(int index) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,141 @@
...
+                tail.next = null;
+            }
+        } else {
+            node.prev.next = node.next;
+            node.next.prev = node.prev;
+        }
+        size--;
+    }
+
+    private Node<T> getNodeByIndex(int index) {
```
kozhukhovsky, line 128: it is better to add check for index in this method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,141 @@
...
+        return getNodeByIndex(index).value;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        indexVerify(index);
+        T oldValue = getNodeByIndex(index).value;
+        getNodeByIndex(index).value = value;
+        return oldValue;
```
kozhukhovsky, line 71: Do we need to call the node getter twice?

kozhukhovsky (changes requested):

kozhukhovsky (approved):

PR #1029, opened 2023-05-20
Homework jv-linked-list

UniyatYura:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,132 @@
 package core.basesyntax;
 
 import java.util.List;
+import java.util.Objects;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    public static class Node<T> {
```
Ivan95kos, line 7: All inner classes should be at the end of class
```suggestion
    private class Node<T> {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,132 @@
...
+        public Node(T value) {
+            this.value = value;
+        }
+    }
+
+    private Node<T> head;
+    private Node<T> tail;
+    private int size = 0;
+
+    private Node<T> getNodeByIndex(int index) {
```
Ivan95kos, line 20: All private methods should be below public methods

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,132 @@
...
     }
 
     @Override
     public boolean remove(T object) {
-        return false;
+        Node<T> tmpNode = head;
+        boolean isElementExist = false;
+        int indexOfElement = 0;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(tmpNode.value, object)) {
```
Ivan95kos, line 110: Don’t use class Objects.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,132 @@
...
     public T set(T value, int index) {
-        return null;
+        Objects.checkIndex(index, size);
+        T oldValue = getNodeByIndex(index).value;
+        getNodeByIndex(index).value = value;
+        return oldValue;
     }
 
     @Override
     public T remove(int index) {
```
Ivan95kos, line 83: Create a separate method unlink(Node node).

Ivan95kos (changes requested):
Please, pay attention to common mistakes

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,149 @@
...
+            head = node.next;
+        } else {
+            node.prev.next = node.next;
+        }
+        if (node.next == null) {
+            tail = node.prev;
+        } else {
+            node.next.prev = node.prev;
+        }
+        size--;
```
okuzan, line 147:
```suggestion
        size--;
        return node.value;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,149 @@
 package core.basesyntax;
 
 import java.util.List;
+import java.util.Objects;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+
```
okuzan, line 7: remove empty line

okuzan (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,134 @@
...
+        tail = newNode;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index == size) {
+            add(value);
+            return;
+        } else {
```
okuzan, line 28: you don't need `else` after if with `return`, unwrap your code, nested code blocks decrease readability

okuzan (changes requested):

okuzan (approved):
Good job!

PR #1035, opened 2023-05-24
Linked List 0

Mutaborr:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,180 @@
...
+        }
+    }
+
+    private void checkPositionIndex(int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException();
+        }
+    }
+
+    private void unlink(Node<T> current) {
```
olekskov, line 37: we should place public methods before private ones

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,180 @@
...
+    @Override
+    public void add(T value, int index) {
+        checkPositionIndex(index);
+        if (index == size) {
+            add(value);
+        } else {
+            Node<T> current = head;
+            for (int i = 0; i < index; i++) {
+                current = current.next;
+            }
```
olekskov, line 75: Why method getNodeByIndex() is not reused here? seems suits perfectly

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,180 @@
...
+            for (int i = 0; i < index; i++) {
+                previous = previous.next;
+            }
+            Node<T> current = previous.next;
+            previous.next = current;
+            current.prev = previous.prev;
+            previous.prev.next = current;
+            size--;
+            return previous.value;
+        }
```
olekskov, line 157: Should be something like below. Recheck if getNodeByIndex() and unlink() are ready for such usage but they should be
```suggestion
        Node<T> nodeToRemove =  getNodeByIndex(index);
        T result = nodeToRemove.value;
        unlink(nodeToRemove);
        return result;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,180 @@
...
+import java.util.List;
+import java.util.Objects;
+
+public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+
+    private Node head;
+    private Node tail;
+    private int size;
+
+    public class Node<T> {
```
olekskov, line 12: we shouldn't expose inner class in this case

src/main/java/core/basesyntax/MyLinkedListInterface.java
```diff
@@ -1,23 +1,23 @@
...
-
-    T remove(int index);
-
-    boolean remove(T object);
-
-    int size();
-
-    boolean isEmpty();
-}
+package core.basesyntax;
```
okuzan, line 1: undo changes to the interface class

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,155 @@
...
-    public boolean isEmpty() {
-        return false;
-    }
-}
+package core.basesyntax;
+
+import java.util.List;
+
+public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+
```
okuzan, line 6: remove the line

src/main/java/core/basesyntax/MyLinkedList.java
okuzan: not fixed, wrong access modifier. also move it to the very end of the file

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,155 @@
...
+        private T value;
+        private Node<T> next;
+        private Node<T> prev;
+
+        public Node(Node<T> prev, T value, Node<T> next) {
+            this.next = next;
+            this.value = value;
+            this.prev = prev;
+        }
+
```
okuzan, line 21: extra empty line

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,155 @@
...
+    public void addAll(List<T> list) {
+        for (T element : list) {
+            add(element);
+        }
+    }
+
+    @Override
+    public T get(int index) {
+        checkIndex(index);
+        return getNodeByIndex(index).value;
```
okuzan, line 67: it's only logical to include the index check inside the methods that retrieves by index

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,155 @@
...
+                node = node.prev;
+            }
+            return node;
+        }
+    }
+
+    @Override
+    public T set(T value, int index) {
+        checkIndex(index);
+        Node<T> current = getNodeByIndex(index);
```
okuzan, line 89:
```suggestion
        Node<T> current = getNodeByIndex(index);

```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,155 @@
...
+        } else if (current.next == null) {
+            tail = current.prev;
+            tail.next = null;
+        } else {
+            current.prev.next = current.next;
+            current.next.prev = current.prev;
+        }
+        size--;
+    }
+
```
okuzan, line 154: extra empty line

okuzan (changes requested):

src/main/java/core/basesyntax/MyLinkedListInterface.java
okuzan: not fixed
 if the file contents were deleted and then added again seemingly the same, Git might not recognize it as the exact same content due to line-ending or whitespace differences. To compare the differences, you can use Git's diff command: `git diff <commit-hash> -- <file-path>`. Replace <commit-hash> with the hash of the commit before the deletion, and <file-path> with the path to the file.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,152 @@
...
+    @Override
+    public void addAll(List<T> list) {
+        for (T element : list) {
+            add(element);
+        }
+    }
+
+    @Override
+    public T get(int index) {
+        checkIndex(index);
```
okuzan, line 52: delete here too

src/main/java/core/basesyntax/MyLinkedList.java
okuzan: not fixed

okuzan (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
aaananas: **Use getNodeByIndex**

aaananas (changes requested):

aaananas (approved):

PR #1040, opened 2023-06-05
create my own linkedlist

karvatskyi:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,155 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    class Node<T> {
```
aaananas, line 6: inner classes should be after fields

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,155 @@
...
+        if (first == null) {
+            first = last = newNode;
+        }
+        if (index == 0) {
+            newNode.next = first;
+            first = newNode;
+        } else if (index == size) {
+            last.next = newNode;
+            last = newNode;
+        } else {
```
aaananas, line 46:
```suggestion
        if (index == size) {
            add(value);
        }
        checkIndex(index);
        Node<T> newNode = new Node<>(value);
        if (first == null) {
            first = last = newNode;
        }
        if (index == 0) {
            newNode.next = first;
            first = newNode;
        } else {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,155 @@
...
+                    first = current.next;
+                } else {
+                    previous.next = current.next;
+                }
+                if (current == last) {
+                    last = previous;
+                }
+                size--;
+                return true;
+            }
```
aaananas, line 126: combine those ifs, you are just duplicating your code

see common mistakes, you should use unlink method here

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,155 @@
...
     public boolean isEmpty() {
-        return false;
+        return first == null;
+    }
+
+    private Node<T> getNodeByIndex(int index) {
+        Node<T> current = first;
+        for (int i = 0; i < index; i++) {
+            current = current.next;
+        }
```
aaananas, line 147: see common mistakes

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,155 @@
...
+            if (first == null) {
+                last = null;
+            }
+        } else {
+            Node<T> prev = getNodeByIndex(index - 1);
+            removedElement = prev.next.element;
+            prev.next = prev.next.next;
+            if (index == size - 1) {
+                last = prev;
+            }
```
aaananas, line 92: you should use unlink method here

aaananas (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
olekskov: Anastasia's comment not fixed.
Please implement iteration from last element if index provided is closer to the end of list

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,191 @@
...
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index == size) {
+            add(value);
+            return;
+        }
+        checkIndex(index, size + 1);
```
olekskov, line 36: 1. You shouldn't allow add with index = size + 1 as your actual last index is size - 1, so size is limit for add operation:
```suggestion
        checkIndex(index, size);
```

2. After such change all your checkIndex method will be called with (index, size) arguments so size may be omitted from method totally:
```suggestion
        checkIndex(index);
```

olekskov (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,191 @@
...
+                    || (object == null && get(i) == null)) {
+                index = i;
+                break;
+            }
+        }
+        if (index != -1) {
+            remove(index);
+            return true;
+        }
         return false;
```
aaananas, line 98:
```suggestion
        Node<T> pointer = head;
        for (int i = 0; i < size; i++) {
            if (Objects.equals(pointer.value, value)) {
                unlink(node);
                return true;
            }
            pointer = pointer.next;
        }
        return false;
```

src/main/java/core/basesyntax/MyLinkedList.java
aaananas: **See common mistakes to see what is wrong with this method**

aaananas (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size;
+    private Node<T> first;
+    private Node<T> last;
+
+    class Node<T> {
```
okuzan, line 10:
```suggestion
    private static class Node<T> {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
+        private T element;
+        private Node<T> next;
+        private Node<T> prev;
+
+        public Node(T element) {
```
okuzan, line 15:
```suggestion
        Node(T element) {
```
A private inner static class is intended to be used only within its containing class and is not directly accessible from the outside. Since it is meant to be used internally, there is typically no need to declare a public constructor

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
+        Node<T> newNode = new Node<>(value);
+        Node<T> current = getNodeByIndex(index);
+        Node<T> prev = current.prev;
+        newNode.next = current;
+        newNode.prev = prev;
+        prev.next = newNode;
+        current.prev = newNode;
+    }
+
+    public void unLink(Node<T> node) {
```
okuzan, line 151:
```suggestion
    public void unlink(Node<T> node) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
+        Node<T> current = first;
+        for (int i = 0; i < index; i++) {
+            current = current.next;
+        }
+        return current;
+    }
+
+    private int checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Index is out of bounds: " + index);
```
okuzan, line 117: would be good to have size here too

src/main/java/core/basesyntax/MyLinkedList.java
okuzan: this is the relevant part
image

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
+                index = i;
+                break;
+            }
+        }
+        if (index != -1) {
+            Node<T> removedNode = getNodeByIndex(index);
+            unLink(removedNode);
+            return true;
+        }
         return false;
```
okuzan, line 94: you dont'need that technique with the variable here (flag if found). Just call unlink on line 85 and return true, otherwise return false outside of the loop

src/main/java/core/basesyntax/MyLinkedList.java
kozhukhovsky: not fixed

kozhukhovsky (approved):

PR #1042, opened 2023-06-07
linked list implementation

krismva:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,152 @@
...
+        return old;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        Node<T> nodeToRemove = findNodeByIndex(index);
+        unlink(nodeToRemove);
+
+        return nodeToRemove.value;
```
okuzan, line 62:
```suggestion
        return unlink(findNodeByIndex(index)).value;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,152 @@
...
 
     @Override
     public boolean remove(T object) {
-        return false;
+        Node<T> removedNode = findByElement(object);
+        if (removedNode == null) {
+            return false;
+        }
+        unlink(removedNode);
+        return true;
```
okuzan, line 72: here it would be actually less code if you moved findByElement logic here, call unlink inside of if + return true, false outside of the loop

aaananas (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,152 @@
...
+        }
+        return null;
+    }
+
+    private class Node<T> {
+        private T value;
+        private Node<T> prev;
+        private Node<T> next;
+
+        public Node(Node<T> prev, T value, Node<T> next) {
```
okuzan, line 148: why public

okuzan (changes requested):
Good job overall 👏

PR #1046, opened 2023-06-17
Hw linked list

DZarudnyi:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,171 @@
...
     }
 
     @Override
     public void add(T value, int index) {
+        checkAddingIndex(index);
+        if (index == size) {
+            linkLast(value);
+        } else {
+            linkNode(value, getNode(index));
+        }
```
olekskov, line 22: With such change we may remove checkAddingIndex() method
```suggestion
        if (index == size) {
            linkLast(value);
        } else {
            checkIndex(index);
            linkNode(value, getNode(index));
        }
```

olekskov (approved):
Good job, approved.
Please check comment below to take into account further

PR #1059, opened 2023-06-29
meyson-hw-jv-linked-list-solution

Meysonsan:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,166 @@
...
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size;
+    private Node<T> head;
+    private Node<T> tail;
+
+    public MyLinkedList() {
+        head = null;
+        tail = null;
+    }
```
Ivan95kos, line 13: useless action, they have default value null
```suggestion
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,166 @@
...
-        return null;
+        checkIndex(index);
+        Node<T> node = getNode(index);
+        T oldValue = node.value;
+        node.value = value;
+        return oldValue;
     }
 
     @Override
     public T remove(int index) {
```
Ivan95kos, line 75: Create a separate method unlink(Node node).

For example:
```
    @Override
    public T remove(int index) {
        checkIndex(index);
        checkIndexEqualsSize(index);
        Node<T> currentNode = getCurrentNodeBy(index);
        unlinkNode(currentNode);
        return currentNode.element;
    }
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,166 @@
...
+            nodeToRemove.next.prev = nodeToRemove.prev;
+        } else {
+            tail = nodeToRemove.prev;
+        }
+        size--;
+        return nodeToRemove.value;
     }
 
     @Override
     public boolean remove(T object) {
```
Ivan95kos, line 93: Example:
```
    @Override
        Node<T> currentNode = head;
        while (currentNode != null) {
            if (currentNode.element == object || object != null && object.equals(currentNode.element)) {
                unlinkNode(currentNode);
                return true;
            }
        }
        return false;
    }
```

Ivan95kos (changes requested):
Good job overall!

Ivan95kos (approved):
Nice

PR #1060, opened 2023-07-01
Create simple MyLinkedList<T> implementing custom MyLinkedListInterface<T>

barbequeue:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,160 @@
...
+            link(value, getNode(index));
+        }
     }
 
     @Override
     public void addAll(List<T> list) {
+        if (list == null) {
+            return;
+        }
+
```
okuzan, line 37:
```suggestion
```
if list is null, should throw NPE
image

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,160 @@
...
+    private void checkIndexWithin(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("LOL, YOU DUMB? XDXDXDXDXDX");
+        }
+    }
+
+    private void checkIndexBetween(int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Well, shit happens!");
```
okuzan, line 98: include index and size in the message, make it clear what's wrong

okuzan (changes requested):
Good job overall!

src/main/java/core/basesyntax/MyLinkedList.java
barbequeue: Oh gosh, i've completely forgot to remove those stupid messages, when i was working on the code implementation 🙈

Ivan95kos (approved):
Nice

PR #1064, opened 2023-07-03
solution for jv-linked-list

vladis9:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,151 @@
...
+        checkIndex(index);
+        Node<T> node = first;
+        for (int i = 0; i < size; i++) {
+            if (i == index) {
+                break;
+            }
+            node = node.next;
+        }
+        return node;
+    }
```
okuzan, line 131: iterate from tail if it will be more efficient

vladis9: i'm not sure what do you mean

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,148 @@
...
+
+    private Node<T> getNodeByIndex(int index) {
+        checkIndex(index);
+        Node<T> node = first;
+        for (int i = 0; i < index; i++) {
+            node = node.next;
+        }
+        return node;
+    }
```
olekskov, line 128: Now you starting to look for proper node in getNodeByIndex() method from first node regardless of index.
E.g. size is 1011, we call getNodeByIndex(1008). So we will have 1008 calls of node.next instead of making 3 calls of node.prev going from last node.
Please implement logic that will go from last in case index is more than half of size

olekskov (approved):
Good job

PR #1074, opened 2023-07-09
version1-Stas

stas83ggg:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,153 @@
...
+
+    private Node<T> getNode(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Invalid index: " + index);
+        }
+        Node<T> current = head;
+        for (int i = 0; i < index; i++) {
+            current = current.next;
+        }
+        return current;
```
olekskov, line 153: Please add logic to iterate from tail when index is closer to the end of list
https://mate-academy.github.io/jv-program-common-mistakes/java-core/collections/linked-list.html#dont-iterate-your-nodes-from-the-head-to-tail-or-vice-versa

olekskov (changes requested):
Please fix build and introduce logic according to comment below:

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
...
+
+    private static class Node<E> {
+        private final E item;
+        private Node<E> next;
+        private Node<E> prev;
+
+        private Node(Node<E> prev, E element, Node<E> next) {
+            this.prev = prev;
+            this.item = element;
+            this.next = next;
```
okuzan, line 141:
```suggestion
            this.prev = prev;
            this.next = next;
            item = element;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
...
+        }
+    }
+
+    private Node<T> getNode(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Invalid index: " + index);
```
okuzan, line 147:
```suggestion
            throw new IndexOutOfBoundsException("Invalid index: " + index, " for size:  "  + size);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
...
+                head = null;
+            }
+        } else {
+            node.next.prev = node.prev;
+            node.prev.next = node.next;
+        }
+        size--;
+    }
+
+    private static class Node<E> {
```
okuzan, line 133: better place the inner class at the very end of the file, after all the methods

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
...
+            newHead.next = head.next;
+            head.next.prev = newHead;
+            head = newHead;
+            return oldValue;
+        }
+        Node<T> nodeToChange = getNode(index);
+        Node<T> node = new Node<>(nodeToChange.prev, value, nodeToChange.next);
+        nodeToChange.prev.next = node;
+        nodeToChange.next.prev = node;
+        return nodeToChange.item;
```
okuzan, line 78: Don't create a new node.
1. get the node (line 74)
2. save old value in a local var (will be returned)
3. reassign item of the node to the new value
4. return old value
We don't need to handle index == 0 separately

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
...
+            tail = newTail;
+        } else {
+            Node<T> nodeByIndex = getNode(index);
+            Node<T> middleCurrent = new Node<>(nodeByIndex.prev, value, nodeByIndex);
+            nodeByIndex.prev.next = middleCurrent;
+            middleCurrent.prev = nodeByIndex.prev;
+            nodeByIndex.prev = middleCurrent;
+            middleCurrent.next = nodeByIndex;
+        }
+        size++;
```
okuzan, line 49: Your implementation of the add method works correctly, but it seems a bit more complex than necessary. Let's review and see how we could simplify it.

Redundant assignments: There are several instances where you are doing the same assignment twice. For example, newHead.next = head; and newTail.prev = tail; are redundant because you are already setting these fields in the Node constructor.

Handling head and tail separately: Your code specifically handles the cases where the index is 0 (head) or equals to size (tail). While this does make sense, it results in duplicated logic that can be avoided. With a proper check for the previous node being null, the "head" case can be handled together with the general case, thus simplifying the code. The "tail" case can be handled as an append operation.

Prev and next references: The references to the previous and next nodes are somewhat scattered throughout your code. Grouping them together can make the logic more clear.

Here is a simpler algorithm to follow:

* If the index equals to size, treat it as an append operation and call the add(value) method directly.
Check the index to ensure it's valid.
* Retrieve the node currently at the specified index. This will become the "next" node for the new node.
* Create a new node with its next reference pointing to the retrieved node and its prev reference pointing to the prev reference of the retrieved node.
* If the prev reference of the new node is null, it means we are inserting at the head, so update the first reference to point to the new node. Otherwise, update the next reference of the prev node to point to the new node.
* Finally, update the prev reference of the next node (the node retrieved initially) to point to the new node.
* Remember to increment the size.

This way, your code will be more concise and to the point. I hope these suggestions are helpful and looking forward to your updated contribution.

okuzan (changes requested):

okuzan (approved):
Splendid! Clean code

PR #1082, opened 2023-07-12
implemented LinkedList class with methods

Someboty:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,117 @@
...
 
+import java.util.ArrayList;
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private static final String NEGATIVE_INDEX_MESSAGE = "Index must be positive number";
+    private static final String INDEX_OUT_OF_BOUNDS_MESSAGE = "Index must be positive number";
+    private final ArrayList<Node<T>> data = new ArrayList<>();
+    private int size;
+
```
OlegTuryk, line 11: The part of task is to add  two pointers to the first and the last Node

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,117 @@
 package core.basesyntax;
 
+import java.util.ArrayList;
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private static final String NEGATIVE_INDEX_MESSAGE = "Index must be positive number";
+    private static final String INDEX_OUT_OF_BOUNDS_MESSAGE = "Index must be positive number";
+    private final ArrayList<Node<T>> data = new ArrayList<>();
```
OlegTuryk, line 9: I don't think you need this in that task

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,117 @@
...
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
+        int newSize = size + list.size() * 2;
+        for (int i = size; i < newSize; i = i + 2) {
+            add(list.get(i - size));
+        }
     }
```
OlegTuryk, line 55: The addAll method should not double the size of the list when adding elements. It can simply iterate over the list parameter and call the add method for each element.

OlegTuryk (changes requested):

Zel1oy (changes requested):

Dimagaa (changes requested):
Using ArrayList for storage data looks like cheating :)
It is meaningless because your implementation doesn't acquire benefits Linked List (Adding Removing)

src/main/java/core/basesyntax/MyLinkedList.java
Someboty: Why?

src/main/java/core/basesyntax/MyLinkedList.java
Someboty: addAll() doesn't doubles list's size. newSize - variable for cycle's process limiting, that's all

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,117 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void indexValidator(int index, boolean deleteContacted) {
```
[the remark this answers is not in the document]
Someboty: I need to separate behaviour for two cases: when you delete smth you can't take size value, because it's actually empty (when list have size 5, it has indexes 0, 1, 2, 3, 4, but there are no 5, you can't delete object with this index). But when you add something at that position, it will be fine

src/main/java/core/basesyntax/MyLinkedList.java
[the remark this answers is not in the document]
Someboty: But where should I save Nodes then?

src/main/java/core/basesyntax/MyLinkedList.java
Someboty: Looks like I'm forgot about it somehow. Thanks

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
...
+            throw new IndexOutOfBoundsException(INDEX_OUT_OF_BOUNDS_MESSAGE);
+        } else if (index < 0) {
+            throw new IndexOutOfBoundsException(NEGATIVE_INDEX_MESSAGE);
+        }
+    }
+
+    private Node<T> findNodeByIndex(int index) {
+        Node<T> current;
+        if (index < size / 2) {
+            current = head;
```
OlegTuryk, line 147: Added null checks in the findNodeByIndex(int index) method to handle cases where the linked list is empty.

OlegTuryk (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private static final String NEGATIVE_INDEX_MESSAGE = "Index must be positive number";
+    private static final String INDEX_OUT_OF_BOUNDS_MESSAGE = "Index must be positive number";
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    private static class Node<T> {
```
Vasul-Pankiv, line 12: make it non static and move to the end of class

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void indexAddValidator(int index) {
```
Vasul-Pankiv, line 128: better to rename it checkIndexToAddNode / checkIndexToRevomeNode

Vasul-Pankiv (changes requested):

Vasul-Pankiv (approved):

PR #1087, opened 2023-07-12
homework solution

margosha99:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,142 @@
...
+            tail.next = newNode;
+            tail = newNode;
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        Node<T> newNode = new Node<>(value);
+        if (index > size || index < 0) {
```
okuzan, line 39:
```suggestion
        if (index < 0 || index > size) {
```
a minor issue but IMO it's more readable when conditions are in a linear direction, from small number to big numbers

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,142 @@
...
+            newNode.prev = tail;
+            tail.next = newNode;
+            tail = newNode;
+        } else {
+            Node<T> prev = getNodeByIndex(index - 1);
+            newNode.prev = prev;
+            newNode.next = prev.next;
+            prev.next = newNode;
+        }
+        size++;
```
okuzan, line 58: Here is a simpler algorithm to follow:

* If the index equals to size, treat it as an append operation and call the add(value) method directly.
* Check the index to ensure it's valid.
* Retrieve the node currently at the specified index. This will become the "next" node for the new node.
* Create a new node with its next reference pointing to the retrieved node and its prev reference pointing to the prev reference of the retrieved node.
* If the prev reference of the new node is null, it means we are inserting at the head, so update the first reference to point to the new node. Otherwise, update the next reference of the prev node to point to the new node.
* Finally, update the prev reference of the next node (the node retrieved initially) to point to the new node.
* Remember to increment the size.
This way, your code will be more concise and to the point.

okuzan (changes requested):
Before you create PR with your homework, you need to go through the checklist under the task and correct all the points described there. The mentor will not check the work until the checklist points are corrected

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,142 @@
...
+        return head == null;
+    }
+
+    private Node<T> getNodeByIndex(int index) {
+        Node<T> current = head;
+        for (int i = 0; i < index; i++) {
+            current = current.next;
+        }
+        return current;
+    }
```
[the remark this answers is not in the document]
kozhukhovsky: not fixed

kozhukhovsky (commented):
Not fixed all previous comment

kozhukhovsky (changes requested):
Not fixed all previous comment

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,152 @@
...
 
 import java.util.List;
+import java.util.Objects;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size;
+    private Node<T> head;
+    private Node<T> tail;
+
+    private static class Node<T> {
```
Ivan95kos, line 11: All inner classes should be at the end of class

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,152 @@
...
+            Node<T> prev = getNodeByIndex(index - 1);
+            newNode.prev = prev;
+            newNode.next = prev.next;
+            prev.next.prev = newNode;
+            prev.next = newNode;
+            size++;
+        }
+    }
+
+    private void addHeadToList(Node<T> newNode) {
```
Ivan95kos, line 54: All private methods should be below public methods

Ivan95kos (changes requested):
Good job overall!
Only a few comments

kozhukhovsky (approved):

PR #1091, opened 2023-07-13
Implementaion methods in LinkedList task

Stelmaschyk:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
+        tail.prev = newNode;
+        if (prev == null) {
+            head = newNode;
+        } else {
+            prev.next = newNode;
+        }
+        size++;
+    }
+
+    private Node<T> node(int index) {
```
okuzan, line 127: method can't be named with a noun, should be verb, getNodeByIndex

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
+        checkElementIndex(index);
+        Node<T> removedElement = node(index);
+        unlink(removedElement);
+        return removedElement.item;
     }
 
     @Override
     public boolean remove(T object) {
+        for (Node<T> removedElement = head; removedElement != null; removedElement =
+                removedElement.next) {
```
okuzan, line 72: while loop here would be more readable

okuzan (changes requested):
Place index checker inside getNodeByIndex, place index related methods together, make sure you really need all of them, work on their naming, isElementIndex doesn't really tell what it does

kozhukhovsky (approved):

PR #1095, opened 2023-07-13
Added solution

AvanesovOleh:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,187 @@
...
+                nodeIndex++;
+            }
+            removedValue = currentNode.value;
+            currentNode.prev.next = currentNode.next;
+            if (size - index != 1) {
+                currentNode.next.prev = currentNode.prev;
+            }
+            nodeIndex = 0;
+        }
+        size--;
```
aaananas, line 127: lets use getNode(index) and unlink(node)

https://mate-academy.github.io/jv-program-common-mistakes/java-core/collections/linked-list.html#create-a-separate-method-unlinknode-node

aaananas (changes requested):
Before you create PR with your homework, you need to go through the checklist under the task and correct all the points described there. The mentor will not check the work until the checklist points are corrected

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,152 @@
...
+        }
     }
 
     @Override
     public void addAll(List<T> list) {
+        for (T listElement : list) {
+            Node<T> newNode = new Node<>(tail, listElement, null);
+            tail.next = newNode;
+            tail = newNode;
+            size++;
```
olekskov, line 67: Don't understand why copy part of code from add() method and why skipped part that allows addition of firts element. 
Just reuse add method here

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,152 @@
...
     @Override
     public void add(T value, int index) {
+        if (index > size || index < 0) {
+            throw new IndexOutOfBoundsException("Index " + index + " is out of list");
+        }
+        if (size == index) {
+            add(value);
+        } else {
+            Node<T> currentNode = getNode(index);
+            Node<T> addedNode = new Node<>(currentNode.prev, value, currentNode);
```
olekskov, line 46: It's better to slightly change order of actions and reuse checkIndex method:
```suggestion
        if (size == index) {
            add(value);
            return;
        } 
        checkIndex(index);
        Node<T> currentNode = getNode(index);
        Node<T> addedNode = new Node<>(currentNode.prev, value, currentNode);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,152 @@
...
+            if (nodeIndex == 0) {
+                head.prev = addedNode;
+                addedNode.prev = null;
```
olekskov, line 49: This line seems excessive. 
On the first glance if we have have index 0 then currentNode is head. As head.prev should be null then addedNode should already have null after using constructor on line 46
```suggestion
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,152 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size;
+    private int nodeIndex;
```
olekskov, line 7: Totally not undestand why we need this class variable.
Mostly it's just create additional code in methods.
Instead of checking (nodeIndex == 0) or (size - nodeIndex != 1) you may just compare (currentNode == head) or  (currentNode != tail) etc. As in this case we check if it's the same object
```suggestion
```

olekskov (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,142 @@
...
+    private void checkIndex(int index) {
+        if (index >= size || index < 0) {
+            throw new IndexOutOfBoundsException("Index " + index + " is out of list");
+        }
+    }
+
+    private Node<T> getNode(int index) {
+        int nodeIndex = 0;
+        Node<T> currentNode = head;
+        while (nodeIndex != index) {
```
aaananas, line 123: see common mistakes

image

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,142 @@
...
+               || currentNode.value != null && currentNode.value.equals(object))) {
+            currentNode = currentNode.next;
+            nodeIndex++;
+            if (nodeIndex == size) {
+                nodeIndex = 0;
+                return false;
+            }
+        }
+        unlink(currentNode);
+        return true;
```
aaananas, line 89:
```suggestion
        Node<T> currentNode = head;
        while (!(currentNode.value == object
               || currentNode.value != null && currentNode.value.equals(object))) {
            currentNode = currentNode.next;
            nodeIndex++;
            if (nodeIndex == size) {
                return false;
            }
        }
        unlink(currentNode);
        return true;
```

aaananas (changes requested):

aaananas (approved):

PR #1109, opened 2023-07-14
implemented MyLinkedListInterface in MyLinkedList class

irynamekh:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,149 @@
...
+        private T value;
+        private Node<T> prev;
+        private Node<T> next;
+
+        public Node(Node<T> prev, T value, Node<T> next) {
+            this.prev = prev;
+            this.value = value;
+            this.next = next;
+        }
+    }
```
ivan0dyatlyukkk, line 20: We shouldn't use the Node class outside because it's useless, so make it private and non-static

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,149 @@
...
+            head = newNode;
+            size++;
+        } else {
+            Node<T> current = getNode(index);
+            newNode.next = current;
+            newNode.prev = current.prev;
+            current.prev.next = newNode;
+            current.prev = newNode;
+            size++;
+        }
```
ivan0dyatlyukkk, line 56: Please, add some new methods to make it more readable.
** Methods like "addInMiddle" or "addAtBeginning"

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,149 @@
...
     }
 
     @Override
     public void addAll(List<T> list) {
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
```
ivan0dyatlyukkk, line 63: It's not a problem, but we can use the for-each loop here as well, instead of the for loop.

ivan0dyatlyukkk (changes requested):
Good job! However, I found some slight problems. I hope you fix it fast  🚀

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,159 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private class Node<T> {
```
andrii-hoienko, line 93: Move to the bottom of class

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,159 @@
...
+            if (tail != null) {
+                tail.next = null;
+            }
+        } else {
+            node.prev.next = node.next;
+            node.next.prev = node.prev;
+        }
+    }
+
+    private Node<T> getNode(int index) {
```
andrii-hoienko, line 128: getNodeByIndex is better naming

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,159 @@
...
+        Node<T> current = head;
+        if (index < size / 2) {
+            for (int i = 0; i < index; i++) {
```
andrii-hoienko, line 131: Extract logic of search from head and tail to methods

andrii-hoienko (changes requested):
Very good, some issues to resolve and toy are done!

PR #1113, opened 2023-07-17
Implemented all methods.

Alex-Solod:
Created Node objects to store data that have links to the left and to the right neighbour Node.

### DIFF
--- a/src/main/java/core/basesyntax/MyLinkedList.java
+++ b/src/main/java/core/basesyntax/MyLinkedList.java
@@ -1,47 +1,144 @@
 package core.basesyntax;
 
 import java.util.List;
+import java.util.Objects;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node head;
+    private Node tail;
+    private int size = 0;
+
     @Override
     public void add(T value) {
+        if (size == 0) {
+            head = new Node(null, value, null);
+            tail = head;
+        } else {
+            Node secondLast = tail;
+            tail = new Node(secondLast, value, null);
+            secondLast.next = tail;
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Invalid index: " + index);
+        }
+
+        if (index == size) {
+            add(value);
+        } else {
+            Node nodeAtIndex = getNode(index);
+            Node newNode = new Node(nodeAtIndex.previous, value, nodeAtIndex);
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,144 @@
 package core.basesyntax;
 
 import java.util.List;
+import java.util.Objects;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node head;
+    private Node tail;
+    private int size = 0;
```
okuzan, line 9:
```suggestion
    private int size;
```
0 is default

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,144 @@
...
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Invalid index: " + index);
+        }
+
+        if (index == size) {
+            add(value);
+        } else {
```
okuzan, line 32:
```suggestion
            add(value);
            return;
        }
```
better return early than wrap and indent big chunks of code

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,144 @@
...
+        node.value = value;
+        return oldValue;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        Node nodeToRemove = getNode(index);
+        unlink(nodeToRemove);
+        return nodeToRemove.value;
```
okuzan, line 69:
```suggestion
        return unlink(getNode(index).value;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,144 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private Node getNode(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Invalid index: " + index);
```
okuzan, line 97: include size too

okuzan (changes requested):
Good job so far!

src/main/java/core/basesyntax/MyLinkedList.java
Alex-Solod: got it

kozhukhovsky (approved):

PR #1114, opened 2023-07-17
linked list my first edition

OlgaStarikova:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,175 @@
...
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> firstNode;
+    private Node<T> lastNode;
+    private int size;
+
+    public MyLinkedList() {
+        Node<T> firstNode = null;
+        Node<T> lastNode = null;
+        size = 0;
+    }
```
kozhukhovsky, line 14:
```suggestion
```

0 is default value for int
null is default value for object

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,175 @@
...
+        }
     }
 
     @Override
     public T get(int index) {
+        checkIndexInBounds(index);
+        Node<T> currentNode = firstNode;
+        int i = 0;
+        while (currentNode != null) {
+            if (i == index) {
```
kozhukhovsky, line 72: The logic for getting a node by index is the same in the get and set methods
Put it in a separate private method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,175 @@
...
+                return oldValue;
+            }
+            currentNode = currentNode.next;
+            i++;
+        }
         return null;
     }
 
     @Override
     public T remove(int index) {
```
kozhukhovsky, line 99: It can be made a little easier.
Get the node by index using the method written earlier.
Then make unlink (for this, it is also better to make a separate method, because the logic is repeated).

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,175 @@
...
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> currentNode = firstNode;
+        int i = 0;
+        while (currentNode != null) {
+            if ((currentNode.value != null && currentNode.value.equals(object))
+                    || currentNode.value == object) {
+                remove(i);
```
kozhukhovsky, line 141: here it will be enough to use the new unlink () method, in our remove (index) method there is much more logic than is necessary in this method
By moving the actions to redistribute prev and next between nodes into the unlink method, it will be better.

kozhukhovsky (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,146 @@
...
+        Node<T> nextNode = findByIndex(index);
+        Node<T> node = new Node<T>(nextNode.prev, value, nextNode);
+        if (index == 0) {
+            firstNode = node;
+        } else {
+            nextNode.prev.next = node;
+        }
+        nextNode.prev = node;
+        size++;
+        return;
```
okuzan, line 43:
```suggestion
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,146 @@
...
+        int i = 0;
+        while (currentNode != null) {
+            if (i == index) {
+                return currentNode;
+            }
+            currentNode = currentNode.next;
+            i++;
+        }
+        return null;
+    }
```
okuzan, line 135: you should iterate from the closest end of the list, as indicated here:
image

okuzan (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,153 @@
...
+        Node<T> node = new Node<T>(lastNode, value, null);
+        if (firstNode == null) {
+            firstNode = node;
+            lastNode = node;
+            size++;
+            return;
+        }
+        if (firstNode.next == null) {
+            firstNode.next = node;
+        }
```
olekskov, line 21: This condition doesn't seem necessary. If you have only 1 node then both firstNode and lastNode point to same object so in this case lines 20 and 22 are just duplication each other
```suggestion
```

olekskov (approved):
Good job

PR #1116, opened 2023-07-20
Added implemented MyLinkedListInterface, and  all the required methods.

hanshyn:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,167 @@
...
     @Override
     public void add(T value, int index) {
+        checkIndex(index, size + 1);
+        if (index == 0) {
+            addFirstEntry(value);
+        } else if (index == size) {
+            addLastEntry(value);
+        } else {
+            addByIndexEntry(value, index);
+        }
```
okuzan, line 24: this is the only usage of checkIndex with the second argument being size + 1. If we could use here the usual `size`, we could eliminate this second argument from the method signature, think about where to place it so we can use the usual check to the same effect

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,167 @@
...
+
+    private void addByIndexEntry(T element, int index) {
+        Node<T> nodeRight = findNodeByIndex(index);
+        Node<T> nodeLeft = new Node<T>(nodeRight.previous, element, nodeRight.next);
+        nodeRight.previous.next = nodeLeft;
+        nodeRight.previous = nodeLeft;
+        size++;
+    }
+
+    private Node<T> findNodeByIndex(int index) {
```
okuzan, line 135: include index check here instead of repeatedly calling them together like this:
image

okuzan (changes requested):
Impressive code on your first attempt! Keep it up 💪

okuzan (approved):
Good job, clean code 🤝

PR #1128, opened 2023-08-10
make LinkedList

silchenko-arsen:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,137 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    public MyLinkedList() {
+    }
+
```
okuzan, line 12: default constructor will be added by default, if no other constructors are present

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,137 @@
...
+        }
+        return current;
+    }
+
+    private static class Node<T> {
+        private T value;
+        private Node<T> next;
+        private Node<T> prev;
+
+        public Node(Node<T> prev, T value, Node<T> next) {
```
okuzan, line 133:
```suggestion
        Node(Node<T> prev, T value, Node<T> next) {
```
won't be accessible from outside anyway, public may be confusing

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,137 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private Node<T> getNodeByIndex(int index) {
+        if (index >= size || index < 0) {
+            throw new IndexOutOfBoundsException("Index is bigger than linked list size!");
```
okuzan, line 119: include size and index in the message

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,137 @@
...
     public T set(T value, int index) {
-        return null;
+        Node<T> node = getNodeByIndex(index);
+        T oldValue = node.value;
+        node.value = value;
+        return oldValue;
     }
 
     @Override
     public T remove(int index) {
```
okuzan, line 70: image

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,137 @@
...
+        } else if (index == size - 1) {
+            getNodeByIndex(index - 1).next = null;
+            tail = getNodeByIndex(index - 1);
+        } else {
+            Node<T> prev = getNodeByIndex(index - 1);
+            Node<T> next = getNodeByIndex(index + 1);
+            prev.next = next;
+            next.prev = prev;
+        }
+        size--;
```
okuzan, line 87: try to simplify this code, you don't need to handle all these cases separately, You only need to call getNodeByIndex once

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,137 @@
...
+            add(value);
+        } else {
+            Node<T> prevNode = getNodeByIndex(index - 1);
+            Node<T> nextNode = getNodeByIndex(index);
+            Node<T> newNode = new Node<>(prevNode, value, nextNode);
+            prevNode.next = newNode;
+            nextNode.prev = newNode;
+            size++;
+        }
     }
```
okuzan, line 47: try to simplify and unravel this code. First suggestion - create if for a case when index == size (add(value); and return). Code wrapped in multiple layers of indentation is harder to read

okuzan (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
kozhukhovsky: not fixed

kozhukhovsky (changes requested):

olekskov (approved):
Good job

PR #1134, opened 2023-08-15
implemented LinkedList

RedArtelerist:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,182 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    public MyLinkedList() {
```
ystankevych, line 10: Do you really need to initialize fields with default values? But you don't need to fix it if you think 'yes'. I leave it to the mentors, maybe it's like good tone to do it this way. I'm just pointing this out cause it's an open question for me =)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,182 @@
...
+        head = null;
+        tail = null;
+        size = 0;
+    }
+
     @Override
     public void add(T value) {
+        Node<T> lastNode = tail;
```
ystankevych, line 18: And I have some question here. Last node is always tail (and head is always first node). But I know that such nodes are created internally by the implementation of the linked list. I leave this moment to the mentors.

ystankevych (approved):
I think really great and simple approach. I have some open question and as I am not sure I leave this question to the mentors (They definitely know the right answer)

ystankevych (approved):

liliia-ponomarenko (changes requested):
Good job! Let’s your solution ;)

teract10s (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Index out of bounds: " + index);
+        }
+    }
+
+    private void checkAddedPosition(int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("You can't add value in position " + index);
+        }
+    }
```
NazarSmith228, line 109: so you have two different index checks:

1. for `get`, `set`, `remove` operations
2. for `add` operation

naming in this case is quite complex, in original `LinkedList` these methods are named `checkPositionIndex` and `checkElementIndex` accordingly

I would suggest to name these methods like `checkIndexInBounds` and `checkIndexInBoundsStrict`, which gives more context to the reader 

apart from that, pls use the same messages for the exceptions - you can even extract it to a separate method like `String outOfBoundsMessage()`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
+
+    private Node<T> findNodeByIndex(int index) {
```
NazarSmith228, line 111: since this method is used quite often and it's pretty obvious what it does - you can simplify its name

how about `find()` or `node()`? :)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
+        } else {
+            current = tail;
+            for (int i = size - 1; i > index; i--) {
+                current = current.prev;
+            }
+        }
+        return current;
+    }
+
+    private Node<T> findNodeByValue(T value) {
```
NazarSmith228, line 127: this method is used in a single place - do we really need to extract it?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
+        Node<T> node = head;
+        while (node != null) {
+            if (node.value == value || node.value != null && node.value.equals(value)) {
+                break;
```
NazarSmith228, line 131: using `break` here is unjustified - it adds complication

why not returning immediately from here?

src/main/java/core/basesyntax/MyLinkedList.java
NazarSmith228: @ystankevych I didn't get your question :)

let's discuss it in the Slack chat

src/main/java/core/basesyntax/MyLinkedList.java
NazarSmith228: @ystankevych actually there is no sense to init the fields when creating the list instance

default value for "object" fields - null
default value for primitive int - 0

this is suitable for us :)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
     public void add(T value, int index) {
+        checkAddedPosition(index);
+        if (index == 0) {
+            addFirst(value);
+            return;
+        }
+        if (index == size) {
+            add(value);
+            return;
+        }
```
NazarSmith228, line 33: why not using `else if` and `else`?

you would avoid these unnecessary `return` statements

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
     @Override
     public int size() {
-        return 0;
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return head == null;
```
NazarSmith228, line 84: even though this statement may be correct, you should operate on `size` here

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
     @Override
     public void add(T value) {
+        Node<T> node = new Node<>(tail, value, null);
+        if (tail != null) {
+            tail.next = node;
+        }
+        tail = node;
+        if (head == null) {
+            head = node;
+        }
```
NazarSmith228, line 19: this is quite complicated

basically you have 2 cases:

1. list is empty - then you init both head and tail like `head = tail = new Node()`
2. list is not empty - then you simply add an element at the end (consider creating a method like `addLast()`, you'll be able to  reuse it)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
+
+    private void addFirst(T value) {
+        Node<T> newNode = new Node<>(null, value, head);
+        if (head != null) {
+            head.prev = newNode;
+        }
+        head = newNode;
+        if (tail == null) {
+            tail = newNode;
+        }
```
NazarSmith228, line 95: method is called `addFirst` - then why do you operate on tail here?

the reason lies behind your implementation of add method - refactor it due to my earlier remarks

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
+        while (node != null) {
+            if (node.value == value || node.value != null && node.value.equals(value)) {
+                break;
+            }
+            node = node.next;
+        }
+        return node;
+    }
+
+    private boolean unlink(Node<T> node) {
```
NazarSmith228, line 138: you're mixing `boolean remove()` and `unlink()`

this method should not return any boolean - it should simply unlink the desired node (potentially you can only return the value of that node if you need)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
+        if (node == null) {
```
NazarSmith228, line 139: why would a node be null here?

if it's possible - then it's a drawback of your implementation

NazarSmith228 (changes requested):
Good job! Now it's time to improve your solution

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,168 @@
...
+            head.prev = newNode;
+            head = newNode;
+        }
+        size++;
+    }
+
+    private void addLast(T value) {
+        if (isEmpty()) {
+            head = tail = new Node<>(null, value, null);
+        } else {
```
NazarSmith228, line 97: you have a duplicate:
```
if (isEmpty()) {
   head = tail = new Node<>(null, value, null);
}
```

in fact you shouldn't even care about such scenario in these method

this should be covered by the outer `add` methods

you handle it in `add(value)` but not in `add(value, index)` - that's the reason why this duplicate occurred :)

NazarSmith228 (approved):
LGTM! :)

RedArtelerist:
> LGTM! :)

Thanks for your hard work and picky comments)

PR #1135, opened 2023-08-16
implemented MyLinkedList added inner Node class

ystankevych:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
...
+
+    private void checkIndexToGet(int index) {
+        if (index < 0 || index >= size()) {
+            throw new IndexOutOfBoundsException("Invalid index");
+        }
+    }
+
+    private void checkIndexToAdd(int index) {
+        if (index < 0 || index > size()) {
+            throw new IndexOutOfBoundsException("Invalid index");
```
RedArtelerist, line 90: Maybe in this case will be better to add a more informative message about exception. And also you can add the corresponding index to the message in both methods.

RedArtelerist (approved):
Great job

src/main/java/core/basesyntax/MyLinkedList.java
ystankevych: It makes sense. Fixed =)

RedArtelerist (approved):
Great job

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
+            throw new IndexOutOfBoundsException("Invalid index " + index + " for size " + size());
+        }
+    }
+
+    private void checkIndexToAdd(int index) {
+        if (index < 0 || index > size()) {
+            throw new IndexOutOfBoundsException(
+                    String.format("Index %d out of bound for size %d", index, size()));
+        }
+    }
```
NazarSmith228, line 93: so you have two different index checks:

1. for `get`, `set`, `remove` operations
2. for `add` operation

naming in this case is quite complex, in original `LinkedList` these methods are named `checkPositionIndex` and `checkElementIndex` accordingly

I would suggest to name these methods like `checkIndexInBounds` and `checkIndexInBoundsStrict`, which gives more context to the reader 

apart from that, pls use the same message for the eceptions - you can even extract it to a separate method like `String outOfBoundsMessage()`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
+
+    private Node<T> getNodeByIndex(int index) {
+        return index <= size() / 2 ? getNodeFromHead(index) :
+                getNodeFromTail(index);
+    }
```
NazarSmith228, line 98: I would simplify the name of this method to just `node(int index)`

and also use proper formatting for ternary operator

take a look, doesn't it look better?)
```suggestion
    private Node<T> node(int index) {
        return index <= size / 2 ? nodeFromHead(index) : nodeFromTail(index);
    }
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
+
+    private Node<T> getNodeFromTail(int index) {
+        int backCounter = size();
+        Node<T> current = tail;
+        while (--backCounter > index) {
+            current = current.previous;
+        }
+        return current;
```
NazarSmith228, line 106: you could rewrite this with a `for` loop - this way you don't need additional `backCounter` variable

```suggestion
        Node<T> current = tail;
        for(int i = size - 1; i > index; i--)
            current = current.previous;
        return current;
```

same can be applied to `fromHead` method

however this remark is minor - you are to decide whether apply it or not :)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
+            current = current.next;
+        }
+        return current;
+    }
+
+    private Node<T> getNodeByValue(T value) {
+        Node<T> current = head;
+        while (current != null) {
+            if (current.value == value || current.value != null
+                    && current.value.equals(value)) {
```
NazarSmith228, line 122: this is weird condition - no need to do reference check, consider using `equals()`

therefore all you need is `current.value.equals(value)`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
+        }
+        tail = newNode;
+        if (head == null) {
+            head = newNode;
+        }
+        count++;
     }
 
     @Override
     public void add(T value, int index) {
```
NazarSmith228, line 24: one important question here - does your implementation allow `null` values?

my personal suggestion is to forbid them, since if you have null elements - it may be hard to implement method like `contains(element)` or `remove(element)`

**the same should be considered within `add(value)` method**

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
+        }
+        return current;
+    }
+
+    private Node<T> getNodeByValue(T value) {
+        Node<T> current = head;
+        while (current != null) {
+            if (current.value == value || current.value != null
+                    && current.value.equals(value)) {
+                break;
```
NazarSmith228, line 123: there is no sense to use `break` when you can return `true` immediately

at the end you can return `false` explicitly

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
     @Override
     public int size() {
-        return 0;
+        return count;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size() == 0;
```
NazarSmith228, line 79: you should not use method here, refer directly to field

`size()` -> `size`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
+    }
+
+    private void checkIndexToGet(int index) {
+        if (index < 0 || index >= size()) {
```
NazarSmith228, line 83: you should not use method here, refer directly to field

`size()` -> `size`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
+            throw new IndexOutOfBoundsException("Invalid index " + index + " for size " + size());
+        }
+    }
+
+    private void checkIndexToAdd(int index) {
+        if (index < 0 || index > size()) {
```
NazarSmith228, line 89: you should not use method here, refer directly to field

`size()` -> `size`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
+        checkIndexToGet(index);
+        Node<T> toRemove = getNodeByIndex(index);
+        unlink(toRemove);
+        return toRemove.value;
     }
 
     @Override
     public boolean remove(T object) {
-        return false;
+        return unlink(getNodeByValue(object));
```
NazarSmith228, line 69: `unlink` method shouldn't return any `boolean` value - it either unlinks a node or not

instead I would get rid of `getNodeByValue` method, since it is used only here and declare a method like `boolean containsNode(object)`

this ways you would be able to check if such node exists and then do unlink

benefits - code becomes more clear and you don't need to deal with null values

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
+    private Node<T> getNodeFromHead(int index) {
+        int counter = 0;
+        Node<T> current = head;
+        while (counter++ < index) {
+            current = current.next;
+        }
+        return current;
+    }
+
+    private Node<T> getNodeByValue(T value) {
```
NazarSmith228, line 118: as I mentioned above - this method can be reworked into `boolean containsNode(T value)`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
+            }
+            current = current.next;
+        }
+        return current;
+    }
+
+    private boolean unlink(Node<T> node) {
+        if (node == null) {
+            return false;
+        }
```
NazarSmith228, line 133: due to the comments regarding `containsNode()` method - you don't need to perform this check here after applying those changes

NazarSmith228 (changes requested):
Great job! Now let's improve your solution :)

src/main/java/core/basesyntax/MyLinkedList.java
ystankevych: This condition current.value == value not only for reference check. This condition returns true if value is null and we need to find node with such value. So,  I leave it as it is. Cos check like if value == null && current.value == null looks not good

src/main/java/core/basesyntax/MyLinkedList.java
ystankevych: I need such method getNodeByValue. Cos in public boolean remove(T object) I have to find Node (param is just Value)

Alexsandr-Yablunovskyi (approved):
Mate Academy platform sent me to review your code. But I think it's a little bit late) So, what can I say - good job!

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,154 @@
...
+        head.previous = toAdd;
+        head = toAdd;
+        size++;
+    }
+
+    private String outOfBoundMessage(int index) {
+        return String.format("Index %d out of bound for size %d", index, size);
+    }
+
+    class Node<T> {
```
liliia-ponomarenko, line 145: Should  be `private static `

liliia-ponomarenko (approved):
Great 🔥

PR #1140, opened 2023-08-19
Initial commit

gordgelin:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,171 @@
...
+
+        if (this.size == 0) {
+            return false;
+        }
+
+        do {
+            if ((object == null)
+                    ? (currentNode.value == null)
+                    : (object.equals(currentNode.value))) {
+                updateHeadAndTail(currentNode, currentIndex);
```
aaananas, line 97: move this logic to unlink method, you can check if the node is head or tail by comparing reference ```node == head```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,171 @@
...
+
+        if (node.next != null) {
+            node.next.prev = node.prev;
+        }
+
+        this.size--;
+    }
+
+    private Node<T> getNodeByIndex(int index) {
+        boolean headOrTail;
```
aaananas, line 133:
```suggestion
        boolean iterateFromHead;
```
better name would be iterateFromHead,
but `if-else` would be much easier to understand here. Instead of checking from which side to iterate you are checking it all the time in while loop

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,171 @@
...
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        Node<T> node = getNodeByIndex(index);
+        T oldValue;
+
+        oldValue = node.value;
+        updateHeadAndTail(node, index);
```
aaananas, line 78: move this logic to unlink method, you can check if the node is head or tail by comparing reference ```node == head```

aaananas (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
gordgelin: agree, i just wanted to do it this way. anyway i do not think this additional branches will have much impact on byte code level in this case.

src/main/java/core/basesyntax/MyLinkedList.java
gordgelin: +

src/main/java/core/basesyntax/MyLinkedList.java
gordgelin: +

olekskov (approved):
Good job

PR #1142, opened 2023-08-20
implemented MyLinkedList class

mmosii:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,169 @@
...
+        checkIndex(index);
+        return find(index).value;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        checkIndex(index);
+        T oldValue = find(index).value;
+        find(index).value = value;
```
RedArtelerist, line 58: You call the function twice, which means you do double work. Get the result into a separate variable, and then work with it.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,169 @@
...
+
+    public MyLinkedList() {
+        size = 0;
+        head = null;
+        tail = null;
+    }
+
     @Override
     public void add(T value) {
+        Node<T> node = new Node<>(null, value, null);
```
RedArtelerist, line 18: You can simplify `else` statement if you immediately add to prev field in constructor tail-node
```suggestion
        Node<T> node = new Node<>(tail, value, null);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,169 @@
...
+        if (size == 0) {
+            head = node;
+            tail = node;
+        } else {
+            node.prev = tail;
+            tail = node;
+            node.prev.next = tail;
```
RedArtelerist, line 25: And in this place you can bind the `next` of tail

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,169 @@
...
+        }
+        for (int i = 0; i < index; i++) {
+            node = node.next;
+        }
+        return node;
+    }
+
+    private void link(T value, int index) {
+        Node<T> next = find(index);
+        Node<T> newNode = new Node<>(null, value, null);
```
RedArtelerist, line 130: Again, you can immediately initialize the next and the first to do fewer actions later.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,169 @@
...
+        if (index > (size >> 1)) {
+            node = tail;
+            for (int i = size - 1; i > index; i--) {
+                node = node.prev;
+            }
+            return node;
+        }
+        for (int i = 0; i < index; i++) {
+            node = node.next;
+        }
```
RedArtelerist, line 124: You can put it in two separate functions

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,169 @@
...
+    }
+
+    private Node<T> find(int index) {
+        Node<T> node = head;
+        if (index == 0) {
+            return node;
+        }
+        if (index == size - 1) {
+            return tail;
+        }
```
RedArtelerist, line 114: I don't think this check is necessary, if you allocate two separate functions for looping from the beginning and end, you just won't enter the loop, but will return head or tail at once, so you can reduce your code.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,169 @@
...
+        return oldValue;
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> newNode = new Node<>(null, object, null);
+        for (int i = 0; i < size; i++) {
+            if (newNode.value == find(i).value
+                    || newNode.value != null && newNode.value.equals(find(i).value)) {
+                unlink(find(i));
```
RedArtelerist, line 77: Again, you call `find` 3 times

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,169 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void checkIndex(int index) {
+        if (!(index >= 0 && index < size)) {
```
RedArtelerist, line 96: Maybe in this way is clearer to write
```suggestion
        if (index < 0 || index >= size) {
```
Agree, it looks much nicer this way.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,169 @@
...
+        } else {
+            newNode.prev = next.prev;
+            next.prev.next = newNode;
+            next.prev = newNode;
+            newNode.next = next;
+        }
+    }
+
+    private void unlink(Node<T> node) {
+        if (node.equals(head)) {
```
RedArtelerist, line 144: I don't sure, but i think in this place we can compare references 
```suggestion
        if (node == head) {
```

RedArtelerist (changes requested):
Good job, try improve your solution.

src/main/java/core/basesyntax/MyLinkedList.java
mmosii: yeah, it works)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -72,9 +72,10 @@ public T remove(int index) {
     public boolean remove(T object) {
         Node<T> newNode = new Node<>(null, object, null);
         for (int i = 0; i < size; i++) {
-            if (newNode.value == find(i).value
-                    || newNode.value != null && newNode.value.equals(find(i).value)) {
-                unlink(find(i));
+            Node<T> node = find(i);
```
RedArtelerist, line 75: You call find method in each iteration, this is not a good solution. You need only once find node and then remove it. For example, you can write a separate method that returns node by value.

RedArtelerist (changes requested):
Almost everything is fixed, it remains to fix one moment.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,169 @@
...
+        checkIndex(index);
+        T oldValue = find(index).value;
+        unlink(find(index));
+        size--;
+        return oldValue;
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> newNode = new Node<>(null, object, null);
```
RedArtelerist, line 73: You could not create a `Node` instanse and just compare it to `object`, but in principle it's not a big deal.

RedArtelerist (approved):
Now, I think everything is good, nice work. Also, I added some comment.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,162 @@
...
+        }
+        return node;
+    }
+
+    private void link(T value, int index) {
+        Node<T> next = find(index);
+        Node<T> newNode = new Node<>(null, value, next);
+        if (size > 0 && index == 0) {
+            next.prev = newNode;
+            head = newNode;
```
NazarSmith228, line 128: the problem here is that your `link` method for some reason works with head, however it shouldn't 

`link` should be used generally for linking a node at a particular position - this is covered by `else` statement

the reason why you handle `head ` here lies behind your implementation of `add(value, index)` method - try to refactor it (think of additional methods like `linkFirst` and `linkLast`)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,162 @@
...
+                head = node.next;
+            }
+        } else if (node == tail) {
+            if (size > 1) {
+                node.prev.next = null;
+                tail = node.prev;
+            }
+        } else {
+            node.prev.next = node.next;
+            node.next.prev = node.prev;
```
NazarSmith228, line 149: this implementation is so complex that you made a mistake :)

what will happen if `node == head` but `size !> 1`? 

exactly - this method will do nothing :)

if you got into `if` - you won't get into `else if` or `else`

try to refactor this method, make it simpler to avoid such mistakes

NazarSmith228 (changes requested):
Amazing implementation! Let's improve it :)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,162 @@
...
+        unlink(find(index));
+        size--;
+        return oldValue;
     }
 
     @Override
     public boolean remove(T object) {
+        for (Node<T> node = head; node != null; node = node.next) {
+            if (object == node.value
+                    || object != null && object.equals(node.value)) {
```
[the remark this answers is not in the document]
mmosii: we have null's added to list in tests, catching NullPointerException if i leave only node.value.equals(object)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,173 @@
...
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        checkIndexInBoundsStrict(index);
+        if (index == 0) {
+            linkFirst(value);
+        } else if (index > 0 && index == size) {
```
NazarSmith228, line 25: can't it be simplified to just `index == size`?

src/main/java/core/basesyntax/MyLinkedList.java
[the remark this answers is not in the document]
NazarSmith228: weird, but okay...

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,173 @@
...
+        for (int i = 0; i < index; i++) {
+            node = node.next;
+        }
+        return node;
+    }
+
+    private void linkFirst(T value) {
+        Node<T> node = new Node<>(null, value, head);
+        if (size > 0) {
+            node.next.prev = node;
```
NazarSmith228, line 126: why not `head.prev = node`?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,173 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void checkIndexInBounds(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException(outOfBoundsMessage() + index);
```
NazarSmith228, line 90: you can change the signature of `outOfBoundsMessage()` to `outOfBoundsMessage(int index)` to avoid concatenation

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,173 @@
...
+    }
+
+    private void unlink(Node<T> node) {
+        if (node == head) {
+            head = node.next;
+        } else {
+            node.prev.next = node.next;
+        }
+        if (node == tail) {
+            node.prev = tail;
```
NazarSmith228, line 154: seems you made a mistake here

shouldn't it be inversed - `tail = node.prev`?

NazarSmith228 (changes requested):
Great! Just some final remarks to address

Andriy-dz (approved):
Very good solution!

PR #1146, opened 2023-08-22
MyLinkedList is implemented.

Alexsandr-Yablunovskyi:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,212 @@
...
+        if (size == 1) {
+            return unlinkTheOne();
+        } else if (index == 0) {
+            return unlinkFirst();
+        } else if (index == size - 1) {
+            return unlinkLast();
+        } else {
+            return unlinkWithin(index);
+        }
+    }
```
liliia-ponomarenko, line 128: Can you simplify this method instead of having a separate logic for each  case?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,212 @@
...
+        private T value;
+        private Node<T> prev;
+        private Node<T> next;
+
+        public Node(Node<T> prev, T value, Node<T> next) {
+            this.value = value;
+        }
     }
 }
+
```
liliia-ponomarenko, line 214: Redundant  empty line

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,212 @@
...
+        }
+    }
+
+    private static class Node<T> {
+        private T value;
+        private Node<T> prev;
+        private Node<T> next;
+
+        public Node(Node<T> prev, T value, Node<T> next) {
+            this.value = value;
```
liliia-ponomarenko, line 210: Let’s  initialize prev and next

liliia-ponomarenko (changes requested):

liliia-ponomarenko (changes requested):
Good! Now get try simplify  your solution ;)

src/main/java/core/basesyntax/MyLinkedList.java
Alexsandr-Yablunovskyi: I'm not sure I've got you right) I simplified 'if-else' construction by removing the redundant 'else-if' condition.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,214 @@
...
+
+    private T unlinkWithin(int index) {
+        T removedValue;
+        Node<T> previous = getNodeByIndex(index - 1);
+        removedValue = previous.next.value;
+        previous.next = previous.next.next;
+        previous.next.prev = previous;
+        size--;
+        return removedValue;
+    }
```
torbinofficial, line 165: Why are we finding previous node, not with our index?
That's not an 1 linked list.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,214 @@
...
+        int index = -1;
+        Node<T> current = head;
+        int iteration = 0;
+        while (current != null) {
+            if (current.value == object || current.value != null && current.value.equals(object)) {
+                return iteration;
+            }
+            current = current.next;
+            iteration++;
+        }
```
torbinofficial, line 114: You can do this with for loop.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,214 @@
...
+        return getNodeByIndex(index).value;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        checkIndex(index);
+        T oldValue = getNodeByIndex(index).value;
+        getNodeByIndex(index).value = value;
+        return oldValue;
```
torbinofficial, line 45: We call for same function twice. Can't we just create a new Node object to work with?

torbinofficial (changes requested):

torbinofficial (approved):
Nice job :).
Now to the mentor review

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,213 @@
...
+        for (int i = 0; i < size; i++) {
+            if (current.value == object || current.value != null && current.value.equals(object)) {
+                return i;
+            }
+            current = current.next;
+        }
+        return index;
+    }
+
+    private T unlink(int index) {
```
Elena-Bruyako, line 117: You can simplify this logic and avoid duplication of code by providing Node to the unlink method and do some checking: if prev == null else ..., if next == null else ... and then set node.value = null, decrease size and return value

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,213 @@
...
+    private int size;
+
     @Override
     public void add(T value) {
+        linkLast(value);
     }
 
     @Override
     public void add(T value, int index) {
+        checkAddIndex(index);
```
Elena-Bruyako, line 17: You can create new Node here once and avoid doing it in other three methods 

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,213 @@
...
+        if (index == size) {
+            linkLast(value);
+        } else if (index == 0) {
+            linkFirst(value);
+        } else {
+            link(value, index);
+        }
```
Elena-Bruyako, line 24: You can add size++; here and remove from these three methods

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,190 @@
...
+            head.prev = null;
+            size--;
+            return deletedNode.value;
+        }
+        if (deletedNode.next == null) {
+            tail.prev.next = null;
+            tail = tail.prev;
+            size--;
+            return deletedNode.value;
+        }
```
Elena-Bruyako, line 132: size--; can be placed here and removed from if statements and unlinkWithin method

Elena-Bruyako (approved):

PR #1147, opened 2023-08-23
created MyLinckedList

teract10s:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,202 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size = 0;
```
liliia-ponomarenko, line 6:
```suggestion
    private int size;
```
0 - default value for int

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,202 @@
...
+        Node<T> newNode = new Node<>(null, value, null);
+        if (size == 0) {
+            head = newNode;
+            tail = newNode;
+            size++;
+            return;
+        }
+        tail.next = newNode;
+        newNode.prev = tail;
+        tail = newNode;
```
liliia-ponomarenko, line 21:
```suggestion
        Node<T> newNode = new Node<>(tail, value, null);
        if (size == 0) {
            head = newNode;
        } else {
            tail.next = newNode;
        }
        tail = newNode;
```
you can simplify  it

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,202 @@
...
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        indexCheck(index);
+        Node<T> node = getNodeByIndex(index);
+        if (node == null) {
+            return null;
+        }
```
liliia-ponomarenko, line 69: Does it work correctly?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,202 @@
...
+        Node<T> nodeToRemove = head;
+        while (nodeToRemove != null) {
+            if (object == nodeToRemove.element
+                    || object != null && object.equals(nodeToRemove.element)) {
+                break;
+            }
+            nodeToRemove = nodeToRemove.next;
+        }
+        unlink(nodeToRemove);
+        return nodeToRemove == null ? false : true;
```
liliia-ponomarenko, line 93: What will be returned if you delete the null value?

liliia-ponomarenko, line 93:
```suggestion
        Node<T> nodeToRemove = head;
        while (nodeToRemove != null) {
            if (object == nodeToRemove.element
                    || object != null && object.equals(nodeToRemove.element)) {
                unlink(nodeToRemove);
                return true;
            }
            nodeToRemove = nodeToRemove.next;
        }
        return false;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,202 @@
...
+        if (nodeToRemove.prev == null) {
+            deleteFirstNode();
+            return nodeToRemove.element;
+        }
+        if (nodeToRemove.next == null) {
+            deleteLastNode();
+            return nodeToRemove.element;
+        }
+        deleteNode(nodeToRemove);
+        return nodeToRemove.element;
```
liliia-ponomarenko, line 119: Can you simplify this method instead of having a separate logic for each case? delete the first and last nodes very similar

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,202 @@
...
+    }
+
+    private Node<T> getNodeByIndex(int index) {
+        return index > size / 2 ? goFromTailToHead(index) : goFromHeadToTail(index);
+    }
+
+    private Node<T> goFromTailToHead(int index) {
+        int currentIndex = size - 1;
+        Node<T> currentNode = tail;
+        while (currentNode != null) {
```
liliia-ponomarenko, line 170: you can use `for` loop for simplifying  this logic

liliia-ponomarenko (changes requested):
Good job! Now let’s simplify your solution!

Elena-Bruyako (approved):

PR #1151, opened 2023-08-25
implemented MyLinkedList class

Vt0rtikV:


VolodymyrSkvira (approved):
LGTM as we discussed the solution while implementing it

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,160 @@
...
+        }
+        current = head;
+        for (int i = 0; i < index; i++) {
+            current = current.next;
+        }
+        return current;
+    }
+
+    private static class Node<T> {
+
```
liliia-ponomarenko, line 151: Redundant  empty line

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,160 @@
...
-        return false;
+        Node<T> current = head;
+        while (current != null && !isEqual(object, current.item)) {
+            current = current.next;
+        }
+        if (current == null) {
+            return false;
+        }
+        unlink(current);
+        return true;
```
liliia-ponomarenko, line 85: you can simplify  it in this way:
``` 
while (current != null) {
   if (isEqual(object, current.item)) {
                unlink(current);
                return true;
                }
            current = current.next;
        }
        return false;```

File: src/main/java/core/basesyntax/MyLinkedList.java [hunk 3]
```diff
@@ -3,45 +3,160 @@
...
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private boolean isEqual(T object, T currentObject) {
+        return object == currentObject || object != null && object.equals(currentObject);
+    }
+
+    private void unlink(Node<T> node) {
```
liliia-ponomarenko, line 102: You can move common code outside  if blocks  
for example : 
```
        Node<T> next = node.next;
        Node<T> prev = node.prev;
        if (prev == null) {
            first = next;
        } else {
            prev.next = next;
            node.prev = null;
        }
        if (next == null) {
            last = prev;
        } else {
            next.prev = prev;
            node.next = null;
        }
        T value = node.value;
        node.value = null;
        size--;
        return value;
    }
```

liliia-ponomarenko (changes requested):
Good job! Now let’s simplify your solution ;)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,152 @@
...
+            Node<T> newNode = new Node<>(null, value, null);
+            head = newNode;
+            tail = newNode;
+            size++;
+            return;
+        }
+        Node<T> newNode = new Node<>(tail, value, tail.next);
+        tail.next = newNode;
+        tail = newNode;
+        size++;
```
Andriy-dz, line 22:
```suggestion
        if (isEmpty()) {
            head = tail = new Node<>(null, value, null);
        } else {
        Node<T> newNode = new Node<>(tail, value, tail.next);
        tail.next = newNode;
        tail = newNode;
        }
        size++;
```

Andriy-dz (approved):

PR #1152, opened 2023-08-25
Implemented required methods

VolodymyrSkvira:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
     @Override
     public void add(T value) {
+        Node<T> node = new Node<>(value);
+        if (size == 0) {
+            head = node;
+            tail = node;
```
sonyasapsan, line 15: you can do it in one line:
head = tail = node; 
but it's up to you!

sonyasapsan (approved):
Good job!

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
+    }
+
+    private static class Node<T> {
+        private T value;
+        private Node<T> next;
+        private Node<T> prev;
+
+        private Node(T value) {
+            this.value = value;
+        }
```
liliia-ponomarenko, line 125: better make it all arg constructor it can help simplify  your solution and move it after private methods

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
+        Node<T> current = head;
+        while (current != null
+                && !(object == current.value || object != null && object.equals(current.value))) {
+            current = current.next;
+        }
+        if (current == null) {
+            return false;
+        }
+        unlink(current);
+        return true;
```
liliia-ponomarenko, line 88:
```suggestion
        while (current != null) {
                if (object == current.value || object != null && object.equals(current.value)) {
                unlink(current);
                return true;
                }
            current = current.next;
        }
        return false;
```
Let’s  simplify this code

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
+
+    }
+
+    private void checkIndex(int index, int size) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Wrong index: " + index);
+        }
+    }
+
+    private void unlink(Node<T> current) {
```
liliia-ponomarenko, line 135: You can move common code outside  if blocks  
for example : 
```
        Node<T> next = node.next;
        Node<T> prev = node.prev;
        if (prev == null) {
            first = next;
        } else {
            prev.next = next;
            node.prev = null;
        }
        if (next == null) {
            last = prev;
        } else {
            next.prev = prev;
            node.next = null;
        }
        T value = node.value;
        node.value = null;
        size--;
        return value;
    }
```

liliia-ponomarenko (changes requested):
Good job! Let’s  simplify  your solution ;)

src/main/java/core/basesyntax/MyLinkedList.java
Andriy-dz: don't fixed

src/main/java/core/basesyntax/MyLinkedList.java
Andriy-dz: agree with @sonyasapsan

Andriy-dz (changes requested):

Andriy-dz (approved):

PR #1153, opened 2023-08-25
implemented methods of MyLinkedListInterface

olkspdhko:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,149 @@
...
+            Node<T> current = getNodeByIndex(index);
+            Node<T> nodeNew = new Node<>(current.prev, value, current);
+            nodeNew.prev.next = nodeNew;
+            current.prev = nodeNew;
+        }
+        size++;
+    }
+
+    private Node<T> getNodeByIndex(int index) {
+        Node<T> current = head;
```
sokimaaa, line 55: Don’t iterate your nodes from the head to tail or vice versa.

see https://mate-academy.github.io/jv-program-common-mistakes/java-core/collections/linked-list.html

sokimaaa, line 55: why not to check the index out of the bound inside the getNodeByIndex(..) method? 
It will reduce code duplication.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,149 @@
...
     public void addAll(List<T> list) {
+        for (T entry : list) {
+            add(entry);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        ifIndexOutOfBounds(index, size - 1);
```
sokimaaa, line 71: simplify get() method after getNodeByIndex method refactoring

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,149 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    public void ifIndexOutOfBounds(int index, int upperBound) {
+        if (index < 0 || index > upperBound) {
```
sokimaaa, line 132: upper bound always equals `size - 1`, doesn't it? 
So, we don't need the `upperBound` method param.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,149 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    public void ifIndexOutOfBounds(int index, int upperBound) {
```
sokimaaa, line 131: why this method is public?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,149 @@
...
+        Node<T> current = getNodeByIndex(index);
+        T valueToReturn = current.value;
+        unlink(current);
+        size--;
+        return valueToReturn;
     }
 
     @Override
     public boolean remove(T object) {
+        if (isEmpty()) {
```
sokimaaa, line 105: redundant if clause, because if size == 0, then the for-loop isn't going to iterate.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,149 @@
...
+            return false;
+        }
+        Node<T> current = head;
+        for (int i = 0; i < size; i++) {
+            if (current.value == object || (current.value != null
+                    && current.value.equals(object))) {
+                unlink(current);
+                size--;
```
sokimaaa, line 113: why not to move `size--` inside `unlink(..)` method? 
The unlink method is responsible for deleting nodes, so the size should always be decreased in such cases.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,149 @@
...
+        Node<T> current = getNodeByIndex(index);
+        valueToReturn = current.value;
+        current.value = value;
+        return valueToReturn;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        ifIndexOutOfBounds(index, size - 1);
```
sokimaaa, line 95: refactor this method after changing `ifIndexOutOfBounds` and `unlink` methods

sokimaaa (changes requested):

aaananas (approved):

PR #1154, opened 2023-08-25
implementation of LinkedList

sonyasapsan:


vsych321 (approved):
LGTM! But I think that  u can simplify this code by adding some methods,(in case of better readability), for instance, adding first  and last elements, etc.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,153 @@
...
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private int getNodeIndexByValue(T value) {
+        Node<T> node = head;
+        for (int i = 0; i < size; i++) {
+            if ((value == null && value == node.value)
+                    || value != null && value.equals(node.value)) {
```
liliia-ponomarenko, line 122:
```suggestion
            if (value == node.value
                    || value != null && value.equals(node.value)) {
```
Let’s  simplify  conditional

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,153 @@
...
+    }
+
+    private void checkIndex(int index) {
+        if ((size == 0 && index != 0)
+                || (index >= size || index < 0)) {
+            throw new IndexOutOfBoundsException("Incorrect index");
+        }
     }
 }
+
```
liliia-ponomarenko, line 155: Redundant empty line

liliia-ponomarenko (changes requested):
good job! Let’s  simplify  your solution ;)

src/main/java/core/basesyntax/MyLinkedList.java
Andriy-dz: not fixed

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,165 @@
...
+            tail = node.prev;
+        } else {
+            node.next.prev = node.prev;
+            node.prev.next = node.next;
+        }
+    }
+
+    private void checkIndex(int index, int size) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Incorrect index");
```
Andriy-dz, line 149: add more info to message

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,165 @@
...
+            head = node;
+        } else if (index == size) {
+            add(value);
+            size--;
+        } else {
+            Node<T> prev = getNodeByIndex(index - 1);
+            Node<T> node = new Node<>(prev, value, prev.next);
+            prev.next.prev = node;
+            prev.next = node;
+        }
```
Andriy-dz, line 38:
```suggestion
        if (index == size) {
            add(value);
            return;
        } 
        if (tail == null && head == null) {
            addFirst(value);
        } else if (index == 0) {
            Node<T> node = new Node<>(null, value, head);
            head.prev = node;
            head = node;
        } else {
            Node<T> prev = getNodeByIndex(index - 1);
            Node<T> node = new Node<>(prev, value, prev.next);
            prev.next.prev = node;
            prev.next = node;
        }
```

Andriy-dz: in this way, you will get rid of the if else construction

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,165 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private int getNodeIndexByValue(T value) {
```
Andriy-dz, line 95:
```suggestion
    private int getIndexByValue(T value) {
```

Andriy-dz (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,166 @@
...
 
     @Override
     public void add(T value, int index) {
+        checkIndex(index, size + 1);
+        if (index == size) {
+            add(value);
+            return;
+        }
+        if (tail == null && head == null) {
+            addFirst(value);
```
NazarSmith228, line 29: in case if `index == 0` and both `head` and `tail` are null these 2 `if` statements are the same

so basically you can remove the second `if` statement

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,166 @@
...
     }
 
     @Override
     public boolean remove(T object) {
-        return false;
+        int index = getIndexByValue(object);
+        if (index == -1) {
+            return false;
+        }
+        unlink(getNodeByIndex(index));
```
NazarSmith228, line 81: there is no sense to do so much stuff here - you iterate first to find an index and then you iterate to find a node

why not using a single for loop here to find a desired node and then call `unlink`?

basically you can remove `getIndexByValue` and use the loop from that method here

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,166 @@
...
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private int getIndexByValue(T value) {
+        Node<T> node = head;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(value, node.value)) {
```
NazarSmith228, line 99: don't use libraries :)

src/test/java/core/basesyntax/MyLinkedListTest.java
```diff
@@ -72,7 +72,7 @@ public void testAddManyData() {
 
     @Test
     public void testAddByIndexToTheTop() {
-        myLinkedList.add(FIRST_ITEM, 0);
+         myLinkedList.add(FIRST_ITEM, 0);
```
NazarSmith228, line 75: you should not modify test classes in any sense

NazarSmith228 (changes requested):
good job! let's improve your solution now

src/main/java/core/basesyntax/MyLinkedList.java
liliia-ponomarenko: not fixed

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
+        return node.value;
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> node = head;
+        for (int i = 0; i < size; i++) {
+            if ((object == node.value) || (object != null && object.equals(node.value))) {
+                unlink(node);
+                size--;
```
liliia-ponomarenko, line 80: move  size-- to unlink()

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
+        unlink(node);
+        size--;
+        return node.value;
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> node = head;
+        for (int i = 0; i < size; i++) {
+            if ((object == node.value) || (object != null && object.equals(node.value))) {
```
liliia-ponomarenko, line 78:
```suggestion
            if (object == node.value || object != null && object.equals(node.value)) {
```
redundant brackets (read about operators' priority)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private int getIndexByValue(T value) {
```
liliia-ponomarenko, line 98: you don’t use this method. Let’s remove it ;)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,165 @@
...
+            tail = node.prev;
+        } else {
+            node.next.prev = node.prev;
+            node.prev.next = node.next;
+        }
+    }
+
+    private void checkIndex(int index, int size) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Particular index is out of bounds for the size");
```
liliia-ponomarenko, line 152:
```suggestion
            throw new IndexOutOfBoundsException("Particular index is out of bounds for the size");
```
```suggestion
            throw new IndexOutOfBoundsException("Particular index:" + index + " is out of bounds for the size");
```

liliia-ponomarenko (changes requested):
Not all comments were fixed. Be attentive ;)

NazarSmith228 (approved):
GTG

PR #1155, opened 2023-08-26
implemented all the methods within Linkedlist, added my own methods f…

vsych321:
…or readability

### DIFF
--- a/src/main/java/core/basesyntax/MyLinkedList.java
+++ b/src/main/java/core/basesyntax/MyLinkedList.java
@@ -3,45 +3,158 @@ package core.basesyntax;
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> first;
+    private Node<T> last;
+    private int size;
+
     @Override
     public void add(T value) {
+        Node<T> newNode = new Node<>(null, value, null);
+        if (isEmpty()) {
+            first = last = newNode;
+        } else {
+            addLast(value);
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        Node<T> newNode = new Node<>(null, value, null);
+        checkPositionByElement(index);
+        if (isEmpty()) {
+            first = last = newNode;
+        } else if (index == 0) {
+            addFirst(value);
+        } else if (index == size) {
+            addLast(value);
+        } else {
+            Node<T> prev = getNodeByIndex(index - 1);
+            Node<T> node = new Node<>(prev, value, prev.next);
+            prev.next.prev = node;
+            prev.next = node;
+        }
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
+        for (T values : list) {
+            add(values);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkIndex(index);
+        return getNodeByIndex(index).value;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        checkIndex(index);
+        Node<T> newNode = getNodeByIndex(index);
+        T previous = newNode.value;
+        newNode.value = value;
+        return previous;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        Node<T> node = getNodeByIndex(index);
+        unlink(node);
+        size--;
+        return node.value;
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> currentElement = first;
+        while (currentElement != null) {
+            if (currentElement.value == object
+                    || currentElement.value != null && currentElement.value.equals(object)) {
+                unlink(currentElement);
+                size--;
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
+            first = last = newNode;
+        } else if (index == 0) {
+            addFirst(value);
+        } else if (index == size) {
+            addLast(value);
+        } else {
+            Node<T> prev = getNodeByIndex(index - 1);
+            Node<T> node = new Node<>(prev, value, prev.next);
+            prev.next.prev = node;
+            prev.next = node;
```
validatorr, line 35: you can implement this in seperate method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
+            first = last = newNode;
+        } else {
+            addLast(value);
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        Node<T> newNode = new Node<>(null, value, null);
```
validatorr, line 23: I would recommend you to put initialization of object in seperate method, but it's up to you

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> first;
+    private Node<T> last;
+    private int size;
+
     @Override
     public void add(T value) {
+        Node<T> newNode = new Node<>(null, value, null);
+        if (isEmpty()) {
+            first = last = newNode;
```
validatorr, line 14: you can use your addFirst method in here

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
+
+    private void addLast(T element) {
+        Node<T> newNode = new Node<>(last, element, null);
+        last.next = newNode;
+        last = newNode;
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Can't add value by index" + index);
```
validatorr, line 110: you can add some different information in exception message, so we could differ it from checkPositionByElement

validatorr (changes requested):
Great work! Only some minor fixes to make

src/main/java/core/basesyntax/MyLinkedList.java
vsych321: my addFirst method differs from that u want to use, cz I initialize it without > 0 and etc

src/main/java/core/basesyntax/MyLinkedList.java
vsych321: it has only 1 line  2 usages, so I think, there is no need, but tnx for this attention!

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,162 @@
...
+        return previous;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        Node<T> node = getNodeByIndex(index);
+        unlink(node);
+        size--;
```
liliia-ponomarenko, line 64: you can move it to unlink()

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,162 @@
...
+            addLast(value);
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        Node<T> newNode = new Node<>(null, value, null);
+        checkPositionByElement(index);
+        if (isEmpty()) {
```
liliia-ponomarenko, line 25: isEmpty it the same that index == size so you can just call method add(T value) 
and [Don’t complicate if-else construction. ](https://mate-academy.github.io/jv-program-common-mistakes/java-core/collections/linked-list.html#dont-complicate-if-else-construction-detailed-explanation)

liliia-ponomarenko (changes requested):
Good job! Let’s  fix a few mistakes  ;)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,159 @@
...
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        checkPositionByElement(index);
+        if (index == size) {
+            add(value);
+            return;
+        } else if (index == 0) {
```
Andriy-dz, line 27: why you use if else construction.
```suggestion
        }
        if (index == 0) {
```

Andriy-dz (approved):
Good Solution!

PR #1161, opened 2023-08-27
HW linked list

OlehReex:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,143 @@
...
+            node.prev.next = node.next;
+        }
+        if (node == tail) {
+            tail = node.prev;
+        } else {
+            node.next.prev = node.prev;
+        }
+    }
+
+    private static class Node<T> {
```
anigilator479, line 136: I think better create this class in the beginning

anigilator479 (changes requested):

anigilator479 (approved):

src/main/java/core/basesyntax/MyLinkedList.java
liliia-ponomarenko: no, private classes should  be placed  after public and private method main class

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,143 @@
...
+        private T value;
+        private Node<T> next;
+        private Node<T> prev;
+
+        public Node(T value) {
```
liliia-ponomarenko, line 141: [When creating a new Node, pass all the fields to Node constructor](https://mate-academy.github.io/jv-program-common-mistakes/java-core/collections/linked-list.html#when-creating-a-new-node-pass-all-the-fields-to-node-constructor)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,143 @@
...
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndexValidation(index, size);
+        Node<T> removedNode = getNodeByIndex(index);
+        T removedValue = removedNode.value;
+        removeLink(removedNode);
+        size--;
```
liliia-ponomarenko, line 73: you can move it to removeLink()

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,143 @@
...
+            newNode.prev = tail;
+            tail.next = newNode;
+            tail = newNode;
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        checkIndexValidation(index, size + 1);
```
liliia-ponomarenko, line 25:
```suggestion
        checkIndex(index, size + 1);
```
use verbs for method names

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,143 @@
...
+        Node<T> newNode = new Node<>(value);
+        if (isEmpty()) {
```
liliia-ponomarenko, line 27: isEmpty it the same that index == size so you can just call method add(T value) 
and [Don’t complicate if-else construction. ](https://mate-academy.github.io/jv-program-common-mistakes/java-core/collections/linked-list.html#dont-complicate-if-else-construction-detailed-explanation)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,143 @@
...
+            addIfFirstNode(new Node<>(value));
+        } else if (index == 0) {
+            newNode.next = head;
+            head = newNode;
+        } else if (index == size) {
+            add(value);
+            size--;
```
liliia-ponomarenko, line 34: why did you reduce the size here?  It looks like a hardcode. Let’s  think about how you can refactor your code to avoid it

liliia-ponomarenko (changes requested):
Good job! Let’s  simplify  your solution ;)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,142 @@
...
 
     @Override
     public void add(T value, int index) {
+        checkIndex(index, size + 1);
+        Node<T> newNode = new Node<>(null, value, null);
+        if (index == size) {
+            add(value);
+            return;
+        } else if (index == 0) {
+            newNode.next = head;
```
Andriy-dz, line 31:
```suggestion
        }
        if (index == 0) {
            newNode.next = head;
```
so we will get rid of the if else construction

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,142 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void addIfFirstNode(T value) {
```
Andriy-dz, line 95: is there really a need for a separate method for this

Andriy-dz (approved):

PR #1166, opened 2023-08-27
Implement MyLinkedList

KDanil2000:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
+        }
+        return current;
+    }
+
+    private static class Node<T> {
+        private T value;
+        private Node<T> next;
+        private Node<T> prev;
+
+        private Node(T value) {
```
SalnikovMarkiian, line 156: i think we should initialize all the fields in the constructor, that way the whole code will be a bit more readable

SalnikovMarkiian (changes requested):
in general, your solution works, but it can definitely be improved. Let's start with a simple one)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
...
+            tail = node;
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        checkAddedIndex(index);
+        Node<T> newNode = new Node<>(value);
+        if (head == null) {
```
SalnikovMarkiian, line 27: Here we can immediately check that the index will be of different size (and in this case we will call the previous add(value)). Further, instead of complex if...else constructions (typical errors), we can use an iterator in the core method that will return us the node by index( getNodeByIndex) and assign it to a new variable, for example newNext . Also, in this case, it will be useful for us to create newPred, which will contain the field newNext.next. This way we can create a new node and assign its value as the previous field to newNext. Below I will give an example of what I mean, but try to implement the condition yourself (it is similar to the condition in the add(value) method)
```
        ....
        checkRangeForIndex(index);
        Node<T> newNext = getElement(index);
        Node<T> newPred = newNext.prev;
        Node<T> newNode = new Node<>(value, newPred, newNext);
        newNext.prev = newNode;
        if (newPred == null) {
            .....
        } else {
           .....
        }
        size++;
    }
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
...
+            add(element);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        outOfBoundsCheck(index);
+        Node<T> current = getNodeByIndex(index);
+        return (T) current.value;
```
SalnikovMarkiian, line 58:
```suggestion
        return getNodeByIndex(index).value;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
...
+        for (T element : list) {
+            add(element);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        outOfBoundsCheck(index);
+        Node<T> current = getNodeByIndex(index);
```
SalnikovMarkiian, line 57:
```suggestion
       
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
...
+        T removedValue = current.value;
+        current.value = value;
+        return removedValue;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        outOfBoundsCheck(index);
+        if (index == 0) {
```
SalnikovMarkiian, line 73: do we really need it if we just delink the element obtained by the method getNodeByIndex().value;? 
This will be more optimal, in my opinion, but the unlink method in this case, also needs to be improved

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
...
+
+    private void unlink(Node current) {
+        if (current == head) {
+            head = head.next;
+        } else if (current == tail) {
+            tail = tail.prev;
+        } else {
+            current.prev.next = current.next;
+            current.next.prev = current.prev;
+        }
```
SalnikovMarkiian, line 136: I recommend using the following implementation of the method:
```
 private Node<T> unlink(Node<T> current) {
        if (current.prev == null) {
            head = current.next;
        } else {
            current.prev.next = current.next;
        }
        if (current.next == null) {
            tail = current.prev;
        } else {
            current.next.prev = current.prev;
        }
        size--;
        return current;
    }
    
```
In this implementation, we return the node and also automatically reduce the size directly encapsulated in this method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
...
+        removedElement = (T) current.value;
+        unlink(current);
+        size--;
+        return removedElement;
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> current = head;
+        for (int i = 0; i < size; i++) {
```
SalnikovMarkiian, line 95:
```suggestion
        for (Node<T> current = head; current != null; current = current.next) {
```
it will be more readable and safer to use a loop that directly iterates through the list

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
...
+            if (current.value != null && current.value.equals(object) || current.value == object) {
```
SalnikovMarkiian, line 96:
```suggestion
            if (current.value != null && object.equals(current.value) || current.value == object) {
```
is easier to read :)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void outOfBoundsCheck(int index) {
```
SalnikovMarkiian, line 116: You have two identical methods, with different names) I think one of them can be deleted, and only the remaining one can be used in the code :)

SalnikovMarkiian (changes requested):
I have proposed many changes, and I think they will be optimal. If you do not agree with them - you can change my opinion)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> tail;
+    private Node<T> head;
+    private int size;
+
     @Override
     public void add(T value) {
+        Node<T> node = new Node<>(value);
```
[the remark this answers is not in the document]
liliia-ponomarenko: agree. You could  see this point in common mistakes  
[When creating a new Node, pass all the fields to Node constructor.](https://mate-academy.github.io/jv-program-common-mistakes/java-core/collections/linked-list.html#when-creating-a-new-node-pass-all-the-fields-to-node-constructor)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
...
+        if (head == null) {
+            head = tail = newNode;
+        } else if (index == 0) {
+            newNode.next = head;
+            head.prev = newNode;
+            head = newNode;
+        } else if (index == size) {
+            newNode.prev = tail;
+            tail.next = newNode;
+            tail = newNode;
```
liliia-ponomarenko, line 36: [Don’t complicate if-else construction. ](https://mate-academy.github.io/jv-program-common-mistakes/java-core/collections/linked-list.html#dont-complicate-if-else-construction-detailed-explanation)
Also, you can call add(T value) when index == size

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
...
     public void addAll(List<T> list) {
+        for (T element : list) {
+            add(element);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        outOfBoundsCheck(index);
```
liliia-ponomarenko, line 56: use verbs for variable names

src/main/java/core/basesyntax/MyLinkedList.java
liliia-ponomarenko: agree. Let’s fix it ;)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
...
+            T removedElement = (T) tail.value;
+            tail = tail.prev;
+            size--;
+            return removedElement;
+        }
+        T removedElement;
+        Node<T> current = getNodeByIndex(index);
+        removedElement = (T) current.value;
+        unlink(current);
+        size--;
```
liliia-ponomarenko, line 88: you can move  size--; to unlink()

src/main/java/core/basesyntax/MyLinkedList.java
liliia-ponomarenko: agree

liliia-ponomarenko (changes requested):
Good job! Let’s  improve  your solution ;)

liliia-ponomarenko (changes requested):
triggered reject

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,152 @@
...
+        if (index < 0 || index > size) {
+            throw new ArrayIndexOutOfBoundsException("Incorrect index: " + index);
+        }
+        Node<T> newNode = new Node<>(value);
+        if (index == size) {
+            add(value);
+        } else if (index == 0) {
+            addFirst(newNode);
+        } else {
+            addTo(newNode,index);
```
Elena-Bruyako, line 34: Please rename the method

Elena-Bruyako (changes requested):
Some comment wasn't fixed

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,152 @@
...
+        }
+        return current;
+    }
+
+    private static class Node<T> {
+        private T value;
+        private Node<T> next;
+        private Node<T> prev;
+
+        private Node(T value) {
```
Elena-Bruyako, line 150: See comment above about the constructor

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,152 @@
...
+            head = head.next;
+            return (T) current.value;
+        } else if (current == tail) {
+            tail = tail.prev;
+            return (T) current.value;
+        } else {
+            current.prev.next = current.next;
+            current.next.prev = current.prev;
+        }
+        return (T) current.value;
```
Elena-Bruyako, line 126: See the comment above. You don't really need to implement return statement in every if-else blocks. Also add size--;

Elena-Bruyako (changes requested):
Pay attention to the comments about Node constructor and unlink method

src/main/java/core/basesyntax/MyLinkedList.java
KDanil2000: Did I do everything right?

Elena-Bruyako (approved):
Good job!

PR #1174, opened 2023-08-28
Added solution for linkedlist

AlexRosenkreuz:
Created class Node and implemented suggested LinkedList methods

### DIFF
--- a/src/main/java/core/basesyntax/MyLinkedList.java
+++ b/src/main/java/core/basesyntax/MyLinkedList.java
@@ -3,45 +3,140 @@ package core.basesyntax;
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private static class Node<T> {
+        protected T item;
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,140 @@
...
+        node.item = value;
+        return oldValue;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        Node<T> node = getNode(index);
+        T removedValue = node.item;
+        if (node.prev != null) {
```
Elena-Bruyako, line 83: To avoid duplicating of code (if - else statements) you can create method unlink(Node<T> node) for example and use it in remove(int index) and remove(T object) methods

Elena-Bruyako (changes requested):
Good job! Let's improve your solution!

liliia-ponomarenko (approved):
Good job! Minor comments ;)

PR #1181, opened 2023-09-05
implementing MyLinkedListInterface

Petro-Smoliar:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,181 @@
...
-        return null;
+        checkPositionIndex(index);
+        Node<T> node = nodeByIndex(index);
+        T removeNode = node.item;
+        delLink(node);
+        return removeNode;
     }
 
     @Override
     public boolean remove(T object) {
```
Ivan95kos, line 89: It is better to rewrite like this:
```
    @Override
        Node<T> nodeToRemove = head;
        while (nodeToRemove != null) {
            if (nodeToRemove.value == object
                    || nodeToRemove.value != null && nodeToRemove.value.equals(object)) {
                unlink(nodeToRemove);
                return true;
            }
            nodeToRemove = nodeToRemove.next;
        }
        return false;
    }
```

Ivan95kos (changes requested):
Good job overall!

kozhukhovsky (approved):

PR #1193, opened 2023-09-26
Added solution

slaybrute:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,143 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node tail;
+    private Node head;
```
Ivan95kos, line 7: Node should be parameterized
```suggestion
    private Node<T> tail;
    private Node<T> head;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,143 @@
...
+            head.next = new Node(head, value, null);
+            head = head.next;
+        }
     }
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Invalid index: " + index + " for size: " + size);
+        }
```
Ivan95kos, line 25: move it to the private method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,143 @@
...
+            return node;
+        }
+        Node node = head;
+        for (int i = size - 1; i > index; i--) {
+            node = node.prev;
+        }
+        return node;
+    }
+
+    private class Node {
```
Ivan95kos, line 133:
```suggestion
    private class Node<T> {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,143 @@
...
+        private T element;
+        private Node prev;
+        private Node next;
+
+        private Node(Node prev, T element, Node next) {
+            this.prev = prev;
+            this.element = element;
+            this.next = next;
+            size++;
```
Ivan95kos, line 142: You should do it in the methods
```suggestion
```

Ivan95kos (changes requested):
Good job overall!

src/main/java/core/basesyntax/MyLinkedList.java
slaybrute: It has no sence because I use it once

src/main/java/core/basesyntax/MyLinkedList.java
slaybrute: > You should do it in the methods

I thought I would be more logically correct and more convenient. It would reduce code lines in add method

src/main/java/core/basesyntax/MyLinkedList.java
slaybrute: > move it to the private method

It has no sense because we use this once

src/main/java/core/basesyntax/MyLinkedList.java
slaybrute: > 

For what we need parametrize class if we can just make it non-static and use Type from LinkedList It will be the same ? I think it is useless action

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,144 @@
...
+            add(value);
+            return;
+        } else if (index == 0) {
+            tail = new Node<>(null, value, tail);
+        } else {
+            Node<T> nextNode = getNode(index);
+            nextNode.prev.next = new Node<>(nextNode.prev, value, nextNode);
+            nextNode.prev = nextNode.prev.next;
+        }
+        size++;
```
olekskov, line 37: then reorder code blocks to simplify code and reuse existing method
```suggestion
        if (index == size) {
            add(value);
            return;
        }
        checkIndexValidation(index);
        if (index == 0) {
            tail = new Node<>(null, value, tail);
        } else {
            Node<T> nextNode = getNode(index);
            nextNode.prev.next = new Node<>(nextNode.prev, value, nextNode);
            nextNode.prev = nextNode.prev.next;
        }
        size++;
```

olekskov (changes requested):

olekskov (approved):

PR #1195, opened 2023-09-28
Realized LinkedList class with Node class and basic functions. Also c…

Trayvor:
…reated 2 private functions: getIndexedElement() that helps to find element by index and unlink() that actually deletes item.

### DIFF
--- a/src/main/java/core/basesyntax/MyLinkedList.java
+++ b/src/main/java/core/basesyntax/MyLinkedList.java
@@ -5,7 +5,7 @@ import java.util.List;
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
     private Node<T> head;
     private Node<T> tail;
-    private int size = 0;
+    private int size;
 
     @Override
     public void add(T value) {
@@ -39,8 +39,8 @@ public class MyLinkedList<T> implements MyLinkedListInterface<T> {
 
     @Override
     public void addAll(List<T> list) {
-        for (T t : list) {
-            add(t);
+        for (T note : list) {
+            add(note);
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,143 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size = 0;
```
okuzan, line 8:
```suggestion
    private int size;
```
0 is default for int

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,143 @@
...
+        Node(Node<T> prev, T element, Node<T> next) {
+            this.item = element;
+            this.next = next;
+            this.prev = prev;
+        }
+
+        @Override
+        public boolean equals(Object obj) {
+            return super.equals(obj);
+        }
```
okuzan, line 143: we don't need equals here, and what's the point to override it if you leave the same implementation

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,143 @@
...
+        return deletedNode.item;
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> deletedNode;
+        for (int i = 0; i < size; i++) {
+            deletedNode = getIndexedElement(i);
+            if ((deletedNode.item != null && deletedNode.item.equals(object))
+                    || (deletedNode.item == null && object == null)) {
```
okuzan, line 72: extra parentheses

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,143 @@
...
+            return;
+        }
+        newElement.prev.next = newElement;
+        indexedElement.prev = newElement;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
+        for (T t : list) {
```
okuzan, line 42:
```suggestion
        for (T note : list) {
```

okuzan (changes requested):
Good job overall!

src/main/java/core/basesyntax/MyLinkedList.java
Trayvor: oops i forgot to delete that

olekskov (approved):
Good job

PR #1196, opened 2023-09-30
Implemented LinkedList

dima-semeniuk:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,149 @@
...
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    private static class Node<T> {
+        private T value;
+        private Node<T> next;
+        private Node<T> previous;
+
+        public Node(T value, Node<T> next, Node<T> previous) {
```
flekros, line 16: Better change sequence of inputs.What you think about previous,value,next?

flekros (approved):
Great job!

src/main/java/core/basesyntax/MyLinkedList.java
dima-semeniuk: It's a good idea! Thanks

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,149 @@
...
-        return null;
+        checkIndex(index);
+        return unlink(index).value;
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> currentNode = head;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(currentNode.value, object)) {
```
flekros, line 91: Don't use Objects.equals!

flekros (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,147 @@
...
+            newNode.previous = currentNode.previous;
+            currentNode.previous.next = newNode;
+            currentNode.previous = newNode;
+            size++;
+        }
     }
 
     @Override
     public void addAll(List<T> list) {
+        if (list != null) {
```
kshuryhin, line 58: Probably it would be useful to throw an exception if `list` is `null` to notify user

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,147 @@
...
-        return null;
+        checkIndex(index);
+        return getNodeByIndex(index).value;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        checkIndex(index);
+        T valueToReturn = getNodeByIndex(index).value;
```
kshuryhin, line 75: Here, you're calling `getNodeByIndex(index) `twice. This is not efficient; consider storing the node in a variable to avoid the double lookup

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,147 @@
...
+            for (int i = size - 1; i > index; i--) {
+                curentNode = curentNode.previous;
+            }
+        }
+        return curentNode;
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Index is invalid!");
```
kshuryhin, line 128: It would be helpful to include the invalid index in the error message.

kshuryhin (changes requested):
Good job, some minor points to consider

kshuryhin (approved):
Great job!

PR #1206, opened 2023-10-09
Created own LinkedList class

HovorukhaBohdan:


Alexander1iubchenko (approved):
everything seems to be ok

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,172 @@
...
+            throw new IndexOutOfBoundsException();
+        }
+    }
+
+    private static class Node<T> {
+        private Node<T> next;
+        private T value;
+        private Node<T> prev;
+
+        public Node(Node<T> next, T value, Node<T> prev) {
```
kshuryhin, line 168: Do we really need `public` here, what do you think?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,172 @@
...
+        if (next != null && prev != null) {
+            next.prev = prev;
+            prev.next = next;
+        }
+
+        size--;
+        return current.value;
+    }
+    
+    private Node<T> findNodeByIndex(int index) {
```
kshuryhin, line 141: The current implementation iterates from the `head` node to the desired `index`. If the desired `index` is closer to the end of the list, it would be more efficient to start from the `tail` and move backward.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,172 @@
...
+        for (int i = 0; i < index; i++) {
+            current = current.next;
+        }
+
+        return current;
+    }
+
+    private void checkIndexPositionException(int index) {
+        if (!(index >= 0 && index <= size)) {
+            throw new IndexOutOfBoundsException();
```
kshuryhin, line 153: Consider providing some descriptive error messages

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,172 @@
...
+        }
+
+        if (prev == null) {
+            head = next;
+        }
+        
+        if (next != null && prev != null) {
+            next.prev = prev;
+            prev.next = next;
+        }
```
kshuryhin, line 135: Probably you don't need to check if both `next` and `prev` are not null since the node's neighbours can be linked directly.

kshuryhin (changes requested):
Great job, but there are some points to consider

kshuryhin (approved):
Great job!

PR #1211, opened 2023-10-13
Add implementation for the LinkedList task

MinoDentori:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,238 @@
...
 
 import java.util.List;
+import java.util.Objects;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private static final String INDEX_OUT_OF_BOUNDS_MESSAGE =
+            "Index: is out of bound for length ";
+    private Node<T> head;
+    private Node<T> tail;
+    private int size = 0;
```
YevheniiVlasenko, line 11: int type default value is 0, no need to initialize it here

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,238 @@
...
+        nodeToUnlink.prev = null;
+        tail = prev;
+        if (prev == null) {
+            head = null;
+        } else {
+            prev.next = null;
+        }
+        size--;
+        return removeItem;
+    }
```
YevheniiVlasenko, line 181: It might be done in way less code.

        if (node.equals(head)) {
            head = head.next;
        } else if (node.equals(tail)) {
            tail = tail.prev;
        } else {
            node.prev.next = node.next;
            node.next.prev = node.prev;
        }
    }
    It won't work in your code, but for example

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,238 @@
...
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException(INDEX_OUT_OF_BOUNDS_MESSAGE + index);
+        }
+    }
+
+    private void checkIndexForAdd(int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException(INDEX_OUT_OF_BOUNDS_MESSAGE + index);
+        }
+    }
```
YevheniiVlasenko, line 193: You use these methods only once and they are short. Not sure you need them

YevheniiVlasenko (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
MinoDentori: agree fixed

src/main/java/core/basesyntax/MyLinkedList.java
MinoDentori: Disagree, I guess in this way I make code more readable.
Also, I had such kind of implementation for ArrayList.
Left for mentor decision mb.

src/main/java/core/basesyntax/MyLinkedList.java
MinoDentori: Simplified, but l will miss those methods.

d1sam (approved):
well done!

PR #1214, opened 2023-10-13
MyLinkedList custom realization

fedorovychh:
realized overrided interface methods
added inner class Node
added getNodeByIndex method
added check on exception

### DIFF
--- a/src/main/java/core/basesyntax/MyLinkedList.java
+++ b/src/main/java/core/basesyntax/MyLinkedList.java
@@ -6,9 +6,7 @@ public class MyLinkedList<T> implements MyLinkedListInterface<T> {
     private static final String INDEX_OUT_OF_BOUNDS_MESSAGE = "Index %d "
             + "does not exist in this LinkedList!";
     private static final int DEFAULT_DIVISOR = 2;
-    private static final int INDEX_STOPPING_FROM_TAIL = 1;
     private static final int FIRST_ELEMENT_INDEX = 0;
-    private static final int EMPTY_LIST_VALUE = 0;
     private Node<T> head;
     private Node<T> tail;
     private Node<T> current;
@@ -17,9 +15,8 @@ public class MyLinkedList<T> implements MyLinkedListInterface<T> {
     @Override
     public void add(T value) {
         current = new Node<>(null, value, null);
-        if (tail == null) {
-            tail = current;
-            head = tail;
+        if (head == null) {
+            head = tail = current;
         } else {
             current.prev = tail;
             current.prev.next = current;
@@ -55,44 +52,29 @@ public class MyLinkedList<T> implements MyLinkedListInterface<T> {
 
     @Override
     public T get(int index) {
-        return getNodeByIndex(index).item;
+        return getNodeByIndex(index).value;
     }
 
     @Override
     public T set(T value, int index) {
         current = getNodeByIndex(index);
-        T oldObject = current.item;
-        current.item = value;
+        T oldObject = current.value;
+        current.value = value;
         return oldObject;
     }
 
     @Override
     public T remove(int index) {
-        current = getNodeByIndex(index);
-        if (current.next == null && current.prev == null) {
-            head = null;
-            tail = null;
-        } else if (current.next == null) {
-            current.prev.next = null;
-            tail = current.prev;
-        } else if (current.prev == null) {
-            current.next.prev = null;
-            head = current.next;
-        } else {
-            current.prev.next = current.next;
-            current.next.prev = current.prev;
-        }
-        size--;
-        return current.item;
+        return unlink(getNodeByIndex(index));
     }
 
     @Override
     public boolean remove(T object) {
         current = head;
         for (int i = 0; i < size; i++) {
-            if ((object == current.item)
-                    || (object != null && object.equals(current.item))) {
-                remove(i);
+            if ((object == current.value)
+                    || (object != null && object.equals(current.value))) {
+                unlink(current);
                 return true;
             }
             current = current.next;
@@ -107,16 +89,16 @@ public class MyLinkedList<T> implements MyLinkedListInterface<T> {
 
     @Override
     public boolean isEmpty() {
-        return size == EMPTY_LIST_VALUE;
+        return size == 0;
     }
 
     private static class Node<T> {
-        private T item;
+        private T value;
         private Node<T> next;
         private Node<T> prev;
 
-        public Node(Node<T> prev, T element, Node<T> next) {
-            this.item = element;
+        private Node(Node<T> prev, T element, Node<T> next) {
+            this.value = element;
             this.next = next;
             this.prev = prev;
         }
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,157 @@
...
+
+    private static class Node<T> {
+        private T item;
+        private MyLinkedList.Node<T> next;
+        private MyLinkedList.Node<T> prev;
+
+        Node(MyLinkedList.Node<T> prev, T element, MyLinkedList.Node<T> next) {
+            this.item = element;
+            this.next = next;
+            this.prev = prev;
```
ponomvrenko, line 121: I can suggest you to remove prev & next parameters in constructor. By default, you assigned prev & next in new Node  as null. So you don't need always write null in constructors)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,157 @@
...
+            throw new IndexOutOfBoundsException(String
+                    .format(INDEX_OUT_OF_BOUNDS_MESSAGE, index));
+        }
+    }
+
+    private void checkIndexExistence(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException(String
+                    .format(INDEX_OUT_OF_BOUNDS_MESSAGE, index));
+        }
```
ponomvrenko, line 157: You can connect these two methods in one. Think what you should give as parameters when you invoke this method.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,157 @@
...
+    private static final int FIRST_ELEMENT_INDEX = 0;
+    private static final int EMPTY_LIST_VALUE = 0;
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
     @Override
     public void add(T value) {
+        Node<T> current = new Node<>(null, value, null);
+        if (tail == null) {
```
ponomvrenko, line 19: If we put null as a value in Node, will it work properly?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,157 @@
...
+            current.prev.next = current;
+            tail = current;
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        checkAddIndexExistence(index);
+        Node<T> current;
```
ponomvrenko, line 33: You use this variable only in last if-else case. It's better to relocate this invocation. You can optimize your code. You shouldn't always create new current node when we use method in other cases.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,157 @@
...
     public T remove(int index) {
-        return null;
+        Node<T> current = getNodeByIndex(index);
+        if (current.next == null && current.prev == null) {
+            head = null;
+            tail = null;
+        } else if (current.next == null) {
+            current.prev.next = null;
+            tail = current.prev;
+        } else if (current.prev == null) {
```
ponomvrenko, line 78: The same situation as on the 19th line.

ponomvrenko (changes requested):

ponomvrenko:
Well done! You just have to think about some cases that I wrote before👀

src/main/java/core/basesyntax/MyLinkedList.java
proxychang: Exactly yes, it works properly

src/main/java/core/basesyntax/MyLinkedList.java
proxychang: I disagree with this point, because this realization of inner class Node is made like as standart for LinkedList. The Node constructor takes 3 parameters under the hood. Not specifying them violates the concept of Node

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,157 @@
...
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private static final String INDEX_OUT_OF_BOUNDS_MESSAGE = "Index %d "
+            + "does not exist in this LinkedList!";
+    private static final int DEFAULT_DIVISOR = 2;
+    private static final int INDEX_STOPPING_FROM_TAIL = 1;
+    private static final int FIRST_ELEMENT_INDEX = 0;
+    private static final int EMPTY_LIST_VALUE = 0;
+    private Node<T> head;
+    private Node<T> tail;
+    private Node<T> current;
```
bhdnchui, line 14: I think it can be local variable.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,157 @@
...
+    private int size;
+
     @Override
     public void add(T value) {
+        current = new Node<>(null, value, null);
+        if (tail == null) {
```
bhdnchui, line 20: Make check with head.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,157 @@
...
-        return false;
+        return size == EMPTY_LIST_VALUE;
+    }
+
+    private static class Node<T> {
+        private T item;
+        private Node<T> next;
+        private Node<T> prev;
+
+        public Node(Node<T> prev, T element, Node<T> next) {
```
bhdnchui, line 118: I think we can make it more closed.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,157 @@
...
+        return current.item;
     }
 
     @Override
     public boolean remove(T object) {
+        current = head;
+        for (int i = 0; i < size; i++) {
+            if ((object == current.item)
+                    || (object != null && object.equals(current.item))) {
+                remove(i);
```
bhdnchui, line 95: You find an element and then you call a method that will search for it by index.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,157 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == EMPTY_LIST_VALUE;
+    }
+
+    private static class Node<T> {
+        private T item;
```
bhdnchui, line 114:
```suggestion
        private T value;
```
It would be better, i think.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,157 @@
...
+            while (index != EMPTY_LIST_VALUE) {
+                node = node.next;
+                index--;
+            }
+        } else {
+            node = tail;
+            int stepCounter = size - index;
+            while (stepCounter != INDEX_STOPPING_FROM_TAIL) {
+                node = node.prev;
+                stepCounter--;
```
bhdnchui, line 140: I think we can make it simplier.
Maybe we can avoid working with two different variables. stepCounter and index.

bhdnchui (changes requested):
Good job! But a few comments need to be processed.

src/main/java/core/basesyntax/MyLinkedList.java
proxychang: We use head, tail and current in few methods, is it necessary to make them local? Or you just noticed about empty list value?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,155 @@
...
+            this.next = next;
+            this.prev = prev;
+        }
+    }
+
+    private Node<T> getNodeByIndex(int index) {
+        checkIndexExistence(index);
+        boolean isIndexLessHalf = index <= size / DEFAULT_DIVISOR;
+        Node<T> node;
+        if (isIndexLessHalf) {
```
bhdnchui, line 111:
```suggestion
        Node<T> node;
        if (index <= size / DEFAULT_DIVISOR) {
```

bhdnchui (changes requested):
Good job!

bhdnchui (approved):
Good job!

PR #1220, opened 2023-10-14
added solution jv-linked-list

TarasYurkovskyi:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
+            currentNode.previous.next = newNode;
+            currentNode.previous = newNode;
+            size++;
+        }
     }
 
     @Override
     public void addAll(List<T> list) {
+        if (list == null) {
+            throw new RuntimeException("The list is null!");
```
proxychang, line 56: I think you can make such messages as constants

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private Node<T> getNodeByIndex(int index) {
+        Node<T> curentNode;
+        if (index < size / 2) {
```
proxychang, line 110: I think you can make all magical numbers constant

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
     public T get(int index) {
-        return null;
+        checkIndex(index);
+        return getNodeByIndex(index).value;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        checkIndex(index);
```
proxychang, line 71: I think you can make index check in getNodeByIndex , it will optimize your code and you won`t call checkIndex many times, just once when you call getNodeByIndex

proxychang (approved):
You`re smart! Noticed a couple points that will make your code better

src/main/java/core/basesyntax/MyLinkedList.java
TarasYurkovskyi: > I think you can make index check in getNodeByIndex , it will optimize your code and you won`t call checkIndex many times, just once when you call getNodeByIndex
i think it is not a good point bec in my add method there are checkIndex method and getNodeByIndex. So in this case i will check index 2 times where the second is unnecessary. So it will not work properly(  And create same methods but one without check is stupid)

TarasYurkovskyi: and thx for comments)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
+        Node<T> nodeOldValue = getNodeByIndex(index);
```
bhdnchui, line 72: Find better name.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
+        return unlink(index).value;
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> currentNode = head;
+        for (int i = 0; i < size; i++) {
+            if (object == currentNode.value
+                    || (object != null && object.equals(currentNode.value))) {
+                remove(i);
```
bhdnchui, line 90: You find a node here and then you are searching for a node by index again.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
+            head = tail = null;
+        } else if (unlinkNode.previous == null) {
+            head = head.next;
+            head.previous = null;
+        } else if (unlinkNode.next == null) {
+            tail = tail.previous;
+            tail.next = null;
+        } else {
+            unlinkNode.previous.next = unlinkNode.next;
+            unlinkNode.next.previous = unlinkNode.previous;
```
bhdnchui, line 148: Remove pointers in the node too.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
+            add(value);
+        } else if (index == 0) {
+            addFirst(newNode);
+        } else {
+            Node<T> currentNode = getNodeByIndex(index);
+            newNode.next = currentNode;
+            newNode.previous = currentNode.previous;
+            currentNode.previous.next = newNode;
+            currentNode.previous = newNode;
+            size++;
```
bhdnchui, line 49: Why did you separate `addFirst` method but a bigger part of the code is here?

bhdnchui (changes requested):
Good job.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
...
+        if (isEmpty()) {
+            head = new Node<>(null, value, null);
+            tail = head;
+        } else {
+            Node<T> newNode = new Node<>(tail, value, null);
+            tail.next = newNode;
+            tail = newNode;
+        }
+        size++;
     }
```
bhdnchui, line 21:
```suggestion
        Node<T> newNode = new Node<>(null, value, null);
        if (isEmpty()) {
            head = newNode;
        } else {
            newNode.prev = tail;
            tail.next = newNode;
        }
        tail = newNode;
        size++;
    }
```

PR #1225, opened 2023-10-15
solution of jv-linked-list task

andriimikhaylovskiy:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,157 @@
...
+                    return true;
+                }
+            }
+        } else {
+            for (Node<T> i = tail; i != null; i = i.next) {
+                if (i.item != null && i.item.equals(object)) {
+                    unlink(i);
+                    return true;
+                }
+            }
```
okuzan, line 75: you have the same body - remove external if-else, combine two fors and adjust condition in the if - if object == i.item or i.item != null && i.item.equals(object)

okuzan (changes requested):
Good job on the first try!

olekskov (approved):

PR #1235, opened 2023-10-20
Create my own LinkedList

Paradoxallist:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,174 @@
...
+        checkElementIndex(index);
+        Node<T> removedNode = node(index);
+        unlink(removedNode);
+        return removedNode.item;
     }
 
     @Override
     public boolean remove(T object) {
+        if (object == null) {
+            for (Node<T> x = first; x != null; x = x.next) {
```
kerrrusha, line 70: use for-each loop

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,174 @@
...
+                if (x.item == null) {
+                    unlink(x);
+                    return true;
+                }
+            }
+        } else {
+            for (Node<T> x = first; x != null; x = x.next) {
```
kerrrusha, line 77: use for-each loop

kerrrusha (changes requested):
need minor change

src/main/java/core/basesyntax/MyLinkedList.java
Paradoxallist: To use the for-each construct, I will need to add the list to my LinkedList. (ArrayList for example) This will make the LinkedList class redundant because it already uses the functionality of ArrayList, which itself implements a list.

src/main/java/core/basesyntax/MyLinkedList.java
Paradoxallist: To use the for-each construct, I will need to add the list to my LinkedList. (ArrayList for example) This will make the LinkedList class redundant because it already uses the functionality of ArrayList, which itself implements a list.

Paradoxallist: Ok, how do I do this? I don't have a separate list for for-each.

PR #1241, opened 2023-10-24
Realization of the LinkedList task

druhov:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,140 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size;
+    private Node<T> head;
+    private Node<T> tail;
+    private Node<T> temporaryNode;
+
+    private class Node<T> {
```
Ivan95kos, line 11: All inner classes should be at the end of class

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,140 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size;
+    private Node<T> head;
+    private Node<T> tail;
+    private Node<T> temporaryNode;
```
Ivan95kos, line 9: You should have only the `head` and `tail` nodes of the class
It is better to create a new Node in the methods
```suggestion
```

Ivan95kos (changes requested):
Good job overall!

src/main/java/core/basesyntax/MyLinkedList.java
druhov: Why is this solution better than the one I wrote?

kozhukhovsky (approved):

src/main/java/core/basesyntax/MyLinkedList.java
kozhukhovsky: not fixed

PR #1246, opened 2023-10-28
Impletented LinkedList

korets270:
Changed in manev java version from original 17th to java 11 in order to build project

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,7 @@
     <version>1.0-SNAPSHOT</version>
 
     <properties>
-        <jdk.version>17</jdk.version>
+        <jdk.version>11</jdk.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
         <maven.checkstyle.plugin.configLocation>
--- a/src/main/java/core/basesyntax/MyLinkedList.java
+++ b/src/main/java/core/basesyntax/MyLinkedList.java
@@ -3,45 +3,162 @@ package core.basesyntax;
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    private static class Node<T> {
+        private Node<T> previous;
+        private T element;
+        private Node<T> next;
+
+        public Node(Node<T> previous, T element, Node<T> next) {
+            this.previous = previous;
+            this.element = element;
+            this.next = next;
+        }
+    }
+
     @Override
     public void add(T value) {
+        linkLast(value);
     }
 
     @Override
     public void add(T value, int index) {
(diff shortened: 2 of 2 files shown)


pom.xml
```diff
@@ -9,7 +9,7 @@
     <version>1.0-SNAPSHOT</version>
 
     <properties>
-        <jdk.version>17</jdk.version>
+        <jdk.version>11</jdk.version>
```
olekskov, line 12: Downgrade of Java version is acceptable in this task but I strongly recommend to uplift Java to version 17 locally. In some of next tasks you will have much more problems during downgrade

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,162 @@
...
+        public Node(Node<T> previous, T element, Node<T> next) {
+            this.previous = previous;
+            this.element = element;
+            this.next = next;
+        }
+    }
+
     @Override
     public void add(T value) {
+        linkLast(value);
```
olekskov, line 25: what is reason to extract whole code to separate method? just put it here

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,162 @@
...
     }
 
     @Override
     public void add(T value, int index) {
+        if (index == size) {
+            linkLast(value);
```
olekskov, line 31: and here will be 
```suggestion
            add(value);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,162 @@
...
+        size++;
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Index cannot be equal or bigger than size!");
+        }
+    }
+
+    private Node<T> node(int index) {
```
olekskov, line 138: method names are more readable in place of usage when they have verbs in naming
```suggestion
    private Node<T> getNodeByIndex(int index) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,162 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    private static class Node<T> {
```
olekskov, line 11: Please put private inner class after public methods

olekskov (changes requested):
Good job overall, only some codestyle remarks

src/main/java/core/basesyntax/MyLinkedList.java
korets270: Do you know any source: youtube or book where I can have all these syntax tips all in 1 place?

Ivan95kos (approved):
Nice

PR #1247, opened 2023-10-28
home-work (LinkedList)

ddryzhov:
implemented all the required methods MyLinkedListInterface

### DIFF
--- a/src/main/java/core/basesyntax/MyLinkedList.java
+++ b/src/main/java/core/basesyntax/MyLinkedList.java
@@ -3,45 +3,139 @@ package core.basesyntax;
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private static final String iOutOfBounds = "Index is out of bounds";
+    private Node<T> head;
+    private int size;
+
+    public MyLinkedList() {
+    }
+
     @Override
     public void add(T value) {
+        add(value, size);
     }
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException(iOutOfBounds);
+        }
+
+        if (index == 0) {
+            Node<T> newNode = new Node<>(null, value, head);
+            if (head != null) {
+                head.prev = newNode;
+            }
+            head = newNode;
+        } else {
+            Node<T> previous = getNode(index - 1);
+            Node<T> newNode = new Node<>(previous, value, previous.next);
+            previous.next = newNode;
+            if (newNode.next != null) {
+                newNode.next.prev = newNode;
+            }
+        }
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
+        for (T value : list) {
+            add(value);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        Node<T> node = getNode(index);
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,139 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private static final String iOutOfBounds = "Index is out of bounds";
+    private Node<T> head;
+    private int size;
```
olekskov, line 8: Solution without tail is interesting but not efficient and violates requirement
https://github.com/mate-academy/jv-linked-list/blob/master/checklist.md#dont-iterate-your-nodes-from-the-head-to-the-tail-or-vice-versa
Please introduce tail and use it to improve performance of logic (add to end of list operation is faster when working with tail and also getting node by index will be faster if index is in 2nd part of list and we iterating from tail)

olekskov (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,156 @@
...
+        while (current != null) {
+            if (object == null) {
+                if (current.value == null) {
+                    updateReferences(current);
+                    return true;
+                }
+            } else if (object.equals(current.value)) {
+                updateReferences(current);
+                return true;
+            }
```
olekskov, line 117:
```suggestion
            if (object == current.value || object != null && object.equals(current.value)) {
                updateReferences(current);
                return true;
            }
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,156 @@
...
+            current = current.next;
+        }
         return false;
     }
 
+    private void updateReferences(Node<T> node) {
```
olekskov, line 123: naming not clearly explains how we are updating references
```suggestion
    private void unlink(Node<T> node) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,156 @@
...
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException(iOutOfBounds);
+        }
+
+        Node<T> current = head;
+        for (int i = 0; i < index; i++) {
+            current = current.next;
+        }
+        return current;
     }
```
olekskov, line 84: Private methods should be placed after public ones

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,156 @@
...
+        if (index < size / 2) {
+            current = head;
+            for (int i = 0; i < index; i++) {
+                current = current.next;
+            }
+        } else {
+            current = tail;
+            for (int i = size - 1; i > index; i--) {
+                current = current.prev;
+            }
```
olekskov, line 69: this logic should be contained in getNode() method and this method should be reused here instead of full search

olekskov (changes requested):

Ivan95kos (approved):
Nice

PR #1250, opened 2023-11-01
Create simple version of LinkedList

AShtuka:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,185 @@
...
+
+    private Node<T> getNodeByIndex(int index) {
+        checkIndex(index);
+        int listMiddleIndex = size / 2;
+        if (index <= listMiddleIndex) {
+            return searchFromStart(index);
+        } else {
+            return searchFromEnd(index);
+        }
+    }
```
kerrrusha, line 153: we should place methods in order of their calling to make reading more simple (from up to down)
also, put these methods right after last `remove()` method, as this is the last place where `getNodeByIndex` being called
```suggestion
    private Node<T> getNodeByIndex(int index) {
        checkIndex(index);
        int listMiddleIndex = size / 2;
        if (index <= listMiddleIndex) {
            return searchFromStart(index);
        } else {
            return searchFromEnd(index);
        }
    }
    
    private Node<T> searchFromStart(int index) {
        Node<T> item = head;
        int counter = 0;
        do {
            if (counter == index) {
                return item;
            }
            item = item.next;
            counter++;
        } while (counter <= index);
        return item;
    }

    private Node<T> searchFromEnd(int index) {
        Node<T> item = tail;
        int counter = size - 1;
        do {
            if (counter == index) {
                return item;
            }
            item = item.prev;
            counter--;
        } while (counter >= index);
        return item;
    }
```

kerrrusha (approved):
good job

PR #1251, opened 2023-11-02
Implemented MyLinkedList class with all required methods

Vitalii-Kazymyrovych:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,253 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private static final int HEAD_INDEX = 0;
```
Rommelua, line 6: We can use 0 without constant in this class

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,253 @@
...
+        public Node(T item) {
+            this.item = item;
+        }
+
+        public Node() {
+
+        }
+
+        @Override
+        public boolean equals(Object node) {
```
Rommelua, line 110: delete this method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,253 @@
...
+            return false;
+        }
+    }
+
+    private boolean isTail(Node<T> node) {
+        return tail.equals(node)
+                && size > 1;
+    }
+
+    private boolean isHead(Node<T> node) {
```
Rommelua, line 138: better delete these methods

Rommelua (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,253 @@
...
+        return targetNode.item;
     }
 
     @Override
     public boolean remove(T object) {
-        return false;
+        Node<T> targetNode = findNode(object);
+        if (isTail(targetNode)) {
+            return false;
+        }
```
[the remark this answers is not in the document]
Vitalii-Kazymyrovych: Because we should return false, if the list doesn't contain an object. So, in findNode(object) we search through the list for required node. If we reached tail, and tail != required nod, we return tail. And if we return tail to remove(object) method, method returns false.

src/main/java/core/basesyntax/MyLinkedList.java
Vitalii-Kazymyrovych: Then I need to repeat code in isTail and isHead methods. Is this practice better, than overwritten equals?

src/main/java/core/basesyntax/MyLinkedList.java
Vitalii-Kazymyrovych: Why?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,238 @@
...
+        removeNode(targetNode);
+        return true;
```
Rommelua, line 77:
```suggestion
        for (Node<T> node = first; node != null; node = node.next) {
            if (compareItems(node.value, object)) {
                removeNode(node);
                return true;
            }
        }
        return false;
```

src/main/java/core/basesyntax/MyLinkedList.java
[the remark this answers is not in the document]
Rommelua: I can not understand your explanation. Better write in MA chat)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,238 @@
...
+                        || (this.next != null
+                        && this.next.equals(current.next)))));
+            }
+            return false;
+        }
+    }
+
+    private boolean isTail(Node<T> node) {
+        return tail.equals(node)
+                && size > 1;
```
Rommelua, line 127:
```suggestion
        return node.next == null;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,238 @@
...
+    }
+
+    private boolean isHead(Node<T> node) {
+        return head.equals(node)
+                && size > 1;
```
Rommelua, line 132:
```suggestion
        return node.prev == null;
```

src/main/java/core/basesyntax/MyLinkedList.java
Rommelua: Ok, but do it without equals

Rommelua (changes requested):

sokimaaa (approved):

PR #1255, opened 2023-11-03
Implemented methods MyLinkedListInterface to MyLinkedList

Sanmartial:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,170 @@
...
     @Override
     public void add(T value) {
+        add(value, size);
     }
 
     @Override
     public void add(T value, int index) {
+        if (index > size || index < 0) {
+            throw new IndexOutOfBoundsException("Index " + index + " out of bounds " + size);
+        }
```
kerrrusha, line 19: there are `checkElementIndex(int index)` method already

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,170 @@
...
+        } else {
+            prevNode.next = newNode;
+        }
+        size++;
+    }
+
+    private Node<T> iterNode(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Index is out of bounds");
+        }
```
kerrrusha, line 106: there are `checkElementIndex(int index)` method already, use it instead

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,170 @@
...
+        } else {
+            Node<T> fromFirstNode = first;
+            while (fromFirstNode != null) {
+                if (object.equals(fromFirstNode.item)) {
+                    deleteLink(fromFirstNode);
+                    return true;
+                }
+                fromFirstNode = fromFirstNode.next;
+            }
+        }
```
kerrrusha, line 77: can be more simple
```suggestion
         Node<T> fromFirstNode = first;
            while (fromFirstNode != null) {
                if (fromFirstNode.item == object) {
                    deleteLink(fromFirstNode);
                    return true;
                }
                fromFirstNode = fromFirstNode.next;
            }
```

kerrrusha (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
Sanmartial: The logic here is a little different

kerrrusha (approved):

PR #1270, opened 2023-11-14
MyLinkedList implemented

ivanlavrinenko:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,143 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> first;
+    private Node<T> last;
+    private int size;
+
+    private static class Node<T> {
```
kerrrusha, line 10: `private static` classes should be declared at the end of the file

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,143 @@
...
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Index: " + index + ", Size: " + size);
+        }
+    }
+
+    private void checkForAdd(int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Index: " + index + ", Size: " + size);
+        }
+    }
```
kerrrusha, line 32: methods should be declared in the order of their calling:
a() calls b() -> a() would be declared first, b() would be right after a()

kerrrusha (approved):

PR #1276, opened 2023-11-28
Solution

SSosedova:


okuzan (changes requested):
Well done on the first try! 👏  Minor changes suggested

okuzan (approved):
🎉

PR #1282, opened 2023-12-06
dsumaroka/jv-linked-list

dsumaroka:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,180 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    MyLinkedList.Node<T> node(int index) {
```
olekskov, line 105: method names are more readable when contain verb and describe what they are doing
```suggestion
    MyLinkedList.Node<T> findNodeByIndex(int index) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,180 @@
...
+                }
+            }
+        } else {
+            for (MyLinkedList.Node<T> x = first; x != null; x = x.next) {
+                if (object.equals(x.item)) {
+                    unlink(x);
+                    return true;
+                }
+            }
+        }
```
olekskov, line 91: There is no need to split with if/else
```suggestion
            for (MyLinkedList.Node<T> x = first; x != null; x = x.next) {
                if (object == x.item || object != null && object.equals(x.item)) {
                    unlink(x);
                    return true;
                }
        }
```

olekskov (changes requested):
good job overall, some style changes requested

PR #1290, opened 2023-12-19
Implemented MyLinkedList class.

davydov-software:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,167 @@
...
+        }
+
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index == size) {
+            add(value);
+        } else {
```
okuzan, line 27: not a mistake, but I'd rather add return after add and unwrap the remaining part of the code

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,167 @@
...
+
+        return node.value;
+    }
+
+    private static class Node<T> {
+        private T value;
+        private Node<T> prev;
+        private Node<T> next;
+
+        public Node(Node<T> prev, T value, Node<T> next) {
```
okuzan, line 163: constructor won't be accessible from outside of this file anyway
```suggestion
        Node(Node<T> prev, T value, Node<T> next) {
```
```suggestion
        private Node(Node<T> prev, T value, Node<T> next) {
```

okuzan (approved):
Good job!

PR #1292, opened 2023-12-20
implemented methods for work with LinkedList

petro2061:
In this task realized all methods from interface MyLinkedListInterface. Cteated spesial methods for chech input data (chack index). Added method for unlink remove node and method for finde note by index.


okuzan (approved):
Good job

PR #1307, opened 2024-01-11
solved task

Leanchers:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,177 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        if (size != 0) {
+            return false;
+        }
+        return true;
```
okuzan, line 147: Consider simplifying the code. You can directly return the result of the condition instead of using if/else or ternary operation to return true/false. For example:

    // Instead of this
    if (condition) {
        return true;
    } else {
        return false;
    }
    // You can do this
    return condition;

This makes the code cleaner and more straightforward.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,177 @@
...
+            if (size == 1) {
+                head = null;
+                size--;
+                return true;
+            }
+            head.next.prev = null;
+            head = head.next;
+            size--;
+            return true;
+        }
```
okuzan, line 122:
```suggestion
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,177 @@
...
+        Node<T> objectNode = head;
+        for (int i = 0; i < size; i++) {
+            if ((objectNode.value == object)
+                    || (objectNode.value != null && objectNode.value.equals(object))) {
+                objectNode.prev.next = objectNode.next;
+                objectNode.next.prev = objectNode.prev;
+                size--;
```
okuzan, line 129: create unlink method and call it here, as mentioned in the checklist

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,177 @@
...
+        findIndexNode(index).value = value;
+        return oldValue;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        T value;
+        checkIndexOutOfBoundsException(index);
+        if (index == 0) {
```
okuzan, line 85: get node by index
save value
call unlink method
return saved value

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,177 @@
...
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkIndexOutOfBoundsException(index);
+        return findIndexNode(index).value;
```
okuzan, line 69: you use them together, would make sense to check index in the retrieval by index (findIndexNode) method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,177 @@
...
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    private static class Node<T> {
+        private T value;
+        private Node<T> prev;
+        private Node<T> next;
+
+        public Node(Node<T> prev, T value, Node<T> next) {
```
okuzan, line 15:
```suggestion
        Node(Node<T> prev, T value, Node<T> next) {
```

okuzan (changes requested):
Good job, some methods need to be simplified

kozhukhovsky (approved):

PR #1312, opened 2024-01-22
LinkedList task

Ovikt0r:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,172 @@
...
+    private Node<T> getNodeByIndex(int index) {
+        if (index < (size >> SIZE_BIT_DIVIDER)) {
+            Node<T> iterateNodeFromFirst = first;
+            for (int i = 0; i < index; i++) {
+                iterateNodeFromFirst = iterateNodeFromFirst.next;
+            }
+            return iterateNodeFromFirst;
+        } else {
+            Node<T> iterateNodeFromLast = last;
+            for (int i = size - SIZE_SHIFTER; i > index; i--) {
```
okuzan, line 156: just use / 2, will be more readable than bitwise operations with constants

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,172 @@
...
+
+    private void checkIndexElement(int index) {
+        if (!isIndexElement(index)) {
+            throw new IndexOutOfBoundsException();
+        }
+    }
+
+    private void checkIndexPosition(int index) {
+        if (!isIndexPosition(index)) {
+            throw new IndexOutOfBoundsException();
```
okuzan, line 102: add index and size to the exception messages

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,172 @@
...
+                }
+            }
+        } else {
+            for (Node<T> current = first; current != null; current = current.next) {
+                if (object.equals(current.value)) {
+                    unlinkNode(current);
+                    return true;
+                }
+            }
+        }
```
okuzan, line 80: merge into a single for loop by combining conditions in `if` statements

okuzan (changes requested):
Very good job!

src/main/java/core/basesyntax/MyLinkedList.java
Ovikt0r: ok, I'll correct it. I always thougth that it is more prioritized to use bitwise operation. I.e. should I try to avoid it? Or should use it appropriately ?

okuzan: IMHO readability should take precedence over a negligible performance gain, as maintainability and clarity of code are generally more important

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,168 @@
...
     @Override
     public boolean remove(T object) {
+        for (Node<T> current = first; current != null; current = current.next) {
+            if (object == null && current.value == null) {
+                unlinkNode(current);
+                return true;
+            } else if (object != null && object.equals(current.value)) {
+                unlinkNode(current);
+                return true;
+            }
```
okuzan, line 73: body is the same in both cases, merge conditions together
look up code for Objects.equals() and write the same here
a == b || a != null && a.equals(b)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,168 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private static final int SIZE_DIVIDER = 2;
+    private static final int SIZE_SHIFTER = 1;
+    private static final int ZERO_SIZE = 0;
```
okuzan, line 8: concept of zero is unlikely to change :)) You can hardcode such self-explaining values

okuzan (changes requested):

kozhukhovsky (approved):

PR #1315, opened 2024-01-24
hw-jv-linked-list

VovaMaryniak:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,148 @@
...
+        private T data;
+        private Node<T> next;
+        private Node<T> prev;
+
+        Node(T data, Node<T> prev, Node<T> next) {
+            this.data = data;
+            this.prev = prev;
+            this.next = next;
+        }
+    }
```
olekskov, line 21: Better to move private inner class after public methods so we have main functionality in the first place

olekskov (approved):
Good job

PR #1318, opened 2024-01-26
Implemented solution

breakingtumble:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
...
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    public MyLinkedList() {
+        head = null;
+        tail = null;
+    }
```
okuzan, line 13: null is a default value for any object
this consturctor doesn't add any new logic to the class
Default constructor with no parameters exists implicitly in each class

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
...
+        for (T t : list) {
+            add(t);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        indexCheckForGetSetRemoveOperations(index);
+        return findNodeByIndex(index).item;
```
okuzan, line 58: move index checker in find by index method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
...
-        return false;
+        return size == 0;
+    }
+
+    private static class Node<T> {
+        private T item;
+        private Node<T> prev;
+        private Node<T> next;
+
+        public Node(T item, Node<T> prev, Node<T> next) {
```
okuzan, line 109:
```suggestion
        Node(T item, Node<T> prev, Node<T> next) {
```
won't be accessible outside of the file anyway

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
...
+
+    private void indexCheckForAddOperation(int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Invalid index");
+        }
+    }
+
+    private void indexCheckForGetSetRemoveOperations(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Invalid index");
```
okuzan, line 124: include index and size in the message to facilitate debugging

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
...
+                head = nodeToAdd;
+                nodeToAdd.next = foundNode;
+            }
+            size++;
+        }
     }
 
     @Override
     public void addAll(List<T> list) {
+        for (T t : list) {
```
okuzan, line 50:
```suggestion
        for (T node : list) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
...
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        indexCheckForAddOperation(index);
+        Node<T> foundNode = findNodeByIndex(index);
+        if (index == size) {
+            add(value);
+        } else {
```
okuzan, line 34: let's call return here and unwrap the rest of the code, returning early is generally [a better practice](https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement)

okuzan (changes requested):
Good job!

okuzan (approved):
Well done

PR #1333, opened 2024-02-15
completed the task

esidarapp:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,173 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size;
+    private Node<T> first;
+    private Node<T> last;
+
+    public MyLinkedList() {
+        this.first = null;
+        this.last = null;
```
okuzan, line 12: [where should you use this](stackoverflow.com/a/2411283)
does initializing with null change anything? What is the default value of uninitialized object?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,173 @@
...
+        private Node<T> prev;
+
+        Node(Node<T> prev, T element, Node<T> next) {
+            this.item = element;
+            this.next = next;
+            this.prev = prev;
+        }
+    }
+
+    public void linkFirst(T t) {
```
okuzan, line 27: public or private?
and remember that private methods go after public ones

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,173 @@
...
     public void addAll(List<T> list) {
+        for (T t : list) {
+            add(t);
+        }
+    }
+
+    private Node<T> getNode(int index) {
+        if (index >= size || index < 0) {
+            throw new IndexOutOfBoundsException("Wrong index " + index);
+        }
```
okuzan, line 86: include size in the message too
repeated logic, extract to a private utility class

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,173 @@
...
+        if (index < (size >> 1)) {
```
okuzan, line 87: bitwise operation provide negligible gain in our case, readability is more important, better stick to the usual operators

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,173 @@
...
+            Node<T> x = first;
+            for (int i = 0; i < index; i++) {
+                x = x.next;
+            }
+            return x;
```
okuzan, line 92: move common part from if-else

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,173 @@
...
+            int index = 0;
+            while (x != null) {
+                if (object.equals(x.item)) {
+                    remove(index);
+                    return true;
+                }
+                x = x.next;
+                index++;
+            }
+        }
```
okuzan, line 145: remove if-else, merge into one block, use equals condition as in Objects.equals (but you should write your own)

okuzan (changes requested):
Good job!

kozhukhovsky (approved):

PR #1341, opened 2024-02-21
done

ChabVlad:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,185 @@
...
     @Override
     public void add(T value) {
+        Node<T> newNode = new Node<T>(value);
+        if (size > 0) {
+            newNode.prev = tail;
+            tail.next = newNode;
+            tail = newNode;
+            size++;
+        }
+        ifEmptyList(newNode);
```
Vasul-Pankiv, line 19: really bad naming

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,185 @@
...
+        Node<T> current = head;
+        for (int i = 0; i < index; i++) {
+            current = current.next;
+        }
+        return current;
+    }
+
+    private Node<T> getNode(T object) {
+        Node<T> current = head;
+        for (int i = 0; i < size; i++) {
```
Vasul-Pankiv, line 161: better to use while loop in such cases

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,199 @@
...
     }
 
     @Override
     public void add(T value, int index) {
+        checkIndexSizeMore(index);
+        if (index == size) {
+            add(value);
+        } else {
+            addIfListFilled(value, index);
+        }
```
olekskov, line 29: You don't need checkIndexSizeMore method
```suggestion
        if (index == size) {
            add(value);
            return;
        }
        checkIndex(index);
        addByIndex(value, index);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,199 @@
...
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Index: " + index + ", size: " + size());
+        }
+    }
+
+    private void checkIndexSizeMore(int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Index: " + index + ", size: " + size());
+        }
+    }
```
olekskov, line 137:
```suggestion
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,199 @@
...
+                i--;
+            }
+        }
+        return current;
+    }
+
+    private Node<T> getNode(T object) {
+        Node<T> current;
+        int index = 0;
+        if (index <= size / 2) {
```
olekskov, line 162: You just created index = 0 and then making `if` block.
There is no sense. Index is always 0
When we search by object we cannot determine if we should start from head or tail. So just choose one option and search for element.

Approach is correct when we search by index and can understand it's closer to head or tail

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,199 @@
...
-        return false;
+        return size == 0;
+    }
+
+    private void ifEmptyList(Node<T> node) {
+        if (isEmpty()) {
+            head = node;
+            tail = node;
+        }
+    }
```
olekskov, line 109: this method is used only once and seems makes logic more comlicated
```suggestion
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,199 @@
...
     @Override
     public void add(T value) {
+        Node<T> addedNode = new Node<T>(value);
+        if (size > 0) {
+            addedNode.prev = tail;
+            tail.next = addedNode;
+            tail = addedNode;
+        }
+        ifEmptyList(addedNode);
+        size++;
```
olekskov, line 19: Just put removed logic here
```suggestion
        Node<T> node = new Node<T>(value);
        if (size > 0) {
            node.prev = tail;
            tail.next = node;
            tail = node;
        } else {
            head = node;
            tail = node;
        }
        size++;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,199 @@
...
+    }
+
+    private void ifEmptyList(Node<T> node) {
+        if (isEmpty()) {
+            head = node;
+            tail = node;
+        }
+    }
+
+    private void addIfListFilled(T value, int index) {
```
olekskov, line 111: filled with what?
```suggestion
    private void addByIndex(T value, int index) {
```

olekskov (changes requested):

kozhukhovsky (approved):

PR #1344, opened 2024-02-22
(^-^)

Bohadan-Markatov:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,127 @@
...
+        Node<T> previousNode = findPreviousNode(index);
+        Node<T> newNode = new Node<>(previousNode, value, previousNode.next);
+        previousNode.next.prev = newNode;
+        previousNode.next = newNode;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
+        for (int i = 0; i < list.size(); i++) {
```
ReamFOX, line 47: Here you can use for each, then it would look even better than it is

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,127 @@
...
+        checkIndexForAdd(index);
+        Node<T> currentNode = head;
+        for (int i = 0; i < index; i++) {
+            currentNode = currentNode.next;
+        }
+        return currentNode;
+    }
+
+    private void checkIndexForAdd(int index) {
+        if (index < 0 || index > size) {
```
ReamFOX, line 105: I passed the size to checkIndex separately so as not to duplicate the method, but maybe your method is more correct.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,127 @@
...
+    }
+
+    private Node<T> getNodeByIndex(int index) {
+        checkIndex(index);
+        Node<T> currentNode = head;
+        for (int i = 0; i <= index; i++) {
+            currentNode = currentNode.next;
+        }
+        return currentNode;
+    }
```
ReamFOX, line 117: The only remark is the implementation of getNodeByIndex.
'''
Don't iterate your nodes from the head to the tail or vice versa.
Pay attention to the index you receive as an input, you can iterate only the first or second half of the list, depending on the index value. That will boost your lists' performance.
'''

ReamFOX (changes requested):
Looks better than mine👌😊
But need some improvements.

PR #1345, opened 2024-02-22
Jv ll

ReamFOX:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,191 @@
...
+            current = tail;
+            for (int i = size - 1; i > index; i--) {
+                current = current.prev;
+            }
+        } else {
+            current = head;
+            for (int i = 0; i < index; i++) {
+                current = current.next;
+            }
+        }
```
k0sm0naft, line 131: You can simplify it:
```suggestion
        current = (index >= size / 2) ? tail : head;
        int iterations = (index >= size / 2) ? size - index - 1 : index;
        for (int i = 0; i < iterations; i++) {
             current = (index >= size / 2) ? current.prev : current.next;
        }
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,191 @@
...
 
     @Override
     public void add(T value, int index) {
+        checkIndex(index, size + 1);
+        Node<T> tempNode = new Node<>(value);
+        if (index == 0) {
+            addNodeToHead(tempNode);
+        } else {
+            Node<T> currentNode = findNodeByIndex(index - 1);
+            addNodeAfter(currentNode, tempNode);
```
k0sm0naft, line 35: Here you could get delete this '1' and use, for example, addNodeBefore instead of addNodeAfter)). 
But if it works anyway, leave it. 
And it would probably be easier if `addNodeAfter` was inside this method, because it is not repeated anywhere and readability is improved.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,191 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void checkIndex(int index, int size) {
```
k0sm0naft, line 113: Would be better.
```suggestion
    private void checkIndex(int index) {
```

k0sm0naft (changes requested):
For me it is complicated to read your code. But it work, so good job!

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -143,14 +133,48 @@ private Node<T> findNodeByIndex(int index) {
...
+        if (head != null) {
+            head.prev = newNode;
+        }
+        head = newNode;
+        if (tail == null) {
+            tail = newNode;
+        }
     }
 
     static class Node<T> {
```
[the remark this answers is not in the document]
ReamFOX: Yeah, u right, thanks.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -26,13 +26,20 @@ public void add(T value) {
...
-        checkIndex(index, size + 1);
-        Node<T> tempNode = new Node<>(value);
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Index out of bounds: " + index);
+        }
+        Node<T> tempNode = new Node<>(null, value, null);
         if (index == 0) {
             addNodeToHead(tempNode);
         } else {
             Node<T> currentNode = findNodeByIndex(index - 1);
```
k0sm0naft, line 36: I mean in this case you can work with node `findNodeByIndex(index)`.

k0sm0naft (approved):
GJ!!

ReamFOX:
You mean like that?😅

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,182 @@
...
+            addNodeToHead(tempNode);
+        } else {
+            Node<T> currentNode = findNodeByIndex(index - 1);
+            tempNode.next = currentNode.next;
+            tempNode.prev = currentNode;
+            currentNode.next = tempNode;
+            if (tempNode.next != null) {
+                tempNode.next.prev = tempNode;
+            }
+            if (index == size) {
```
andrii-hoienko, line 40: I beleive this logic should be outside else block, because we have 3 possible cases: addToHead, toMiddle, toTail. And this checks should be on the same level

andrii-hoienko (changes requested):
Nice, left minor comments

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,176 @@
...
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Index: " + index + ", Size: " + size);
+        }
+    }
+
+    private Node<T> findNodeByIndex(int index) {
+        Node<T> current = (index >= size / 2) ? tail : head;
```
andrii-hoienko, line 123: code duplication

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,176 @@
...
+            addNodeToTail(tempNode);
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Index: " + index + ", Size: " + size);
+        }
```
andrii-hoienko, line 30:
```suggestion
        if (index == size) {
            throw new IndexOutOfBoundsException("Index: " + index + ", Size: " + size);
        }
        checkIndex(index);
```

andrii-hoienko (changes requested):
Final charge

PR #1346, opened 2024-02-22
hw-jv-linked-list

k0sm0naft:
Implemented all the required methods in MyLinkedList class.

### DIFF
--- a/src/main/java/core/basesyntax/MyLinkedList.java
+++ b/src/main/java/core/basesyntax/MyLinkedList.java
@@ -10,7 +10,7 @@ public class MyLinkedList<T> implements MyLinkedListInterface<T> {
     @Override
     public void add(T value) {
         Node<T> newNode = new Node<>(null, value, null);
-        if (head == null) {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,136 @@
...
+                } else {
+                    currentNode.next.prev = currentNode.prev;
+                }
+                size--;
+                return true;
+            }
+            currentNode = currentNode.next;
+        }
         return false;
     }
```
SkillfulDev, line 98: In these two methods we need to create method unlink(Node node) according checklist 
image

SkillfulDev (approved):

src/main/java/core/basesyntax/MyLinkedList.java
k0sm0naft: You're right. Didn't sean that cod is the same.

PR #1347, opened 2024-02-22
MyLinkedList interface was implemented

slizko1:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,182 @@
...
+        if (size == 0) {
+            head = new Node(null, value, null);
+            tail = head;
+            size++;
+        } else if (size > 0) {
+            Node lastNode = tail;
+            tail = new Node(lastNode, value, null);
+            lastNode.next = tail;
+            size++;
+        }
```
EduarfF, line 21: You can write size++ after if else

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,178 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size = 0;
```
karma-o, line 6: primitive class fields have default values
```suggestion
    private int size;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,178 @@
...
+        return (object == null && node.value == null)
+                || (node.value != null && node.value.equals(object));
+    }
+
+    private Node findNodeByIndex(int index) {
+        if (index == 0) {
+            return head;
+        }
+        if (index == size) {
+            return tail;
```
karma-o, line 124: I think it should work without head/tail checks, by doing 0 iterations, in the for loop, and I think we should add index check here just to be sure, and remove this check where it'll become duplicated

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,178 @@
...
+
+    private Node findFromTail(int index) {
+        Node currentNode = tail;
+        for (int i = size - 1; i != index; i--) {
+            currentNode = currentNode.prev;
+        }
+        return currentNode;
+    }
+
+    private void unlink(Node node) {
```
karma-o, line 149: * size-- can come out of those `if`s to the end of the method to not duplicate it
* there is no point in using both `else` and `return`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,178 @@
...
+            return;
+        } else if (node == tail) {
+            Node previousNode = node.prev;
+            previousNode.next = null;
+            node.prev = null;
+            tail = previousNode;
+            size--;
+            return;
+        } else {
+            Node preciousNode = node.prev;
```
karma-o, line 170:
```suggestion
            Node previousNode = node.prev;
```

src/main/java/core/basesyntax/MyLinkedListInterface.java
```diff
@@ -21,3 +21,4 @@ public interface MyLinkedListInterface<T> {
 
     boolean isEmpty();
 }
+
```
karma-o, line 24:
```suggestion
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,178 @@
...
+        node.value = value;
+        return oldValue;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        isIndexExist(index);
+        T value = (T) findNodeByIndex(index).value;
+        unlink(findNodeByIndex(index));
```
karma-o, line 69: `findNodeByIndex()` is an expensive operation, do not call it 2 times

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,178 @@
...
+        private Node<T> prev;
+
+        private Node(Node<T> prev, T element, Node<T> next) {
+            this.value = element;
+            this.next = next;
+            this.prev = prev;
+        }
+    }
+
+    private void isIndexExist(int index) {
```
karma-o, line 108: name implies that it should return boolean, but it does not, so let's name it like `checkIndex()`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,178 @@
...
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException();
+        }
+    }
+
+    private boolean needToRemove(T object, Node node) {
```
karma-o, line 114: let's change this method a bit to make it more pretty and understandable:
* since this method basically compares values, would be nice receive values of the same type
* if all it does is comparing, then let's call it accordingly
```suggestion
    private boolean areValuesEqual(T first, T second) {
```

karma-o (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,172 @@
...
+            previousNode.next = nextNode;
+            nextNode.prev = previousNode;
+            node.next = null;
+            node.prev = null;
+        }
+        size--;
+
     }
 }
+
```
karma-o, line 174:
```suggestion
    }
}
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,172 @@
...
+        private T value;
+        private Node<T> next;
+        private Node<T> prev;
+
+        private Node(Node<T> prev, T element, Node<T> next) {
+            this.value = element;
+            this.next = next;
+            this.prev = prev;
+        }
+
```
karma-o, line 109:
```suggestion
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,172 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size;
+    private Node head;
+    private Node tail;
```
karma-o, line 8: Do not use raw classes when you have generics, I think IDEA also highlights it
```suggestion
    private Node<T> head;
    private Node<T> tail;
```

karma-o (approved):
GJ, please also see some final comments for this task

PR #1348, opened 2024-02-24
Hm jv linked list

vladshender:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,191 @@
...
 
 import java.util.List;
+import java.util.Objects;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head = null;
+    private Node<T> tail = null;
+    private int size = 0;
+
+    static class Node<T> {
```
olekskov, line 11: Better to put inner class after public methods in order to have public logic before details of realization

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,191 @@
 package core.basesyntax;
 
 import java.util.List;
+import java.util.Objects;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head = null;
+    private Node<T> tail = null;
+    private int size = 0;
```
olekskov, line 9:
```suggestion
    private Node<T> head;
    private Node<T> tail;
    private int size;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,191 @@
...
+            Node<T> currentNode = head;
+            for (int i = 0; i < index; i++) {
+                currentNode = currentNode.next;
+            }
+            return currentNode;
+        }
+        return null;
+    }
+
+    public boolean checkIndex(int index) {
```
olekskov, line 183: Naming should give idea of return type. checkIndex looks like void method, not boolean
```suggestion
    public boolean isValidIndex(int index) {
```

But as you throw exception inside, this method cannot return false at all. So maybe better to make it void and call without if statements in place of usage?
```suggestion
    public void checkIndex(int index) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,191 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size() == 0;
+    }
+
+    public T unlink(Node<T> object) {
```
olekskov, line 129: Please do not make all service methods public. Only official logic should be public. All details of the realization have more strict modifiers.
Please recheck all methods

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,191 @@
...
     }
 
     @Override
     public void add(T value, int index) {
+        Node<T> currentNode;
+        if (index == 0) {
+            if (head == null) {
+                Node<T> newNode = new Node<T>(null, value, null);
+                head = newNode;
+                tail = head;
```
olekskov, line 44: duplicates code from add(T value) method
```suggestion
                add(value);
                return;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,191 @@
...
+            } else {
+                currentNode = new Node<>(null, value, head);
+                head.prev = currentNode;
+                head = currentNode;
+            }
+        } else if (index == size) {
+            add(value);
+            return;
```
olekskov, line 52: And this code also becomes duplicated after change above, please try to rearrange conditions to avoid it.
Also would be nice to see reusage of searchNodeByIndex(int index) method and introduction of insertBefore or InsertAfter method after it instead of raw logic.
Because overall method is too heavy

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,191 @@
...
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        T deleteElement = null;
+        if (checkIndex(index)) {
+            deleteElement = (index == 0) ? unlinkFirst()
+                    : (index == size - 1) ? unlinkLast()
+                    : unlink(searchNodeByIndex(index));
```
olekskov, line 96: This ternary operator is very hard to read. Please play with code formatting or use if/else structure

olekskov (changes requested):

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+public class Main {
+    public static void main(String[] args) {
```
okuzan, line 4: `git rm --cached Main.java`
remove this from the repo

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,202 @@
...
 
     @Override
     public void add(T value, int index) {
+        Node<T> currentNode;
+        if (index == 0) {
+            if (head == null) {
+                addFirst(value);
+            } else {
+                insertBefore(value);
+            }
```
okuzan, line 29: duplicated logic, just call `add()` here

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,202 @@
...
+        Node<T> currentNode = searchNodeByElement(object);
+        if (currentNode == null) {
+            return false;
+        } else if (currentNode.equals(head)) {
+            unlinkFirst();
+            return true;
+        } else if (currentNode.equals(tail)) {
+            unlinkLast();
+            return true;
+        }
```
okuzan, line 93: we don't need `else` after `return`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,202 @@
...
+    private boolean checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Index: "
+                    + index
+                    + " don`t exist");
+        }
+        return true;
+    }
+
+    static class Node<T> {
```
okuzan, line 191:
```suggestion
    private static class Node<T> {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,202 @@
...
+        private T item;
+        private Node<T> next;
+        private Node<T> prev;
+
+        public Node(Node<T> prev, T item, Node<T> next) {
```
okuzan, line 196:
```suggestion
        Node(Node<T> prev, T item, Node<T> next) {
```
```suggestion
        private Node(Node<T> prev, T item, Node<T> next) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,202 @@
...
+            Node<T> currentNode = head;
+            for (int i = 0; i < index; i++) {
+                currentNode = currentNode.next;
+            }
+            return currentNode;
+        }
+        return null;
+    }
+
+    private boolean checkIndex(int index) {
```
okuzan, line 182: not fixed, name doesn't tell us boolean will be returned. Let's make it void
```suggestion
    private void checkIndex(int index) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,202 @@
...
+
+    private Node<T> searchNodeByIndex(int index) {
+        if (checkIndex(index)) {
+            Node<T> currentNode = head;
+            for (int i = 0; i < index; i++) {
+                currentNode = currentNode.next;
+            }
+            return currentNode;
+        }
+        return null;
```
okuzan, line 179: now checkIndex acts like a guard clause, unnesting rest of the code, it's a more accepted [pattern](https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement)
```suggestion
        checkIndex(index));
        Node<T> currentNode = head;
        for (int i = 0; i < index; i++) {
            currentNode = currentNode.next;
        }
        ... iterate from tail if it's closer
        return currentNode;
```
[checklist issue](https://github.com/mate-academy/jv-linked-list/blob/master/checklist.md#dont-iterate-your-nodes-from-the-head-to-the-tail-or-vice-versa) - you should iterate from closest end

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,202 @@
...
+    }
+
+    private boolean checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Index: "
+                    + index
+                    + " don`t exist");
```
okuzan, line 186: including size would facilitate debugging for cases when index >= size

okuzan (changes requested):
Good job since last review

vladshender:
I couldn't add the add(value) method to the add(value, index) method because add(value) is a different method than the code block you highlighted
insertBefore() and insertAfter()

vladshender:
git rm --cached Main.java
fatal: pathspec 'Main.java' did not match any files

okuzan:
> I couldn't add the add(value) method ..
Sorry, I commented to the wrong ` if` , it's the same as case when ` index == size`, but in your code is actually shorter, so just leave as is 
> 'Main.java' did not match any files
make sure you're running it in the folder which actually contains this file

okuzan (changes requested):
Follow idea's suggestions (yellow warnings) and clean up your code, there are a lot of redundant assignments 
Remove Main, ask for help in the chat if needed
image

okuzan (approved):
Well done 👍

PR #1349, opened 2024-02-26
main version

EduarfF:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,151 @@
 package core.basesyntax;
 
 import java.util.List;
+import java.util.Objects;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
```
makaroshyna, line 7: Don't you need to have tail?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,151 @@
...
+    private int size;
+
+    public MyLinkedList() {
+        head = null;
+        size = 0;
```
makaroshyna, line 12: It is implied that they are that size, no need to write it.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,151 @@
...
+        if (head == null) {
+            head = newNode;
+        } else {
+            Node<T> current = head;
+            while (current.next != null) {
+                current = current.next;
+            }
+            current.next = newNode;
+        }
+        size++;
```
makaroshyna, line 27: With tail variable it will be easier to add to the end.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,151 @@
...
+
+    private Node<T> getNode(int index) {
+        Node<T> current = head;
+        for (int i = 0; i < index; ++i) {
+            current = current.next;
+        }
+        return current;
+    }
+
+    private void isNotValidIndex(int index) {
```
makaroshyna, line 136: Change the naming, it is not a boolean to be named like that. Maybe indexValidation or checkIndex

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,151 @@
...
+        if (index >= size || index < 0) {
+            throw new IndexOutOfBoundsException("Index: " + index + ", Size: " + size);
+        }
+    }
+
+    private class Node<T> {
+        private T element;
+        private Node<T> next;
```
makaroshyna, line 144: Add prev to refer to the previous element of array.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,151 @@
...
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private Node<T> getNode(int index) {
+        Node<T> current = head;
+        for (int i = 0; i < index; ++i) {
+            current = current.next;
+        }
```
makaroshyna, line 132: You should check the index position first, and than start from head or tail.
image

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,151 @@
...
+                } else {
+                    prev.next = current.next;
+                }
+                size--;
+                return true;
+            }
+            prev = current;
+            current = current.next;
+        }
+
```
makaroshyna, line 114: You should create a new method unlink as was mentioned in common mistakes.
image

makaroshyna (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
andrii-hoienko: agree

src/main/java/core/basesyntax/MyLinkedList.java
andrii-hoienko: agree

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
...
+        }
+        tail = newNode;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Index: " + index + ", Size: " + size);
+        }
```
andrii-hoienko, line 26: Move to a separate method or think how to reuse already existing indexValidation

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
...
+        if (index == size) {
+            add(value);
+            return;
+        }
+
+        Node<T> newNode = new Node<>(value);
+        if (index == 0) {
+            newNode.next = head;
+            head = newNode;
```
andrii-hoienko, line 35: I suggest to move this code to new method addToHead

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
...
+        } else {
+            Node<T> prevNode = getNode(index - 1);
+            newNode.next = prevNode.next;
+            prevNode.next = newNode;
```
andrii-hoienko, line 39: I suggest to move this code to new method addByIndex

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
...
     public boolean remove(T object) {
+        if (isEmpty()) {
+            return false;
+        }
+
+        Node<T> current = head;
+        Node<T> prev = null;
+
+        while (current != null) {
+            if (object == null && current.element == null) {
```
andrii-hoienko, line 99: if and else if could be collapsed to avoid code duplication

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
...
+        return current;
+    }
+
+    private void validateIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Index: " + index + ", Size: " + size);
+        }
+    }
+
+    private void unlink(Node<T> node, Node<T> prev) {
```
andrii-hoienko, line 140: Why unlink should accept as argument node and prev, we can always take prev from node, no reason to accept it

andrii-hoienko (changes requested):
Good job, but some improvements required

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,180 @@
...
+        }
+        size++;
+    }
+
+    private void addToHead(Node<T> newNode) {
+        newNode.next = head;
+        head = newNode;
+    }
+
+    private void addByIndex(Node<T> newNode, int index) {
```
karma-o, line 44: all private methods should go after all public ones

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,180 @@
...
-        return null;
+        validateIndex(index);
+        Node<T> node = getNode(index);
+        T oldValue = node.element;
+        node.element = value;
+        return oldValue;
     }
 
     @Override
     public T remove(int index) {
```
karma-o, line 73: why don't we use `unlink()` method here?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,180 @@
...
+        }
+
+        Node<T> current = head;
+        Node<T> prev = null;
+
+        while (current != null) {
+            if ((object == null && current.element == null)
+                    || (object != null && object.equals(current.element))) {
+                unlink(current);
+                size--;
```
karma-o, line 108: I think `size--` can be a part of `unlink()`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,180 @@
...
+            }
+        } else {
+            prev.next = node.next;
+            if (prev.next == null) {
+                tail = prev;
+            }
+        }
+    }
+
+    private Node<T> findPreviousNode(Node<T> targetNode) {
```
karma-o, line 159: it is too complicated, remove that, and just add `prev` field to the `Node` class, as it was pointed out in the 2 previous reviews, so please be attentive to ALL comments. They must be either fixed as suggested or commented on why not

src/main/java/core/basesyntax/MyLinkedList.java
karma-o: not fixed

src/main/java/core/basesyntax/MyLinkedList.java
karma-o: not fixed

karma-o (changes requested):
Not resolved comments + a couple of new ones

PR #1350, opened 2024-02-26
Implement linked list.

makaroshyna:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,147 @@
...
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException(
+                    "Index " + index + " is out of bounds " + size + "!");
+        }
+
+        if (index == size) {
+            add(value);
+        } else {
```
k0sm0naft, line 32: You already have logic to check index. For example:
```suggestion
       if (index == size) {
           add(value);
       } else {
           checkIndex(index)
```

k0sm0naft, line 32: You already have logic to check index. For example:
```suggestion
       if (index == size) {
           add(value);
       } else {
           checkIndex(index)
```

k0sm0naft (approved):
In general great!

k0sm0naft (approved):

Recdt (approved):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,141 @@
...
+        Node<T> currentNode = head;
+        if (index < (size / 2)) {
+            for (int i = 0; i < index; i++) {
+                currentNode = currentNode.next;
+            }
+        } else {
+            for (int i = index; i > 0; i--) {
+                currentNode = currentNode.next;
+            }
+        }
```
karma-o, line 116: you are iterating only from the start to the end of the nodes list because only `currentNode.next` is being called not `previous`, and start is always from the head, but the key to optimize performance if we know that index is closer to an end - we can start iteration from the tail to the index

karma-o (changes requested):
You did a very good job! Left a single comment 🔥

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -104,16 +105,14 @@ private void unlink(Node<T> node) {
...
-                currentNode = currentNode.next;
-            }
+        boolean isFirstHalf = size / 2 >= index;
+        Node<T> currentNode = isFirstHalf ? head : tail;
+        int iteration = isFirstHalf ? index : size - index - 1;
+
+        for (int i = 0; i < iteration; i++) {
+            currentNode = isFirstHalf ? currentNode.next : currentNode.prev;
         }
+
```
karma-o, line 115: I'd suggest just making 2 methods: `findFromHead()` & `findFromTail()` and calling them based on the `isFirstHalf` with a bit fewer ternary operators to improve readability

karma-o (approved):
Well done, left a minor suggestion 😉

PR #1354, opened 2024-02-27
Implemented my version of LinkedList.

nalforov:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,202 @@
...
+            currenNode = currenNode.nextNode;
+        }
+        throw new NoSuchElementException("No such element in list");
+    }
+
+    private String getBoundariesMessage(int index) {
+        return "Index: " + index + ", Size:" + size;
+    }
+
+    private class Node<T> {
```
okuzan, line 191:
```suggestion
    private static class Node<T> {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,202 @@
...
+
+    private Node<T> findFirstNodeByValue(T value) {
+        Node<T> currenNode = head;
+        while (currenNode != null) {
+            if (currenNode.nodeValue == value
+                    || (currenNode.nodeValue != null && currenNode.nodeValue.equals(value))) {
+                return currenNode;
+            }
+            currenNode = currenNode.nextNode;
+        }
```
okuzan, line 183: method is used only once, better move this logic to remove by value. Throwing exception here and catching in remove isn't the best way to go

okuzan (changes requested):
Good job!

a-kozhukhovsky (approved):

a-kozhukhovsky (approved):

kozhukhovsky (approved):

PR #1355, opened 2024-02-27
impemented all methods

DmytroHadiuchko:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,155 @@
...
+            add(item);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index < NON_NEGATIVE_INDEX || index >= size) {
+            throw new IndexOutOfBoundsException(INDEX + index + SIZE + size);
+        }
```
nklimovych, line 65: This code can be added to separate method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,155 @@
 package core.basesyntax;
 
 import java.util.List;
+import java.util.Objects;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private static final String INDEX = "Index: ";
+    private static final String SIZE = ", Size: ";
+    private static final int NON_NEGATIVE_INDEX = 0;
```
nklimovych, line 9: It would be better to find out better name for this constatnt couse every number biger than 0 is non negative

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,155 @@
...
+    }
+
+    private static class Node<T> {
+        private T value;
+        private Node<T> next;
+        private Node<T> prev;
+
+        public Node(T value) {
+            this.value = value;
+        }
```
nklimovych, line 153: At this inner class it would be better to make constructor for all three variables, not only value

nklimovych (changes requested):
Look good, but there are some suggestions

PR #1356, opened 2024-02-27
Added implementation for task jv-linked-list.

nklimovych:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,164 @@
...
+            tail = node;
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Invalid index: " + index);
+        }
```
fedaktaras, line 28:
```suggestion
        validateIndex(index)
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,164 @@
...
+                node = node.prev;
+            }
+        }
+        return node;
+    }
+
+    private Node<T> findNodeByValue(T item) {
+        Node<T> node = head;
+        while (node != null) {
+            if ((item == null && node.value == null) || (item != null && item.equals(node.value))) {
```
fedaktaras, line 121:
```suggestion
    private Node<T> findNodeByValue(T item) {
        Node<T> node = head;
        while (node != null) {
            if (data == object || object != null && object.equals(data)) {
```
if it has same pace at heap no need to call equals() and == also checks if boths are null

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,164 @@
...
+            node.prev = null;
+        }
+
+        if (next != null) {
+            next.prev = prev;
+            node.next = null;
+        }
+
+        size--;
+    }
```
fedaktaras, line 151: Good method.

fedaktaras (commented):
Really good just one change required.

fedaktaras (changes requested):
All is good. Just one change required.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,174 @@
...
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        validateAddIndex(index);
+
+        if (index == size) {
+            add(value);
+        } else {
```
andrii-hoienko, line 32: I suggest add 3 if`s in this method: 1 if head, 2 if tail, 3 if in middle in will make code more readable

karma-o: I would also suggest using `return` instead of `else`, and I do not see the changes that Andrii proposed, so please see what we can do about this or leave a comment on why you think otherwise

karma-o (changes requested):
1 comment is still not resolved

src/main/java/core/basesyntax/MyLinkedList.java
nklimovych: I was trying to implement Andrii suggestionfor more then an hour, but tets were failing every tyme...
Maybe i dont get it right...

nklimovych: > I would also suggest using `return` instead of `else`, and I do not see the changes that Andrii proposed, so please see what we can do about this or leave a comment on why you think otherwise

As i understend u, i should do like this: (this solution is passing all tests)
```@Override
        validateIndexOnAdd(index);

        if (index == size) {
            add(value);
            return;
        }
        
        Node<T> nodeAtIndex = findNodeByIndex(index);
        Node<T> node = new Node<>(nodeAtIndex.prev, value, nodeAtIndex);

        if (index == 0) {
            head = node;
        } else {
            nodeAtIndex.prev.next = node;
        }
        nodeAtIndex.prev = node;
        size++;
    }

karma-o (approved):

PR #1359, opened 2024-02-28
Implemented solution for the LinkedList task

ghost:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,169 @@
...
+            node.prev = nodeByIndex.prev;
+            nodeByIndex.prev.next = node;
+            nodeByIndex.prev = node;
+            node.next = nodeByIndex;
+        }
+    }
+
+    private void throwException() {
+        throw new IndexOutOfBoundsException(EXCEPTION_MESSAGE);
+    }
```
JohnGrey17, line 133: i guess we can remove that method and throw new Exception when it appear

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,169 @@
...
+        return resultNode;
+    }
+
+    private static class Node<T> {
+        private Node<T> prev;
+        private Node<T> next;
+        private T item;
+
+        public Node(Node<T> previousNode, T item, Node<T> nextNode) {
+            this.prev = previousNode;
```
JohnGrey17, line 166: for better readability put that class declaration on top

JohnGrey17 (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,169 @@
...
+        }
+        return resultNode;
+    }
+
+    private static class Node<T> {
+        private Node<T> prev;
+        private Node<T> next;
+        private T item;
+
+        public Node(Node<T> previousNode, T item, Node<T> nextNode) {
```
JohnGrey17, line 165: for better readability put that class declaration on top

JohnGrey17 (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
VlasenkoYevhenii: It's a very discussable point. As I know, depends on the team codestyle where the inner class declaration should be located. I'm guided by the logic that since it's private, it should be declared after all the public stuff

src/main/java/core/basesyntax/MyLinkedList.java
VlasenkoYevhenii: It shortens the methods it's being used in. I think it looks cute =)

karma-o: but it also does not give us information on what is this exception/message and reduces readability + a method 1 statement long is not always, but usually a not-that-great idea :)

src/main/java/core/basesyntax/MyLinkedList.java
karma-o: it does actually depend on the team, but I lean towards the end of the file since it is private

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,169 @@
...
+        return returnValue;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        validateIndex(index);
+        Node<T> nodeToRemove = getNodeByIndex(index);
+        unlink(nodeToRemove);
+        size--;
```
karma-o, line 61: I think `size--` can be a part of `unlink()`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,169 @@
...
+        if (nodeToUnlink.equals(head)) {
+            head = head.next;
+            nodeToUnlink.next = null;
+        } else if (nodeToUnlink.equals(tail)) {
+            tail = tail.prev;
+            nodeToUnlink.prev = null;
+        } else {
+            nodeToUnlink.prev.next = nodeToUnlink.next;
+            nodeToUnlink.next.prev = nodeToUnlink.prev;
+        }
```
karma-o, line 103: what if it is a head and a tail at the same time?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,169 @@
...
+    private void validateIndex(int index) {
+        if (index < 0 || index > size) {
+            throwException();
+        }
+    }
+
+    private Node<T> getNodeByIndex(int index) {
+        if (index >= size || index < 0) {
+            throw new IndexOutOfBoundsException(EXCEPTION_MESSAGE);
+        }
```
karma-o, line 144: let's extract that check out of that method as well

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,169 @@
...
+            resultNode = head;
+            for (int i = 0; i < index; i++) {
+                resultNode = resultNode.next;
+            }
+        } else {
+            resultNode = tail;
+            for (int i = size - 1; i > index; i--) {
+                resultNode = resultNode.prev;
+            }
+        }
```
karma-o, line 156: here we also can get 2 methods, like `findFromTail()` and `findFromHead()`

karma-o (changes requested):
Well done, just minor things

yevheniisynytsia (approved):
🔥

PR #1360, opened 2024-02-28
Made custom LinkedList class with required methods

ValentynSharshon:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,160 @@
...
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException(String.format(ERROR_MESSAGE_TEMPLATE, index, size));
+        }
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException(String.format(ERROR_MESSAGE_TEMPLATE, index, size));
+        }
+    }
```
azzilian, line 90: Good idea, but basically those methods do the same job with same message template - so maybe be better to leave only one method?

azzilian (changes requested):
Your code is much better then my because of KISS implementations, but 1 thing I think must be changed

src/main/java/core/basesyntax/MyLinkedList.java
ValentynSharshon: Yes, you're totally right, the methods are very similar, but have a different range in the loop itself 
`(index < 0 || index >= size)`
`(index < 0 || index > size)`. 
I don't even know how to get out of this situation...

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,160 @@
...
+        T oldValue = node.value;
+        node.value = value;
+        return oldValue;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        Node<T> node = getNodeByIndex(index);
```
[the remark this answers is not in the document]
ValentynSharshon: Oh, sorry, my bad :disappointed:

PR #1361, opened 2024-02-28
hw solution

Shr1mpa:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,238 @@
...
+    private static final int NOT_FOUND = -1;
+    private static final int ONE = 1;
+    private static final int HALF = 2;
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    public MyLinkedList() {
+        head = tail = null;
+        size = 0;
```
kerrrusha, line 15: In Java, declaring the initial value of a variable when it's first declared is optional. When you write private int size = 0;, you're explicitly setting the initial value of size to 0. However, in Java, if you don't initialize an instance variable, it gets a default value based on its type. For int, the default initial value is 0 anyway. So, writing private int size = 0; is redundant in this case because size would already be initialized to 0 by default. You could simply write private int size; to achieve the same result.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,238 @@
...
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private static final int NOT_FOUND = -1;
+    private static final int ONE = 1;
+    private static final int HALF = 2;
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    public MyLinkedList() {
+        head = tail = null;
```
kerrrusha, line 14: they are null by default

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,238 @@
...
+        size = 0;
+    }
```
kerrrusha, line 16:
```suggestion
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,238 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private static final int NOT_FOUND = -1;
```
kerrrusha, line 6:
```suggestion
    private static final int NOT_FOUND_INDEX = -1;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,238 @@
...
+    private static final int ONE = 1;
+    private static final int HALF = 2;
```
kerrrusha, line 8: sense of extracting numbers to the constants is to give more explanations about them by defining constant name, but your name explains nothing, so it makes no sense
```suggestion
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,238 @@
...
+        size = 0;
+    }
+
+    private static class Node<T> {
+        private T item;
+        private Node<T> next;
+        private Node<T> prev;
+
+        public Node(Node<T> prev, T item, Node<T> next) {
+            this.item = item;
```
kerrrusha, line 24: why dont you save prev and next values?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,238 @@
...
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    public MyLinkedList() {
+        head = tail = null;
+        size = 0;
+    }
+
+    private static class Node<T> {
```
kerrrusha, line 18: inner classes should be located at the very end of the class

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,238 @@
...
+        newNode.next = head;
+        if (head != null) {
+            head.prev = newNode;
+        }
+        head = newNode;
+        tail = newNode;
+    }
+
+    private void addToMiddle(Node<T> newNode, int index) {
+        int half = size / HALF;
```
kerrrusha, line 187: not used var
```suggestion
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,238 @@
...
+                int indexFromEnd = size - index - ONE;
+                newNode = secondHalfFindNode(newNode, indexFromEnd);
+            }
+            return newNode;
+        } else {
+            throw new IndexOutOfBoundsException("Index out of bounds");
+        }
+    }
+
+    private int getNode(T value) {
```
kerrrusha, line 227:
```suggestion
    private int getNodeIndex(T value) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,238 @@
...
+        return size == 0;
+    }
+
+    private boolean isValidIndex(int index) {
+        if (index >= 0 && index <= size) {
+            return true;
+        } else {
+            throw new IndexOutOfBoundsException("Index out of bounds");
+        }
+    }
```
kerrrusha, line 161: we can make it more simple
```suggestion
    private void checkIndex(int index) {
        if (index < 0 || index > size) {
            throw new IndexOutOfBoundsException("Index out of bounds");
        }
    }
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,238 @@
...
+        newNode.next = nextNode;
+        newNode.prev = current;
+    }
+
+    private boolean firstHalf(int index) {
+        return index <= size / HALF;
+    }
+
+    private Node<T> getNode(int index) {
+        if (index >= 0 && index < size) {
```
kerrrusha, line 210:
```suggestion
        checkIndex(index);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,238 @@
...
+        for (T el: list) {
+            add(el);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        Node<T> newNode = null;
+        if (index >= 0 && index < size) {
```
kerrrusha, line 72:
```suggestion
        checkIndex(index);
```

koval-kn (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,231 @@
...
```
okuzan, line 54: returning early is a [better practice](https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement) in devs community - we can unnest the rest of the code - and so make it easier to read
In cases like this, invert condition - if NEGATIVE throw ; rest of code

you don't need else after returning/throwing, the rest of the code goes as is

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,231 @@
...
     public T set(T value, int index) {
-        return null;
+        Node<T> node = getNode(index);
+        if (node != null) {
+            T item = node.item;
+            node.item = value;
+            return item;
+        } else {
+            throw new IndexOutOfBoundsException("Invalid index");
+        }
```
okuzan, line 80: instead of this, just put this guarding method inside of a getNode. It shouldn't return null, it either returns an Object or exception is thrown

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,231 @@
...
     }
 
     @Override
     public T remove(int index) {
```
okuzan, line 84: [common mistake](https://github.com/mate-academy/jv-linked-list/blob/master/checklist.md#create-a-separate-method-unlinknode-node)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,231 @@
...
+    }
+
+    private void checkIndex(int index, boolean setOrRemove) {
+        if (setOrRemove) {
+            if (index < 0 || index >= size) {
+                throw new IndexOutOfBoundsException("Index out of bounds");
+            }
+        } else if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Index out of bounds");
+        }
```
okuzan, line 144: include index and size in the message, so it is useful for debugging

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,231 @@
...
+        return NOT_FOUND_INDEX;
+    }
+
+    private static class Node<T> {
+        private T item;
+        private Node<T> next;
+        private Node<T> prev;
+
+        public Node(Node<T> prev, T item, Node<T> next) {
```
okuzan, line 227: why public

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,231 @@
...
+        }
+        return newNode;
+    }
+
+    private int getNodeIndex(T value) {
+        Node<T> currentNode = head;
+        int count = 0;
+        while (currentNode != null) {
+            if ((currentNode.item == null && value == null)
+                    || (currentNode.item != null && currentNode.item.equals(value))) {
```
okuzan, line 213:
```suggestion
            if ((currentNode.item == value)
                    || (currentNode.item != null && currentNode.item.equals(value))) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,231 @@
...
+    }
+
+    private Node<T> secondHalfFindNode(Node<T> current, int index) {
+        for (int i = 0; i < index; i++) {
+            current = current.prev;
+        }
+        return current;
+    }
+
+    private void addToBegin(Node<T> newNode) {
```
okuzan, line 162:
```suggestion
    private void addToHead(Node<T> newNode) {
```
```suggestion
    private void addToStart(Node<T> newNode) {
```
```suggestion
    private void addToBeginning(Node<T> newNode) {
```
begin is a verb

okuzan (changes requested):

koval-kn (approved):

PR #1362, opened 2024-02-28
jv-linked-list solution

vikkushnir:


MolchanovAlexander (approved):
good idea to print out index

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,166 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private static final String INDEX_OUT_OF_BOUNDS_EXCEPTION
+            = "Invalid index: ";
+    private static final String SIZE = ", Size: ";
```
karma-o, line 8: String format would be more handy to use instead of this

karma-o (approved):
GJ! Approved with a minor suggestion

src/main/java/core/basesyntax/MyLinkedList.java
vikkushnir: Understood, will stick to it in the future

PR #1363, opened 2024-02-28
Jv linked list home work solution branch

JohnGrey17:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,170 @@
...
+        }
+        return current;
+    }
+
+    private void validateIndexForAdd(int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException(INDEX_FOR_EXCEPTION
+                    + index + SIZE_FOR_EXCEPTION + size);
+        }
+    }
```
vikkushnir, line 162: Probably, if you using this code one time, you can add it to the add((T value, int index) method and don't need to create additional

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,170 @@
...
+            addFirst(newNode);
+        } else if (index == size) {
+            addLast(newNode);
+        } else {
+            Node<T> previousNode = getNode(index - NUMBER_TO_SUBTRACT);
+            newNode.next = previousNode.next;
+            previousNode.next = newNode;
+            size++;
+        }
     }
```
vikkushnir, line 42: I think you can combine this method with addFirst and addLast to something like addNode(T value, int index)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,170 @@
...
+                tail = null;
+            }
+        } else {
+            previous.next = current.next;
+            if (current.next == null) {
+                tail = previous;
+            }
+        }
+    }
+
```
vikkushnir, line 119: maybe the unlink method should be after isEmpty()

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,170 @@
...
+            newNode.next = previousNode.next;
+            previousNode.next = newNode;
+            size++;
+        }
     }
 
     @Override
     public void addAll(List<T> list) {
+        for (int i = 0; i < list.size(); i++) {
+            add(list.get(i));
```
vikkushnir, line 47: Just in case, for me this variant is more readable
```suggestion
       for (T value : list) {
            add(value);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,170 @@
...
+    private Node<T> tail;
+
+    private static class Node<T> {
+        private T value;
+        private Node<T> next;
+
+        Node(T value) {
+            this.value = value;
+        }
+    }
```
vikkushnir, line 21: All private inner classes and methods should be after the public ones (recommended)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,170 @@
...
+        tail = newNode;
+        size++;
+    }
+
+    private Node<T> getNode(int index) {
+        Node<T> current = head;
+        for (int i = 0; i < index; i++) {
+            current = current.next;
+        }
+        return current;
```
vikkushnir, line 154: Screenshot 2024-02-29 at 14 42 35

vikkushnir (changes requested):
Left some comments to improve, as for me some changes should be, what about addNode(T value, int index) as you wish

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,167 @@
+
 package core.basesyntax;
 
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private static final String INDEX_FOR_EXCEPTION = "Index: ";
+    private static final String SIZE_FOR_EXCEPTION = ", Size: ";
+    private static final int NUMBER_TO_SUBTRACT = 1;
+    private static final int ZERO = 0;
```
karma-o, line 10: `NUMBER_TO_SUBTRACT` is not an informative name, it should reflect for what is it used/why exactly this value?
same with `ZERO`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,167 @@
...
+            } else {
+                tail.next = newNode;
+            }
+            tail = newNode;
+        } else {
+            Node<T> previousNode = getNode(index - NUMBER_TO_SUBTRACT);
+            newNode.next = previousNode.next;
+            previousNode.next = newNode;
+        }
+        size++;
```
karma-o, line 115: It's a bit hard to read, maybe we can do something about it?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,167 @@
...
+    }
+
+    private void unlinkByValue(Node<T> previous, Node<T> current) {
```
karma-o, line 118: couple points regarding the unlink methods:
* in this one you pass  2 neighboring nodes in params - we need only one
* you have 2 methods, but when it can be only a single `unlink(Node<T> node)` which should just handle all cases if it is tail/mid/head, where you do not need to call `getNode()` or iterate through the list to get that node

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,167 @@
...
+        node.value = value;
+        return oldValue;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        validateIndexForGet(index);
+        Node<T> removedNode = unlinkByIndex(index);
+        return removedNode.value;
```
karma-o, line 57:
```suggestion
        Node<T> removedNode = unlink(getNode(index));
        return removedNode.value;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,167 @@
...
+        validateIndexForGet(index);
+        Node<T> node = getNode(index);
+        return node.value;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        validateIndexForGet(index);
+        Node<T> node = getNode(index);
```
karma-o, line 47: you already have this check in `getNode()`
```suggestion
        Node<T> node = getNode(index);
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,167 @@
...
+        for (T value : list) {
+            add(value);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        validateIndexForGet(index);
+        Node<T> node = getNode(index);
```
karma-o, line 40: same

karma-o (changes requested):
Well done, left some comments

andrii-hoienko (approved):
Well done!

PR #1365, opened 2024-02-28
My linked list

Recdt:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,50 +3,164 @@
...
+        return getNode(index).value;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        checkIndex(index);
+        Node<T> element = getNode(index);
+        T deletedValue = element.value;
+
```
Nikname2303, line 60: empty line

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,50 +3,164 @@
...
+            add(value);
+            return;
+        } else if (index == 0) {
+            addFront(value);
+            return;
+        }
+        checkIndex(index);
+        Node<T> temporaryNode = getNode(index - 1);
+        Node<T> newNode = new Node<>(temporaryNode, value, temporaryNode.next);
+
```
Nikname2303, line 36: empty line

Nikname2303 (approved):
good job, just some small changes

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,164 @@
...
     @Override
     public void add(T value, int index) {
+        if (index == size) {
+            add(value);
+            return;
+        } else if (index == 0) {
+            addFront(value);
+            return;
+        }
+        checkIndex(index);
```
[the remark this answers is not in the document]
Recdt: In my implementation, I can't do that, since my list starts with 0 elements. And moves the index to the next one. That is, in fact, for most checks for the presence of an element, it is OutOfBound, so it is in this case that the check should be done after these isolated cases. Or test's will break.

karma-o (approved):
Looks very good, also left a comment on the index check approach, please check it

PR #1366, opened 2024-02-29
Java core homework jv linked list

azzilian:
Implemented all methods in task + additional methods
getNode - decide to move and move from head or tail
unlink - remove by link
tail - set a tail for List

fixes - method head removed
variable Node<T> link moved back to method

### DIFF
--- a/src/main/java/core/basesyntax/MyLinkedList.java
+++ b/src/main/java/core/basesyntax/MyLinkedList.java
@@ -5,11 +5,10 @@ import java.util.List;
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
     private Node<T> head;
     private Node<T> tail;
-    private int size;
     private Node<T> node;
+    private int size;
 
     public MyLinkedList() {
-
     }
 
     @Override
@@ -43,7 +42,6 @@ public class MyLinkedList<T> implements MyLinkedListInterface<T> {
         for (T nodes : list) {
             add(nodes);
         }
-//        tail = size;
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


bodya4243 (approved):
All good

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,157 @@
...
     @Override
     public void add(T value) {
+        add(value, size);
     }
 
     @Override
     public void add(T value, int index) {
+        checkIndexForAdd(index);
+        if (index == size) {
+            tail(value);
```
karma-o, line 24: you see, your gradation and hierarchy of method calls in methods has gone the other way around because you are calling a more complex method inside a simpler one, when usually you descent from more abstract complex and vogue functionality to the smaller and simpler methods.
For example here it should have been like this
```suggestion
    public void add(T value) {
        // simpler logic than add by index
    }

    @Override
    public void add(T value, int index) {
        checkIndexForAdd(index);
        if (index == size) {
            add(value)
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,157 @@
...
+        } else {
```
karma-o, line 25: use return here instead of else

karma-o (changes requested):
GJ! commented a minor things

azzilian:
Wait something wrong with my push

azzilian:
https://i.imgur.com/zVs8b2u.png oops I didnt saw it first so now I got..kinda queue of pushes ..

azzilian:
Tests - image
mvn clean package - 
image

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,155 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private static class Node<T> {
```
andrii-hoienko, line 90: Nested classes should be in the end of outer classes

andrii-hoienko (approved):
Nice

PR #1367, opened 2024-02-29
tests and checklist are done i think

MolchanovAlexander:


Eidenn005 (approved):
GJ.
Nice custom logic.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,197 @@
...
+            head = tail = new Node<>(null, value, null);
+        } else {
+            tail = tail.next = new Node<>(tail, value, null);
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        addByIndexRoutine(value, index);
```
karma-o, line 22: let's just remove the private method and move it's logic to here instead, because the signature of the method is identical and it contains only a call to the private one, thus can be a public, it is also used only here

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,197 @@
...
     @Override
     public T remove(int index) {
-        return null;
+        return unlinkNodeByIndex(index);
     }
 
     @Override
     public boolean remove(T object) {
-        return false;
+        return unlinkNodeByObject(object);
```
karma-o, line 50: same here, there is no need to encapsulate each method's logic into a private methods, if you are thinking about reusage - we can reuse them by calling public methods as well

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,197 @@
...
+
+        }
+        size++;
+    }
+
+    private T setValueRoutine(T value, int index) {
+        checkSetIndexInBound(index);
+        Node<T> nodeToSet;
+        if (index == 0) {
+            nodeToSet = head;
```
karma-o, line 85: I do not think it is necessary, `getNode()` should be able to handle such a case, if not - we should fix it there

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,197 @@
...
+        for (int i = 0; i < size; i++) {
+            if (isCurrentNodeValueEquals(value, current)) {
+                return current;
+            }
+            current = current.next;
+        }
+        return current;
+    }
+
+    private boolean isCurrentNodeValueEquals(T value, Node<T> current) {
```
karma-o, line 121: let's simplify this method a bit
we could even make this method static since it is more of a util method for comparison, but let's leave it as it is ;)
```suggestion
    private boolean areValuesEqual(T first, T second) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,197 @@
...
+
+    private Node<T> getNodeByValue(T value) {
+        Node<T> current = head;
+        for (int i = 0; i < size; i++) {
+            if (isCurrentNodeValueEquals(value, current)) {
+                return current;
+            }
+            current = current.next;
+        }
+        return current;
```
karma-o, line 118: you are returning head if you did not find the node by value, we should return null instead, because it is misinformative
```suggestion
        return null;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,197 @@
...
+            }
+            if (index == size - 1) {
+                tail = previousNode;
+            }
+        }
+        size--;
+        return removedElement;
+    }
+
+    private boolean unlinkNodeByObject(T object) {
```
karma-o, line 151: just iterate over nodes, find the one you need, and call `remove(index)` on it

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,197 @@
...
+        }
+    }
+
+    private void checkSetIndexInBound(int index) {
+        if (index >= size || index < 0) {
+            throw new IndexOutOfBoundsException("No such index exist");
+        }
+    }
+
+    static class Node<T> {
```
karma-o, line 188:
```suggestion
    private static class Node<T> {
```

karma-o (changes requested):
some comments were left

PR #1368, opened 2024-02-29
implemented simplified version of linked list

kaiiiseeel:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,153 @@
...
+
+    private Node<T> getNode(int index) {
+        checkIndex(index);
+        boolean isInFirstHalf = size / 2 >= index;
+        Node<T> currentNode = isInFirstHalf ? head : tail;
+        int countOfIterations = isInFirstHalf ? index : size - index - 1;
+        for (int i = 0; i < countOfIterations; i++) {
+            currentNode = isInFirstHalf ? currentNode.next : currentNode.prev;
+        }
+        return currentNode;
```
karma-o, line 141: this may work but very hard to read, your logic is split with ternary operators instead of something like this:
```java
if (isfirstHalf) {
    findfromHead(index);
} else {
    findFromTail(index);
}
```

karma-o (approved):
Approved, but please check the comment for future use

PR #1370, opened 2024-02-29
First solution

Nikname2303:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,149 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size = 0;
```
karma-o, line 6: redundant because primitives class-fields have default values
```suggestion
    private int size;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,149 @@
...
     @Override
     public void add(T value) {
+        add(value, size);
     }
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Invalid index " + index);
+        }
```
karma-o, line 19: move to separate method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,149 @@
...
+                if (tail != null) {
+                    tail.next = newNode;
+                }
+                tail = newNode;
+                if (size == 0) {
+                    head = newNode;
+                }
+                size++;
+            }
+        }
```
karma-o, line 56: I'd reuse `add(T value)` method in a loop here

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,149 @@
...
+        return current;
+    }
+
+    private void checkIndexElement(int index) {
+        if (!getIndexElement(index)) {
+            throw new IndexOutOfBoundsException("Invalid index" + index);
+        }
+    }
+
+    private void removeItem(Node<T> node) {
```
karma-o, line 138: since item is also a name of the field where we store value of the node I'd recommend changing method name to something else to avoid any confusion between them
```suggestion
    private void unlink(Node<T> node) {
```

karma-o (changes requested):
Very good job! Just some minor things

PR #1371, opened 2024-02-29
LinkedList task

bodya4243:
1) Create class Node.
2) Create size, head, tail.
3) Create unlink method.
4) Create node method.

### DIFF
--- a/src/main/java/core/basesyntax/MyLinkedList.java
+++ b/src/main/java/core/basesyntax/MyLinkedList.java
@@ -3,45 +3,196 @@ package core.basesyntax;
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size = 0;
+    private Node<T> head;
+    private Node<T> tail;
+
+    private static class Node<T> {
+        private T value;
+        private Node<T> prev;
+        private Node<T> next;
+
+        public Node(Node<T> prev, T value, Node<T> next) {
+            this.value = value;
+            this.prev = prev;
+            this.next = next;
+        }
+    }
+
     @Override
     public void add(T value) {
+        size++;
+        final Node<T> l = tail;
+        final Node<T> newNode = new Node<>(l, value, null);
+        tail = newNode;
+        if (l == null) {
+            head = newNode;
+        } else {
+            l.next = newNode;
+        }
     }
 
     @Override
     public void add(T value, int index) {
+        size++;
+        checkPositionIndex(index);
+
+        Node<T> newNode = defaultNode(value);
+
+        if (index == 0) {
+            newNode.next = head;
+            head = newNode;
+        } else if (index == size - 1) {
+            node(size - 2).next = newNode;
+            newNode.prev = node(size - 2);
+            tail = newNode;
+        } else {
+            Node<T> prevNode = node(index - 1);
+            newNode.next = prevNode.next;
+            newNode.prev = prevNode;
+            prevNode.next.prev = newNode;
+            prevNode.next = newNode;
+        }
+
     }
 
     @Override
     public void addAll(List<T> list) {
+        Node<T> pred;
+        Node<T> succ;
+        succ = null;
+        pred = tail;
+
+        for (T o : list) {
+            Node<T> newNode = new Node<>(pred, o, null);
+            if (pred == null) {
+                head = newNode;
+            } else {
+                pred.next = newNode;
+            }
+            pred = newNode;
+        }
+
+        if (succ == null) {
+            tail = pred;
+        } else {
+            pred.next = succ;
+            succ.prev = pred;
+        }
+
+        size += list.size();
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkPositionIndex(index);
+        return node(index).value;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        checkPositionIndex(index);
+        Node<T> x;
+        T oldVal;
+        if (index == size) {
+            x = tail;
+            x.value = value;
+            oldVal = x.value;
+            return oldVal;
+        }
+        x = node(index);
+        oldVal = x.value;
+        x.value = value;
+        return oldVal;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        checkPositionIndex(index);
+        return unlink(node(index));
     }
 
     @Override
     public boolean remove(T object) {
+        if (object == null) {
+            for (Node<T> x = head; x != null; x = x.next) {
+                if (x.value == null) {
+                    unlink(x);
+                    return true;
+                }
+            }
+        } else {
+            for (Node<T> x = head; x != null; x = x.next) {
+                if (object.equals(x.value)) {
+                    unlink(x);
+                    return true;
+                }
+            }
+        }
         return false;
     }
 
     @Override
     public int size() {
-        return 0;
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,196 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size = 0;
```
karma-o, line 6:
```suggestion
    private int size;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,196 @@
...
+    private Node<T> head;
+    private Node<T> tail;
+
+    private static class Node<T> {
```
karma-o, line 10: nested classes are usually at the end of the file

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,196 @@
...
+            this.value = value;
+            this.prev = prev;
+            this.next = next;
+        }
+    }
+
     @Override
     public void add(T value) {
+        size++;
+        final Node<T> l = tail;
```
karma-o, line 25: bad naming, and there is no need in that var,just use tail that you already have

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,196 @@
...
+
+    private void checkPositionIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException(outOfBoundsMsg(index));
+        }
+    }
+
+    private Node<T> defaultNode(T value) {
+        return new Node<>(null, value, null);
+    }
```
karma-o, line 173: I wouldn't say that we need a method just for that

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,196 @@
...
+            prevNode.next = newNode;
+        }
+
     }
 
     @Override
     public void addAll(List<T> list) {
+        Node<T> pred;
+        Node<T> succ;
+        succ = null;
```
karma-o, line 63: I do not get the names of these variables

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,196 @@
...
+            newNode.next = prevNode.next;
+            newNode.prev = prevNode;
+            prevNode.next.prev = newNode;
+            prevNode.next = newNode;
+        }
+
     }
 
     @Override
     public void addAll(List<T> list) {
```
karma-o, line 60: just iterate over that list and call `add(T value)` method instead of this logic

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,196 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    Node<T> node(int index) {
```
karma-o, line 145: very strange method
* method names should be verbs, this name looks more like at object not an action that we can do
* missing access modifier

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,196 @@
...
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException(outOfBoundsMsg(index));
+        }
+    }
+
+    private Node<T> defaultNode(T value) {
+        return new Node<>(null, value, null);
+    }
+
+    T unlink(Node<T> x) {
```
karma-o, line 175: missing access modifier

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,196 @@
...
+                    return true;
+                }
+            }
+        } else {
+            for (Node<T> x = head; x != null; x = x.next) {
+                if (object.equals(x.value)) {
+                    unlink(x);
+                    return true;
+                }
+            }
```
karma-o, line 130: combine these 2 loops and its if condition into one, because they have the same body

karma-o (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
...
+        size--;
+        return element;
+    }
+
+    private static class Node<T> {
+        private T value;
+        private Node<T> prev;
+        private Node<T> next;
+
+        public Node(Node<T> prev, T value, Node<T> next) {
```
andrii-hoienko, line 157:
```suggestion
        private Node(Node<T> prev, T value, Node<T> next) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
...
-        return null;
+        checkPositionIndex(index);
+        return searchIndex(index).value;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        checkPositionIndex(index);
+        Node<T> x;
```
andrii-hoienko, line 61: bad name

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
...
     @Override
     public T remove(int index) {
-        return null;
+        checkPositionIndex(index);
+        return unlink(searchIndex(index));
     }
 
     @Override
     public boolean remove(T object) {
+        for (Node<T> x = head; x != null; x = x.next) {
```
andrii-hoienko, line 83: Bad name

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
...
+        } else {
+            Node<T> current = tail;
+            for (int i = size - 1; i > index; i--) {
+                current = current.prev;
+            }
+            return current;
+        }
+    }
+
+    private String outOfBoundsMsg(int index) {
```
andrii-hoienko, line 118: No need to nake it method, it is possible to use String.format

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,161 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private Node<T> searchIndex(int index) {
```
andrii-hoienko, line 102: Bad name, you searching node, not index

andrii-hoienko (changes requested):
Nice, but some improvements required

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
+        }
+    }
+
+    private void isValidIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Index: " + index + ", Size: " + size);
+        }
+    }
+
+    private T unlink(Node<T> x) {
```
karma-o, line 125: we should not name anything with just a single letter, and please be more attentive in fixing the comments me and Andrii left quite a few regarding the naming and all of them are very similar. So if you receive a comment try and look up similar cases and fix them as well to avoid facing the same problem each review

PR #1374, opened 2024-03-02
All assumptions met

TylTejlor:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,205 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+
+    static class Node<T> {
```
okuzan, line 9:
```suggestion
    private static class Node<T> {
```
implementation details should be hidden

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,205 @@
...
+        private T value;
+        private Node<T> next;
+        private Node<T> prev;
+
+        public Node(Node<T> prev, T value, Node<T> next) {
```
okuzan, line 14:
```suggestion
        Node(Node<T> prev, T value, Node<T> next) {
```
```suggestion
        private Node(Node<T> prev, T value, Node<T> next) {
```
won't be accessible outside of the file either way

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,205 @@
...
+        } else {
+            tail.next = newNode;
+        }
+        tail = newNode;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size()) {
+            throw new IndexOutOfBoundsException("Indeks jest poza zakresem listy: " + index);
```
okuzan, line 36: Please write exception messages in English
Including size can make it more informative - imagine getting an exception "Index is out of range: 11, for size: 10", you immediately know you are off by one somewhere in your code

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,205 @@
...
+        }
         return null;
     }
 
     @Override
     public T set(T value, int index) {
+        if (index < 0 || index >= size()) {
+            throw new IndexOutOfBoundsException("Indeks jest poza zakresem listy: " + index);
+        }
+
```
okuzan, line 98: Code duplication
All guard clauses are the same for each method besides `add`, extract them to a private utility checker method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,205 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
```
okuzan, line 7: please introduce a field to track collection's size, return it in size()
it's inefficient to iterate the whole collection just to know current size

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,205 @@
...
+
+    private Node<T> getNodeAtIndex(int index) {
+        if (index < 0 || index >= size()) {
+            throw new IndexOutOfBoundsException("Invalid index: " + index);
+        }
+
+        Node<T> current = head;
+        for (int i = 0; i < index; i++) {
+            current = current.next;
+        }
```
okuzan, line 186: [checklist](https://github.com/mate-academy/jv-linked-list/blob/master/checklist.md#dont-iterate-your-nodes-from-the-head-to-the-tail-or-vice-versa
image
)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,205 @@
...
+    private boolean removeNonNull(T object) {
+        Node<T> current = head;
+
+        while (current != null) {
+            if (object.equals(current.value)) {
+                removeNode(current);
+                return true;
+            }
+            current = current.next;
+        }
```
okuzan, line 159: code duplication
merge this logic and put it into remove, you just need to combine the condition in `if`
it should be as in Objects.equals (but write it out yourself since you can't use it in this task)
`a == b || a != null && a.equals(b)`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,205 @@
...
 
     @Override
     public void addAll(List<T> list) {
+        for (T value : list) {
+            add(value);
+        }
     }
 
     @Override
     public T get(int index) {
```
okuzan, line 75: just use your getNodeAtIndex private method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,205 @@
...
+        while (current != null && currentIndex < index) {
+            current = current.next;
+            currentIndex++;
+        }
+
+        if (current != null && currentIndex == index) {
+            T oldValue = current.value;
+            current.value = value;
+            return oldValue;
+        }
```
okuzan, line 111: use getNodeAtIndex private method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,205 @@
...
+            if (current.next == null) {
+                Node<T> newNode = new Node<>(current, value, null);
+                current.next = newNode;
+                tail = newNode;
+            } else {
+                Node<T> newNode = new Node<>(current, value, current.next);
+                current.next.prev = newNode;
+                current.next = newNode;
+            }
+        }
```
okuzan, line 64: a bit too bulky, try to simplify (for example, using add(value) when size == index) or extract to private methods

okuzan (changes requested):
Good job overall. Reuse your existing logic to avoid code duplication, optimize iteration to the index, add `size` field

src/main/java/core/basesyntax/MyLinkedList.java
TylTejlor: not sure how to implement getNodeAtIndex there

nowickimj: You've already implemented `getNodeAtIndex` method, so you can call it here and retrieve the item.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,199 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return head == null;
+    }
+
+    private void isIndexOutOfArray(int index) {
```
nowickimj, line 187: More appropriate name for this method should be `checkIndexBounds`, `validateIndex` or something similar. According to general Java naming convention, "is" prefix implies that this method is of a boolean type.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,199 @@
...
+        if (index < 0 || index >= size()) {
+            throw new IndexOutOfBoundsException("Index is out of range: "
+                    + index + " for size: " + size());
+        }
+    }
+
+    private void isIndexValid(int index) {
```
nowickimj, line 194: Why did you duplicate this method? You've already created `isIndexOutOfArray` but there is no difference in their behavior.

nowickimj (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
TylTejlor: Once im calling that index is out of bounds, second one calls for invalid index, forgot how to implement it when its called(to change messege when i call it).

src/main/java/core/basesyntax/MyLinkedList.java
TylTejlor: I renamed those as u said, but i still have no idea how i can implement getNodeAtIndex into addAll method. Sorry that i write it there not under above comment, but i don't have comment window to do so.
image

src/main/java/core/basesyntax/MyLinkedList.java
TylTejlor: i change it, so i call only one of those methods without changing message

PR #1381, opened 2024-03-07
implemented linked list with base methods

vfilimo:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,166 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+
```
okuzan, line 6: don't leave an empty line after class/method definition

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,166 @@
...
+    private int size;
+    private Node<T> head;
+    private Node<T> tail;
+
+    public MyLinkedList() {
+    }
```
okuzan, line 12: no args constructor is created by default if no other constructor is present

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,166 @@
...
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private class Node<T> {
+        private T item;
```
okuzan, line 145: better put it at the very end

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,166 @@
...
+        private Node<T> prev;
+        private Node<T> next;
+
+        public Node(T item) {
```
okuzan, line 149:
```suggestion
        private Node(T item) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,166 @@
...
+            throw new IndexOutOfBoundsException("Index " + index
+                    + " must be less than " + size());
+        }
+    }
+
+    private Node<T> foundElementByIndex(int index) {
+        Node<T> node = head;
+        for (int i = 0; i < index; i++) {
+            node = node.next;
+        }
```
okuzan, line 165: worst case scenario is to iterate N nodes where N is size of the list
rewrite your solution to make worst case N/2
[checklist issue](https://github.com/mate-academy/jv-linked-list/blob/master/checklist.md#dont-iterate-your-nodes-from-the-head-to-the-tail-or-vice-versa)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,166 @@
...
+            size--;
+            return removedElement;
+        }
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> currentNode = head;
+        if (currentNode.item.equals(object)) {
+            head = currentNode.next;
```
okuzan, line 112: there's logic overlap in remove methods, extract it to a private method
[checklist issue ](https://github.com/mate-academy/jv-linked-list/blob/master/checklist.md#create-a-separate-method-unlinknode-node)

okuzan (changes requested):
Good job so far

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,174 @@
...
+            return removedElement;
+        }
+        if (index == size - 1) {
+            removedElement = tail.item;
+            tail = tail.prev;
+            size--;
+            return removedElement;
+        } else {
+            Node<T> currentNode = findElementByIndex(index);
+            removedElement = currentNode.item;
```
okuzan, line 96: if you had implemented unlink correclty, this should just be:
`return unlink(findElementByIndex(index));`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,174 @@
...
+        if (currentNode.next != null) {
+            currentNode.next.prev = currentNode.prev;
+        } else {
+            tail = currentNode.prev;
+            tail.next = null;
+        }
+        size--;
+    }
+
+    private class Node<T> {
```
okuzan, line 167: static too

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,174 @@
...
+        currentNode.prev.next = currentNode.next;
+        if (currentNode.next != null) {
+            currentNode.next.prev = currentNode.prev;
+        } else {
+            tail = currentNode.prev;
+            tail.next = null;
+        }
+        size--;
+    }
+
```
okuzan, line 166:
```suggestion
private T unlink(Node<T> node) {
        if (node.prev != null) {
            node.prev.next = node.next;
        } else {
            head = node.next;
        }

        if (node.next != null) {
            node.next.prev = node.prev;
        } else {
            tail = node.prev;
        }

        --size;

        return node.item;
    }
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,174 @@
...
+            newNode.prev = tail;
+            tail.next = newNode;
+            tail = newNode;
+            size++;
+        } else {
+            Node<T> currentNode = findElementByIndex(index);
+            newNode.next = currentNode;
+            newNode.prev = currentNode.prev;
+            currentNode.prev.next = newNode;
+            currentNode.prev = newNode;
```
okuzan, line 49: let's try to simplify this
first, reuse add(value)
we should call it index == size
```
        if (index == size) {
            add(value);
            return;
        }
```
then get node by index, create node to be inserted
You will need a constructor for
```
        Node<T> node = findElementByIndex(index);
        Node<T> newNode = new Node<>(node.prev, value, node);
```
or 
```
        newNode.prev = node.prev;
        newNode.next = node;
```
and we can do everything else with just one neat `if-else`
```
        if (newNode.prev == null) {
            head = newNode;
        } else {
            newNode.prev.next = newNode;
        }

        node.prev = newNode;
        size++;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,174 @@
...
+        for (T element : list) {
+            add(element);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        checkIndexValue(index);
+        return findElementByIndex(index).item;
```
okuzan, line 64: include checker inside of findElement By Index

okuzan (changes requested):

kozhukhovsky (approved):

PR #1387, opened 2024-03-27
solved task with linkedList

Yaromyros:
To start with, I created inner class Node to work with it all along homework. Next, I overrided methods of adding some values to list and relinked all links to new variables(Nodes). After that, I wrote set and get functions with removal methods. These methods check indexes and relink links without our removed value. Also, I added size method and isEmpty which overall are not difficult.

By myself, I added method getNodeByIndex, which optimized many rows of code. The last method was about unlinking nodes.

Overall, it was truly difficult to work with LinkedList, but crucial to understand process of it, in order to use it in future tasks.

### DIFF
--- a/src/main/java/core/basesyntax/MyLinkedList.java
+++ b/src/main/java/core/basesyntax/MyLinkedList.java
@@ -3,45 +3,160 @@ package core.basesyntax;
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node head;
+
+    private class Node {
+        private T data;
+        private Node prev;
+        private Node next;
+
+        public Node(T data) {
+            this.data = data;
+        }
+    }
+
     @Override
     public void add(T value) {
+        Node newNode = new Node(value);
+        if (head == null) {
+            head = newNode;
+        } else {
+            Node current = head;
+            while (current.next != null) {
+                current = current.next;
+            }
+            current.next = newNode;
+            newNode.prev = current;
+        }
     }
 
     @Override
     public void add(T value, int index) {
+        if (index < 0) {
+            throw new IndexOutOfBoundsException("Index cannot be negative: " + index);
+        }
+        Node newNode = new Node(value);
+        if (index == 0) {
+            newNode.next = head;
+            if (head != null) {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,160 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node head;
+
```
MAMentorFX, line 7: Linked list has head and tail.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,160 @@
...
+    private class Node {
+        private T data;
+        private Node prev;
+        private Node next;
+
+        public Node(T data) {
```
MAMentorFX, line 13:
```suggestion
        private class Node<T> {
        private T data;
        private Node<T> prev;
        private Node<T> next;
        public Node(Node<T> prev, T value, Node<T> next)
```

MAMentorFX (changes requested):
Default LinkedList has head and tail, please, redo

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,168 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node head;
+    private Node tail;
+
+    private class Node {
```
kozhukhovsky, line 9: It is better to move the inner class to the very bottom of the outer class

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,168 @@
...
+            if (object == null ? current.data == null : object.equals(current.data)) {
+                removeNode(current);
+                return true;
+            }
+            current = current.next;
+        }
         return false;
     }
 
+    private void removeNode(Node node) {
```
kozhukhovsky, line 105: all private methods must be after public ones

kozhukhovsky (approved):

PR #1403, opened 2024-04-17
Implemented logic LinkedList to MyLinkedList class

vovms:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,158 @@
...
+
+        if (foundNode.next == null) {
+            tail = foundNode.prev;
+        } else {
+            foundNode.next.prev = foundNode.prev;
+        }
+        size--;
+    }
+
+    private class Node {
```
Elena-Bruyako, line 147: Make it not dependant of the outer class
```suggestion
    private static class Node<T> {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,158 @@
...
+        private Node prev;
```
Elena-Bruyako, line 148:
```suggestion
        private Node<T> prev;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,158 @@
...
+        private Node next;
```
Elena-Bruyako, line 149:
```suggestion
        private Node<T> next;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,158 @@
...
+            return;
+        }
+        checkIndex(index);
+        Node foundNode = getElementByIndex(index);
+        link(foundNode.prev, value, foundNode);
     }
 
     @Override
     public void addAll(List<T> list) {
+        for (int i = 0; i < list.size(); i++) {
```
Elena-Bruyako, line 29: Minor comment, for-each loop looks better

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,158 @@
...
+            node = descendingOrder ? node.prev : node.next;
+            currentIndex = descendingOrder ? currentIndex - 1 : currentIndex + 1;
+        }
+        throw new NoSuchElementException("There is no such element");
+    }
+
+    private Node getElementByValue(T value) {
+        Node node = head;
+        while (node != null) {
+            if (valuesAreEqual(node.value, value)) {
```
Elena-Bruyako, line 103: There isn't necessary to use private method

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,158 @@
...
+                return node;
+            }
+            node = descendingOrder ? node.prev : node.next;
+            currentIndex = descendingOrder ? currentIndex - 1 : currentIndex + 1;
+        }
+        throw new NoSuchElementException("There is no such element");
+    }
+
+    private Node getElementByValue(T value) {
+        Node node = head;
```
Elena-Bruyako, line 101:
```suggestion
        Node<T> node = head;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,158 @@
...
+        return size == 0;
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Index is out of the bounds");
+        }
+    }
+
+    private Node getElementByIndex(int index) {
```
Elena-Bruyako, line 85:
```suggestion
    private Node<T> findNodeByIndex(int index) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,158 @@
...
+            if (index == currentIndex) {
+                return node;
+            }
+            node = descendingOrder ? node.prev : node.next;
+            currentIndex = descendingOrder ? currentIndex - 1 : currentIndex + 1;
+        }
+        throw new NoSuchElementException("There is no such element");
+    }
+
+    private Node getElementByValue(T value) {
```
Elena-Bruyako, line 100:
```suggestion
    private Node<T> findNodetByValue(T value) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,158 @@
...
+import java.util.NoSuchElementException;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node head;
+    private Node tail;
+    private int size;
+
     @Override
     public void add(T value) {
+        link(tail, value, null);
```
Elena-Bruyako, line 13: It's really difficult to read the code of add methods, because the developer should always look below link method and remember what you provide in the method. Also, the logic of this two add methods are different, so please try to make changes.

For example, this add method:
```
        if (head == null) {
            head = tail = node;
        } else {
            tail.next = node;
            node.prev = tail;
            tail = node;
        }
        size++;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,158 @@
...
     }
 
     @Override
     public void add(T value, int index) {
+        if (size == index) {
+            add(value);
+            return;
+        }
+        checkIndex(index);
```
Elena-Bruyako, line 22: This method inappropriate in this case, index > size not index >= size

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,158 @@
...
+        Node foundNode = getElementByIndex(index);
+        T value = foundNode.value;
+        unlink(foundNode);
+        return value;
     }
 
     @Override
     public boolean remove(T object) {
-        return false;
+        Node foundNode = getElementByValue(object);
```
Elena-Bruyako, line 61:
```suggestion
        Node<T> current = getElementByValue(object);
```
Please, check all places where you should use generic type

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
okuzan: the way you placed it after check for index == size is appropriate, they are functionally equivalent, good job :)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,156 @@
...
+    }
+
+    private Node<T> getElementByIndex(int index) {
+        boolean descendingOrder = index > (size / 2);
+
+        Node<T> node = descendingOrder ? tail : head;
+        int currentIndex = descendingOrder ? size - 1 : 0;
+        while (node != null) {
+            if (index == currentIndex) {
+                return node;
```
okuzan, line 118: cool implementation

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,156 @@
...
+            }
+            node = descendingOrder ? node.prev : node.next;
+            currentIndex = descendingOrder ? currentIndex - 1 : currentIndex + 1;
+        }
+        throw new NoSuchElementException("There is no such element");
```
okuzan, line 123: would be informative to include element in the message

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,156 @@
...
+            node.prev = tail;
+            tail = node;
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Index isn`t fit for adding");
```
okuzan, line 27: include index and size in the message for debugging purposes

okuzan (approved):

PR #1407, opened 2024-04-22
linkedlist implementation

BelitskyiAlexandr:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,162 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    private static class Node<T> {
```
Elena-Bruyako, line 10: Inner class should be placed at the end of outer class

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,162 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
```
Elena-Bruyako, line 5: You don't need to create methods addFirst and addLast with duplication of code. 
For implement logic of LinkedList, you can create private methods like indexValidation, unlink, findNodeByIndex, findIndexByValue. 
Try to refactor the code according to my comment and make it more readable and simplified

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,162 @@
...
     @Override
     public T remove(int index) {
-        return null;
+        Node<T> nodeToRemove = getNode(index);
+        return unlink(nodeToRemove);
     }
 
     @Override
     public boolean remove(T object) {
+        for (Node<T> x = head; x != null; x = x.next) {
```
Elena-Bruyako, line 76: x - is bad naming

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,162 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private T unlink(Node<T> node) {
```
Elena-Bruyako, line 95: You could simplified your logic 
if (node.equals(head)) {} else {}
if (node.equals(tail)) {} else {}

Elena-Bruyako, line 95:
```suggestion
    private void unlink(Node<T> node) {
```

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
BelitskyiAlexandr: image
Why should I make it void if I need to return T type?

BelitskyiAlexandr: simplification of the condition? 
` if (previousNode == null)  ->  if (node.equals(head))`
 Isn't it more logical that we work with bindings/links to objects(I don't know how it is named), and not with nodes and their contents (equals)?

Rommelua (approved):

PR #1411, opened 2024-04-29
First solution

mateuszwojtkowiak:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,215 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+    private int halfOfSize = size / 2;
```
nowickimj, line 9: Keeping this field forces you to maintain it's proper value - please remove it.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,215 @@
...
+            Node currentNode = new Node<>(null, value, null);
+            tail = currentNode;
+            head = currentNode;
+            size++;
+        } else if (size > 0) {
+            Node previousNode = head;
+            Node currentNode = new Node(previousNode, value, null);
+            previousNode.next = currentNode;
+            head = currentNode;
+            size++;
```
nowickimj, line 23: 1. This code could be extracted to private method and be used for other cases, like inserting on index. How about `addLast()`?
2. Move common code outside `if/else` block

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,215 @@
...
+    private int halfOfSize = size / 2;
+
     @Override
     public void add(T value) {
+        if (size == 0) {
+            Node currentNode = new Node<>(null, value, null);
+            tail = currentNode;
+            head = currentNode;
+            size++;
+        } else if (size > 0) {
```
nowickimj, line 18: `else` - we don't consider negative size :)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,215 @@
...
+                }
+                if (size < halfOfSize) {
+                    counter++;
+                    currentNode = currentNode.next;
+                } else if (size >= halfOfSize) {
+                    counter--;
+                    currentNode = currentNode.prev;
+                }
+            }
+        }
```
nowickimj, line 67: This method is too big. 
How about extracting this code to two separate methods: `getByIndexForward()` and `getByIndexBackward()`?

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,215 @@
...
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Your index was: " + index);
+        }
+        if (index == size) {
+            add(value);
+            return;
+        } else if (index == 0 && size > 0) {
```
nowickimj, line 35: You've already checked for `index == size` so in this case size will always be greater than 0 - simplify the condition.

nowickimj (changes requested):
There's a lot of code copy-pasting in your solution - please extract common parts into private method and use them. Few ideas:
- addFirst()
- addLast()
- link() - to insert object on specific index
- unlink() - to remove object on specific index
- getByIndexForwards() - iterate from head to half
- getByIndexBackwards() - iterate from tail to half

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,198 @@
...
+            }
+            currentNode = currentNode.next;
+            counter++;
+        }
+        return null;
+    }
+
+    private Node<T> getByIndexBackwards(int index) {
+        Node currentNode = tail;
+        int counter = size - 1;
```
d1sam, line 154: please avoid magic numbers.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,198 @@
...
+            if (index < size / 2) {
+                Node addedNode = getByIndexForwards(index);
+                link(addedNode, value);
+                size++;
+            } else if (index >= size / 2) {
+                Node addedNode = getByIndexBackwards(index);
+                link(addedNode, value);
+                size++;
+            }
+        }
```
d1sam, line 41: please don't complicate if-else constructions

d1sam (changes requested):
Good job, but there are some changes, that should be done.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,197 @@
...
+            currentNode = currentNode.next;
+            counter++;
+        }
+        return null;
+    }
+
+    private Node<T> getByIndexBackwards(int index) {
+        Node currentNode = tail;
+        int nodesLength = size - 1;
+        int counter = nodesLength;
```
MateuszRostek, line 153: You can remove `counter` variable and just use `nodesLength`

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,197 @@
...
+            size++;
+        } else if (index < size / 2) {
+            Node addedNode = getByIndexForwards(index);
+            link(addedNode, value);
+            size++;
+        } else if (index >= size / 2) {
+            Node addedNode = getByIndexBackwards(index);
+            link(addedNode, value);
+            size++;
+        }
```
MateuszRostek, line 39: You can use `return` instead of first else-if to improve readability, also you can extract some common logic (like node creation or size incrementation) out of the else-if blocks

MateuszRostek (approved):
Good job!, See some minor advice

PR #1423, opened 2024-05-17
add all methods in MyLinkedList

AnMishkin:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,185 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size() == 0;
+    }
+
+    private static class Node<T> {
```
Elena-Bruyako, line 155: Inner class should be at the end of the class

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,185 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node first;
+    private Node last;
+    private int size;
+
+    public MyLinkedList() {
+        this.size = 0;
```
Elena-Bruyako, line 11: Remove, int type = 0 by default

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,185 @@
...
+        }
+        for (T value : list) {
+            add(value);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        if (index < 0 || index >= size) {
```
Elena-Bruyako, line 81: Create separate method `indexValidation(index);` to avoid duplication

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,185 @@
...
     public T set(T value, int index) {
-        return null;
+        if (value == null) {
+            throw new IllegalArgumentException("Cannot set the value because it is null");
+        }
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Invalid index: " + index);
+        }
+        Node<T> currentNode = first;
+        for (int i = 0; i < index; i++) {
```
Elena-Bruyako, line 100: Let's create `findNodeByIndex` method to avoid duplicaiton of code

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,185 @@
...
     }
 
-    @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Invalid index: " + index);
+        }
+
+        if (index == size) {
+            Node<T> oldLast = last;
```
Elena-Bruyako, line 35:
```suggestion
            add(value);
            return;
```

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,176 @@
...
     }
 
     @Override
     public void addAll(List<T> list) {
+        if (list == null) {
+            throw new RuntimeException("Can't add list because it is null");
+        }
+        if (list.size() == 0) {
+            return;
+        }
```
okuzan, line 65:
```suggestion
```
it's redundant. If size == 0, for each won't iterate anything. Original impl fails with NPE too (like you would without addition check), so first `if`  is also not needed

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,176 @@
...
+        return currentNode;
+    }
+
+    private static class Node<T> {
+        private T value;
+        private Node<T> prev;
+        private Node<T> next;
+
+        public Node(Node<T> prev, T value, Node<T> next) {
```
okuzan, line 172:
```suggestion
        Node(Node<T> prev, T value, Node<T> next) {
```

```suggestion
        private Node(Node<T> prev, T value, Node<T> next) {
```
it has no effect, so better be upfront about it, to not confuse anyone

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,176 @@
...
+            next.prev = prev;
+            node.next = null;
+        }
+        node.value = null;
+        size--;
+    }
+
+    private void indexValidation(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Invalid index: " + index);
```
okuzan, line 155: include size in the message too, as it default Java's arrayindexoutofbounds

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,176 @@
...
+        }
+    }
+
+    private Node<T> findNodeByIndex(int index) {
+        Node<T> currentNode = first;
+        for (int i = 0; i < index; i++) {
+            currentNode = currentNode.next;
+        }
+        return currentNode;
+    }
```
okuzan, line 165: checklist issue, please read it

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,176 @@
...
+            }
+        } else {
+            while (currentNode != null) {
+                if (object.equals(currentNode.value)) {
+                    unlink(currentNode);
+                    return true;
+                }
+                currentNode = currentNode.next;
+            }
+        }
```
okuzan, line 118: remove if-else
combine conditions 104,112 into a single null safe comparison condition, like in objects equals

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,176 @@
...
+        Node<T> currentNode = findNodeByIndex(index);
+        return currentNode.value;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        if (value == null) {
+            throw new IllegalArgumentException("Cannot set the value because it is null");
+        }
```
okuzan, line 82: Java's standard LinkedList implementation in the Java Collections Framework does not prohibit setting null values

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,176 @@
...
+        indexValidation(index);
+        Node<T> currentNode = findNodeByIndex(index);
```
okuzan, line 84: Move index validation inside the "find by index" function since you always use them together.

src/main/java/core/basesyntax/MyLinkedList.java
okuzan: not fixed

okuzan (changes requested):

MAMentorFX (approved):

PR #1425, opened 2024-05-19
mplemented methods

svitlanagalicka:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,143 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    public MyLinkedList() {
+        size = 0;
```
Elena-Bruyako, line 11: redundant, int type = 0 by default

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,143 @@
...
+    }
+
+    private static class Node<T> {
```
Elena-Bruyako, line 14: inner classes should be placed at the end of outer class

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,143 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    public void checkIndex(int index) {
```
Elena-Bruyako, line 124:
```suggestion
    private void checkIndex(int index) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,143 @@
...
+        if (index >= size || index < 0) {
+            throw new IndexOutOfBoundsException("Elements by index do not exist");
```
Elena-Bruyako, line 126: Wrong message 
```suggestion
            throw new IndexOutOfBoundsException("There is no such index in the list, " + index);
```

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #1427, opened 2024-05-24
solution

mgsorigo:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,145 @@
...
+        } else {
+            Node<T> currentNode = findNodeByIndex(index);
+            Node<T> newNode = new Node<>(value, currentNode.prev, currentNode);
+            if (currentNode.prev == null) {
+                head = newNode;
+            } else {
+                currentNode.prev.next = newNode;
+            }
+            currentNode.prev = newNode;
+            size++;
```
MateuszRostek, line 36: You increase the size twice (if index == size), first by calling `add(value)` and then by using `size++`, you need to use `return` after calling `add(value)`.

MateuszRostek (approved):
Good job, approved! You made a small mistake, so check my comment.

PR #1428, opened 2024-05-24
Hw jv linked list

apanyutin:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,193 @@
...
+            newNode.prev = oldNode.prev;
+            oldNode.prev.next = newNode;
+            oldNode.prev = newNode;
+            newNode.next = oldNode;
+        }
+        size++;
+    }
+
+    private void checkIndexForAdd(int index) {
+        if (index < 0 || (size == 0 ? index > 0 : index > size)) {
```
Elena-Bruyako, line 113:
```suggestion
        if (index < 0 || index > size) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,193 @@
...
+            throw new IndexOutOfBoundsException("incorrect index");
+        }
+    }
+
+    private void checkIndexForGet(int index) {
+        if (index < 0 || (size == 0 ? index > 0 : index >= size)) {
```
Elena-Bruyako, line 119:
```suggestion
        if (index < 0 || index >= size) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,193 @@
...
+        size++;
+    }
+
+    private void checkIndexForAdd(int index) {
+        if (index < 0 || (size == 0 ? index > 0 : index > size)) {
+            throw new IndexOutOfBoundsException("incorrect index");
+        }
+    }
+
+    private void checkIndexForGet(int index) {
```
Elena-Bruyako, line 118:
```suggestion
    private void checkIndex(int index) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,193 @@
...
+            int count = 0;
+            currentNode = head;
+            while (count != index) {
+                if (index == count++) {
+                    break;
+                }
+                currentNode = currentNode.next;
+            }
+        } else {
+            int count = size - NUMBER_ONE;
```
Elena-Bruyako, line 137: NUMBER_ONE it's not quite informative constant. 
In this case, you can just use int count = size - 1;

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,193 @@
...
     @Override
     public T get(int index) {
-        return null;
+        return getNodeByIndex(index).value;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        checkIndexForGet(index);
```
Elena-Bruyako, line 42: You already checked index in getNodeByIndex(index); method, so you can remove this line

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
apanyutin: it has two different methods for checking, one of them checkIndexForGet() (when get data) other - checkIndexForAdd (when put data).

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,191 @@
...
+
+    private void checkIndexForAdd(int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("incorrect index");
+        }
+    }
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("incorrect index");
```
okuzan, line 118: add index and size to the message

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,191 @@
...
+            if (value != null
+                    && value.equals(currentNode.value) || value == currentNode.value) {
+                return currentNode;
+            }
+            currentNode = currentNode.next;
+        }
+        return null;
+    }
+
+    private class Node<T> {
```
okuzan, line 182:
```suggestion
    private static class Node<T> {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,191 @@
...
+                if (index == count--) {
+                    break;
+                }
+                currentNode = currentNode.prev;
+            }
+        }
+        return currentNode;
+    }
+
+    private boolean removeNode(Node<T> currentNode) {
```
okuzan, line 147: try to simplify this method

src/main/java/core/basesyntax/MyLinkedList.java
okuzan: if you can make an inner class static (there's no reference to enclosing class) - you should do it, better practice

okuzan (changes requested):
Good job so far!

src/main/java/core/basesyntax/MyLinkedList.java
apanyutin: I deleted one redundant condition. Is that enough?

Elena-Bruyako (approved):

PR #1429, opened 2024-05-26
Solved tasks.

oleksii-sukhenko:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,141 @@
...
+            Node<T> current = getNode(index);
+            Node<T> newNode = new Node<>(current.prev, value, current);
+            if (current.prev == null) {
+                head = newNode;
+            } else {
+                current.prev.next = newNode;
+            }
+            current.prev = newNode;
+            size++;
+        }
```
okuzan, line 39:
```suggestion
        if (index == size) {
            add(value);
            return;
        }
        Node<T> current = getNode(index);
        Node<T> newNode = new Node<>(current.prev, value, current);
        if (current.prev == null) {
            head = newNode;
        } else {
            current.prev.next = newNode;
        }
        current.prev = newNode;
        size++;
    }
```
returning early is a preferred approach in software engineering, in cases like this, call return and unwrap the rest to the code - layers of indentation make it harder to read

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,141 @@
...
+                } else {
+                    current.prev.next = current.next;
+                }
+                if (current.next == null) {
+                    tail = current.prev;
+                } else {
+                    current.next.prev = current.prev;
+                }
+                size--;
+                return true;
```
okuzan, line 96: repeated logic, checklist issue - create a separate `unlink` method to use in `remove` methods

okuzan (changes requested):
Well done!

MAMentorFX (approved):

PR #1430, opened 2024-05-27
implemented LinkedList

mhvaleria:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,163 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    private static class Node<T> {
```
Elena-Bruyako, line 10: Inner class should be placed in the end of outer class

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,163 @@
...
+        private Node(Node<T> prev, T value, Node<T> next) {
+            this.prev = prev;
+            this.value = value;
+            this.next = next;
+        }
+    }
+
     @Override
     public void add(T value) {
+        if (isEmpty()) {
```
Elena-Bruyako, line 24: Could be simpler
```suggestion
        Node<T> node = new Node<>(null, value, null);
        if (head == null) {
            head = tail = node;
        } else {
            tail.next = node;
            node.prev = tail;
            tail = node;
        }
        size++;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,163 @@
...
+            addToStart(value);
+        } else {
+            addToEnd(value);
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index == 0) {
```
Elena-Bruyako, line 34: Check index first 
```
private void checkIndexForAddMethod(int index) {
        }
    }
```

```
        if (head == null) {
        } else if (index == 0) {
        } else if (index == size) {
            add(value);
            return;
        } else {
        
        } 
        size++;
```

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #1432, opened 2024-05-29
Implemented all the required methods in MyLinkedList class.

Alona-Martynova:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,136 @@
...
+        }
+        if (currentNode.next == null) {
+            last = currentNode.prev;
+        } else {
+            currentNode.next.prev = currentNode.prev;
+        }
+        size--;
+    }
+
+    private static class Node<T> {
```
Elena-Bruyako, line 100: Inner class should be placed in the end of outer class

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,136 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> first;
+    private Node<T> last;
+    private int size;
+
     @Override
     public void add(T value) {
+        Node<T> currentNode = last;
```
Elena-Bruyako, line 12:
```suggestion
        Node<T> node = new Node<>(null, value, null);
        if (head == null) {
            head = tail = node;
        } else {
            tail.next = node;
            node.prev = tail;
            tail = node;
        }
        size++;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,136 @@
...
+            first = newNode;
+        } else {
+            currentNode.next = newNode;
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index == size) {
```
Elena-Bruyako, line 25: Do we need to check index first?
```
        }
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,136 @@
...
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Index out of range: " + index
+                    + "; size: " + size);
+        }
+    }
+
+    private Node<T> getNode(int index) {
+        checkIndex(index);
```
Elena-Bruyako, line 120: Looks like duplication of code, in methods above you call checkIndex() and then calling getNode() where call checkIndex again

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,136 @@
...
-        return null;
+        checkIndex(index);
+        return getNode(index).value;
     }
 
     @Override
     public T set(T value, int index) {
-        return null;
+        checkIndex(index);
+        T replacementValue = getNode(index).value;
```
Elena-Bruyako, line 49: To avoid calling getNode twice
```suggestion
        Node<T> necessaryNode = getNode(index);
        T replacementValue = necessaryNode.value;
        necessaryNode.value = value;
        return replacementValue;
```

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,137 @@
...
+                    + "; size: " + size);
+        }
+    }
+
+    private Node<T> getNode(int index) {
+        checkIndex(index);
+        Node<T> currentNode = first;
+        for (int i = 0; i < index; i++) {
+            currentNode = currentNode.next;
+        }
```
okuzan, line 113: checklist issue - iterate from tail if it's closer

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,137 @@
...
+            node.prev = last;
+            last = node;
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("There is no such index in the list, " + index);
```
okuzan, line 26: add size too

okuzan (changes requested):
almost done

MAMentorFX (approved):

PR #1437, opened 2024-06-13
jv-linked-list

anna-halas:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,172 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
```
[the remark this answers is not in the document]
anna-halas: if I remove the empty line the tests fail

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,172 @@
...
+            next.prev = prev;
+            node.next = null;
+        }
+
+        node.value = null;
+        size--;
+        return element;
+    }
+
+    private static class Node<T> {
```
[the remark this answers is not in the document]
anna-halas: I didn't use them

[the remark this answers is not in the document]
anna-halas: Node is at the end

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,172 @@
...
+            head = newNode;
+        }
+        tail = newNode;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        validatePositionIndex(index);
+        if (index == size) {
```
Elena-Bruyako, line 25:
```suggestion
        if (head == null) {
           
        } else if (index == 0) {

        } else if (index == size) {
            add(value);
            return;
        } else {
            
```

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,166 @@
...
     @Override
     public boolean remove(T object) {
+        if (object == null) {
+            for (Node<T> x = head; x != null; x = x.next) {
+                if (x.value == null) {
+                    unlink(x);
+                    return true;
+                }
+            }
+        } else {
```
okuzan, line 84: remove if-else - adjust internal if's condition instead, it should be null-safe check like one in Objects.equals(look it up - you can't use it directly in this hw yet)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,166 @@
...
+        if (head == null) {
+            Node<T> node = new Node<>(value, null, null);
+            head = tail = node;
+        } else if (index == 0) {
+            Node<T> newNode = new Node<>(value, head, null);
+            head = newNode;
+        } else if (index == size) {
+            add(value);
+            return;
+        } else {
```
okuzan, line 35:
```suggestion
        if (index == size) {
            add(value);
            return;
        } 
```
put this case separately, we don't need else after return

okuzan (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
-        return null;
+        validateElementIndex(index);
+        return unlink(getNodeByIndex(index));
     }
 
     @Override
-    public boolean remove(T object) {
+    public boolean remove(Object object) {
+        for (Node<T> x = head; x != null; x = x.next) {
+            if ((object == null && x.value == null) || (object != null && object.equals(x.value))) {
```
Vasul-Pankiv, line 79:
```suggestion
            if (object ==  x.valu || object != null && object.equals(x.value)) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,158 @@
...
+            Node<T> newNode = new Node<>(value, successor, predecessor);
+            successor.prev = newNode;
+            if (predecessor != null) {
+                predecessor.next = newNode;
+            }
+        }
+        if (index == size) {
+            add(value);
+            return;
+        }
```
Vasul-Pankiv, line 44: It has to be at the beginning

Vasul-Pankiv (changes requested):

sarakhmen (approved):
lgtm!

PR #1438, opened 2024-06-15
home_task_buildSuccess

VovaZSU:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,140 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
     @Override
     public void add(T value) {
+        if (head == null) {
```
Elena-Bruyako, line 12:
```suggestion
        Node<T> node = new Node<>(null, value, null);
        if (head == null) {
            head = tail = node;
        } else {
            tail.next = node;
            node.prev = tail;
            tail = node;
        }
        size++;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,140 @@
...
     @Override
     public int size() {
-        return 0;
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return head == null;
```
Elena-Bruyako, line 89:
```suggestion
        return size == 0;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,140 @@
...
+        }
+        if (nextNode != null) {
+            nextNode.prev = prevNode;
+        } else {
+            tail = prevNode;
+        }
+        size--;
+    }
+
+    private Node<T> getNode(int index) {
```
Elena-Bruyako, line 114:
```suggestion
    private Node<T> getNodeByIndex(int index) {
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,140 @@
...
+        Node<T> nodeToReplace = getNode(index);
+        T lastNodeValue = nodeToReplace.item;
+        nodeToReplace.item = value;
+        return lastNodeValue;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        Node<T> nodeToRemove = getNode(index);
```
Elena-Bruyako, line 64:
```suggestion
        checkIndex(index);
        Node<T> nodeToRemove = getNode(index);
```
It's better to check infex first and then do other logic

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,140 @@
...
+            Node<T> newNode = new Node<>(tail, value, null);
+            tail.next = newNode;
+            tail = newNode;
+            size++;
+        }
     }
 
     @Override
     public void add(T value, int index) {
+        if (index == size) {
```
Elena-Bruyako, line 26: the logic should be 

```suggestion
        if (head == null) {
        
        } else if (index == 0) {
            
        } else if (index == size) {
            add(value);
            return;
        } else {
```

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,152 @@
...
+
+    private void checkIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Invalid index" + index);
+        }
+    }
+
+    private void checkIndexByAdd(int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Invalid index" + index);
```
okuzan, line 106: include size too

okuzan (approved):
Well done

PR #1442, opened 2024-06-20
add own implementation of linked list

nikitaabo:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,191 @@
...
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private void ensureIndexWithinBounds(int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException();
```
sarakhmen, line 146: please provide error messages here and in all other places

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,191 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> first;
+    private Node<T> last;
```
sarakhmen, line 7: by convention they should be named head and tail

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,191 @@
...
+                first = null;
+            }
+        } else {
+            Node<T> current = first;
+            for (int i = 0; i < index; i++) {
+                current = current.next;
+            }
+            toRemove = current;
+            unlink(current);
+        }
```
sarakhmen, line 110: could be simplified. ALso please move the logic to unlink method

sarakhmen (changes requested):
let's improve your solution :)

Vasul-Pankiv (approved):

PR #1455, opened 2024-07-20
added solution

VadikRakitin:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,167 @@
...
     public T remove(int index) {
-        return null;
+        checkIndex(index);
+        Node<T> current = getNodeByIndex(index);
+        return unlink(current);
     }
 
     @Override
     public boolean remove(T object) {
+        if (object == null) {
```
Elena-Bruyako, line 84:
```suggestion
        int index = findIndexByValue(object);
        if (index == -1) {
            return false;
        } else {
            unlink(getNodeByIndex(index));
        }
        return true;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,167 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private static class Node<T> {
```
Elena-Bruyako, line 112: replace to the end of class

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,167 @@
...
+            Node<T> newNode = new Node<>(null, value, head);
+            if (head != null) {
+                head.prev = newNode;
+            }
+            head = newNode;
+            if (tail == null) {
+                tail = newNode;
+            }
+        } else if (index == size) {
+            Node<T> newNode = new Node<>(tail, value, null);
```
Elena-Bruyako, line 38:
```suggestion
            add(value);
            return;
```

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,175 @@
...
+            node.next = null;
+        }
+
+        node.value = null;
+        size--;
+        return value;
+    }
+
+    private int findIndexByValue(T object) {
+        int index = 0;
```
Elena-Bruyako, line 147:
```suggestion
        Node<T> node = head;
        for (int i = 0; i < size; i++) {
            if (value == node.item || value != null && value.equals(node.item)) {
                return i;
            }
            node = node.next;
        }
        return -1;
```


also, x is a bad naming

Elena-Bruyako (changes requested):

PR #1463, opened 2024-07-31
Implemented LinkedList

alisahoys:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,157 @@
...
+        Node(T value, Node<T> prev, Node<T> next) {
+            this.value = value;
+            this.prev = prev;
+            this.next = next;
+        }
+    }
+
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
```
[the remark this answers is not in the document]
alisahoys: they are already private

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,157 @@
...
 
     @Override
     public T remove(int index) {
-        return null;
+        checkIndex(index, false);
+        Node<T> current = getNode(index);
+        T removedValue = current.value;
+        unlink(current);
+        return removedValue;
+
```
okuzan, line 83:
```suggestion
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,157 @@
...
-        return false;
+        return size == 0;
+    }
+
+    private void checkIndex(int index, boolean isForAdd) {
+        if (isForAdd) {
+            if (index < 0 || index > size) {
+                throw new IndexOutOfBoundsException("Index: " + index + ", Size: " + size);
+            }
+        } else {
```
okuzan, line 115: It's OK, but IMO one usage in add method is not worth the complexity, you could just put this check there, without using the boolean flag

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,157 @@
...
+        tail = nodeToAdd;
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        checkIndex(index, true);
+        if (index == size) {
+            add(value);
+        } else {
```
okuzan, line 40:
```suggestion
        if (index == size) {
            add(value);
            return;
        }
```
it's a [preferred pattern](https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement) in such cases, indented code is harder to read, use return and we won't need else as a clause

okuzan (changes requested):
Well done! Minor code style suggestion

src/main/java/core/basesyntax/MyLinkedList.java
alisahoys: I tried, but i've got an error that my method does not override method from its superclass. Not sure why

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,156 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+
+    private static class Node<T> {
```
Elena-Bruyako, line 7: inner class should be placed at the end of the outer class

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #1468, opened 2024-08-05
Try it

Andormy:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,146 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    private static class Node<T> {
```
Elena-Bruyako, line 10: inner class should be placed at the end of the outer class

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,146 @@
...
+        T removedValue = toRemove.value;
+        unlink(toRemove);
+        return removedValue;
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> current = head;
+        while (current != null) {
+            if (current.value == object || (object != null && object.equals(current.value))) {
```
Elena-Bruyako, line 88:
```suggestion
            if (current.value == object || object != null && object.equals(current.value)) {
```

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #1489, opened 2024-09-03
first solution

Rzarcik97:


MateuszRostek (changes requested):
Please create method `findNodeAtIndex` which you then can then use in:

add(T value, int index)
get(int index)
set(T value, int index)
remove(int index)

There will be a lot of refactoring, so once you are done we will review your solution once more:)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,155 @@
...
+                node.prev.next = new Node<T>(node.prev, value, node);
+                size++;
+            }
+        }
+
     }
 
     @Override
     public void addAll(List<T> list) {
+        for (int i = 0; i < list.size(); i++) {
```
MateuszRostek, line 42: Even easier with for each:

```java
        for (T element : list) {
            add(element);
        }
```

MateuszRostek (changes requested):
Good job overall, but there are some changes required. Before creating a PR please check the `common mistakes` provided at the end of the task description: https://github.com/mate-academy/jv-linked-list/blob/master/checklist.md

You need to make changes in the `remove` methods:

image

PR #1492, opened 2024-09-08
linked_list implement

KostiaO:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,181 @@
...
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> currentNode = head;
+        int indexCounter = 0;
+
+        while (currentNode != null) {
+            if (currentNode.element != null ? currentNode.element.equals(object) : object == null) {
+                remove(indexCounter);
```
liliia-ponomarenko, line 110: simplify your solution by calling the unlink method and getting rid of the variable indexCounter

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,181 @@
...
+        public Node(Node<T> prev, T data, Node<T> next) {
+            this.element = data;
+            this.next = next;
+            this.prev = prev;
+        }
+    }
+
+    private void outOfBoundsCheck(int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException();
```
liliia-ponomarenko, line 146: Let's add informative message ;)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,181 @@
...
+        }
+    }
+
+    private Node<T> getNodeByIndex(int index) {
```
liliia-ponomarenko, line 150: Don't iterate your nodes from the head to the tail or vice versa.
Pay attention to the index you receive as an input, you can iterate only the first or second half of the list, depending on the index value. That will boost your lists' performance.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,181 @@
...
+        while (currentNode != null) {
+            if (indexCounter == index) {
+                return currentNode;
+            }
+
+            currentNode = currentNode.next;
+            indexCounter++;
+        }
+
+        throw new IndexOutOfBoundsException();
```
liliia-ponomarenko, line 165: the same

liliia-ponomarenko (changes requested):
Good job! Let's improve your solution ;)

liliia-ponomarenko (approved):
Good job! 😉

PR #1500, opened 2024-09-17
Solution

vsupruniuk:


liliia-ponomarenko (changes requested):
Let's fix some mistakes) Also, remember how to use `this` keyword

liliia-ponomarenko (approved):
Great! 😉

PR #1521, opened 2024-10-16
Implemented my own LinkedList.

Rom-Cola:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,169 @@
 package core.basesyntax;
 
 import java.util.List;
+import java.util.Objects;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+
```
okuzan, line 7: don't leave an empty line after method/class definition

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,169 @@
...
+        if (node.prev != null) {
+            node.prev.next = node.next;
+        } else {
+            tail = node.next;
+        }
+        node.next = null;
+        node.prev = null;
+    }
+
+    private class Node<E> {
```
okuzan, line 158: if it's possible to make inner class static, do it

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,169 @@
...
+                throw new IndexOutOfBoundsException();
+            }
+            foundedNode = foundedNode.next;
+        }
+        return foundedNode;
+    }
+
+    private Node<T> getNode(T value) {
+        Node<T> currentNode = tail;
+        while (!Objects.equals(value, currentNode.item)) {
```
okuzan, line 134: we still can't use Objects.equals (we will be able to in HashMap HW), so write your own code here

okuzan (changes requested):

PR #1526, opened 2024-10-23
Solution

IhorRome:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,145 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> head;
+    private Node<T> tail;
+    private int size;
+
+    private static class Node<T> {
```
liliia-ponomarenko, line 10: move the private nested class after public and private methods top-level class

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,145 @@
...
+        removeNode(node);
+        return value;
     }
 
     @Override
     public boolean remove(T object) {
+        Node<T> node = head;
+        for (int i = 0; i < size; i++) {
+            T value = node.value;
+            if (object == value || (object != null && object.equals(value))) {
```
liliia-ponomarenko, line 90:
```suggestion
            if (object == value || object != null && object.equals(value)) {
```
redundant brackets

liliia-ponomarenko (approved):
Great! A few minor comments ;)

PR #1565, opened 2024-12-28
initial commit

Tetiana-Hrytsenko:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,193 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node<T> first;
+    private Node<T> last;
+    private int size;
+
+    public MyLinkedList() {
+    }
```
okuzan, line 11: no args constructor is created by default if no other arg is present

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,193 @@
...
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        checkPositionIndex(index);
+        if (index == size) {
+            add(value);
+        } else {
```
okuzan, line 31:
```suggestion
        if (index == size) {
            add(value);
            return;
        }
```
prefer early return instead of indenting a lot of code, it's more readable

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,193 @@
...
+        }
+
+        if (next == null) {
+            last = prev;
+        } else {
+            next.prev = prev;
+        }
+    }
+
+    public static class Node<T> {
```
okuzan, line 184: make it private

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,193 @@
...
+        }
+    }
+
+    private void checkElementIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Invalid index:" + index);
+        }
+    }
+
+    private boolean equalsElement(T value, Object object) {
```
okuzan, line 153: looks like overengineering, use the same code that we have in Objects.equals

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,193 @@
...
+        if (linkAfter != null) {
+            linkAfter.prev = newNode;
+        } else {
+            last = newNode;
+        }
+    }
+
+    private void checkElementIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Invalid index:" + index);
```
okuzan, line 149: include size in the message

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,193 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
+        return size == 0;
+    }
+
+    private void checkPositionIndex(int index) {
+        if (index < 0 || index > size) {
```
okuzan, line 101: private class members should go after public ones

okuzan (changes requested):
Good job! some comments

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,180 @@
...
+
+    private void checkElementIndex(int index) {
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException("Invalid index: " + index + ". Size: "
+                    + size);
+        }
+    }
+
+    private boolean equalsElement(Object object, T value) {
+        return object == value || (object != null && object.equals(value));
```
liliia-ponomarenko, line 152:
```suggestion
        return object == value || object != null && object.equals(value);
```
redundant brackets

liliia-ponomarenko (approved):
Great!

PR #1604, opened 2025-03-01
first solution

anpiet:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,232 @@
...
+        size--;
+
+        return nodeToRemove.value;
     }
 
     @Override
     public boolean remove(T object) {
-        return false;
+        Node<T> nodeToRemove = getNodeByValue(object);
+        if (nodeToRemove == null) {
```
slade13, line 160: The `remove(T object)` method should handle the case where the object is null. The current implementation does not explicitly handle null values, which could lead to unexpected behavior.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,232 @@
...
+            tail.next = newNode;
+            newNode.prev = tail;
+            tail = newNode;
+        }
+        size++;
     }
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
```
slade13, line 68: The method `add(T value, int index)` should check if the index is strictly greater than the size, not equal to it. The current check allows adding an element at the position equal to the size, which is not correct for an index-based add operation.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,232 @@
...
+            head = newNode;
+        }
+
+        current.prev = newNode;
+        size++;
     }
 
     @Override
     public void addAll(List<T> list) {
+        if (list == null || list.isEmpty()) {
```
slade13, line 92: The `addAll(List<T> list)` method should throw a `NullPointerException` if the provided list is null, to be consistent with the behavior of `java.util.LinkedList`.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,232 @@
...
+            actualNode = tail;
+            for (int i = size - 1; i > index; i--) {
+                actualNode = actualNode.prev;
+            }
+        }
+
+        return actualNode;
+    }
+
+    private Node<T> getNodeByValue(T value) {
```
slade13, line 222: The `getNodeByValue(T value)` method should handle null values explicitly. The current implementation uses `Objects.equals`, which is correct, but it's good practice to document or handle null cases explicitly.

slade13 (approved):
Good job, please review my comments.

PR #1672, opened 2025-07-25
creating own class linked list

Lyten1:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,142 @@
 package core.basesyntax;
 
 import java.util.List;
+import java.util.Objects;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+
+    private class Node<T> {
```
Elena-Bruyako, line 8: inner class should be placed at the end of outer class

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,142 @@
...
+            this.prev = prev;
+            this.value = value;
+            this.next = next;
+        }
+    }
+
+    private Node<T> head = null;
+    private Node<T> tail = null;
+
+    private int size = 0;
```
Elena-Bruyako, line 23: int type = 0 by default
```suggestion
    private int size;
```

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -1,47 +1,142 @@
...
-        return false;
+        return size == 0;
+    }
+
+    private Node<T> getNode(int index) {
+        if (index >= size || index < 0) {
+            throw new IndexOutOfBoundsException("Getting index is out of bounds");
+        }
+        Node<T> currNode = head;
+        for (int i = 1; i <= index; i++) {
```
Elena-Bruyako, line 135: it's better to divide list / 2 and moving from head or from tail
depends on index

Elena-Bruyako (changes requested):

PR #1743, opened 2026-01-29
Mvn complete

qreqit:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,148 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private int size = 0;
```
fetis-off:
```suggestion
    private int size;
```
int contains 0 by the default

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,148 @@
...
+            tail.next = node;
+            tail = node;
+            size++;
+        }
     }
 
     @Override
     public void add(T value, int index) {
+        if (index < 0 || index > size) {
+            throw new IndexOutOfBoundsException("Can't find value with index: " + index);
```
fetis-off:
```suggestion
            throw new IndexOutOfBoundsException("Can't add  value with index: " + index);
```
because we are in `add` method :)

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,148 @@
...
+        }
+
+        Node<T> currentNode = returnNodeByIndex(index);
+        Node<T> newNode = new Node<>(currentNode.prev, value, currentNode);
+
+        if (currentNode.prev == null) {
+            head = newNode;
+        } else {
+            currentNode.prev.next = newNode;
+
```
fetis-off:
```suggestion
```
remove redundant empty line

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,148 @@
...
     public void addAll(List<T> list) {
+        for (T value : list) {
+            add(value);
+        }
     }
 
     @Override
     public T get(int index) {
-        return null;
+        return returnNodeByIndex(index).value;
```
fetis-off:
```suggestion
        return findNodeByIndex(index).value;
```
`return` is a keyword. It leads to misleading in the code. 
this method should be renamed

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,148 @@
...
+        currentNode.value = value;
+        return removerValue;
     }
 
     @Override
     public T remove(int index) {
-        return null;
+        if (index >= size || index < 0) {
+            throw new IndexOutOfBoundsException("Incottect index: " + index);
+        }
```
fetis-off:
```suggestion
        validateIndex(index)
```
I see you validate index in different places, so it would be better to make a separate method for that and reuse it in every place

fetis-off (changes requested):
Great work! Left some comments. Let`s improve your implementation

PR #1808, opened 2026-07-21
solved all tasks

Maryna-creator:


src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,147 @@
...
+    private Node first;
+    private Node last;
+    private int size;
+
+    private class Node {
+        private T value;
+        private Node next;
+        private Node prev;
+
+        Node(T value) {
```
Elena-Bruyako: The Node constructor only accepts value but should accept all fields (value, next, prev) according to the requirement.

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,147 @@
 import java.util.List;
 
 public class MyLinkedList<T> implements MyLinkedListInterface<T> {
+    private Node first;
+    private Node last;
+    private int size;
+
+    private class Node {
```
Elena-Bruyako: private class should be placed at the end of outer class

Elena-Bruyako (changes requested):
see comments

src/main/java/core/basesyntax/MyLinkedList.java
```diff
@@ -3,45 +3,150 @@
...
+        return size;
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
+    }
+
+    private class Node {
```
Elena-Bruyako, line 119: inner class should be placed at the end of the class after all methods

Elena-Bruyako (approved):
see comment for the future

